PR #169, opened 2014-04-22
added loadBlob() function / added "pause" event / added "finish" event

cbf02000:
I added a loadBlob function so that users can load blobs instead of urls.

I wanted to use your awesome library to visualize/preview a local sound recording. 
When you load() a base64-encoded string of raw audio, XHR does not like.
Since entire data is locally available in such case, I created another load function
loadBlob() so that you can directly ingest local data.

As for events, I noticed "finish" was broken and "pause" did not exist. I just went ahead and added those :)


thijstriemstra:
Looks like we've incorporated most, if not all, of this PR and it can be closed.

Issue #321, opened 2014-12-30
Dragging regions


RocAlayo:
If you make a large region you can drag it's start or end outside of the wave. This is causing several errors in my implementation as it tries to play a region and crash because the start is negative.

Thanks for your help in advance.

j0rd:
Same problem.

katspaugh:
Solved via #328.

PR #328, opened 2015-01-08
fix for issue #321

j0rd:
Attempt at a fix. Seems to work pretty well for me. I don't use resizable regions though.


katspaugh:
Thanks, Jordan!

PR #330, opened 2015-01-16
Adds support for using video element

RocAlayo:
I added support for video elements. The code is a bit messy with the double dependencies but I can't think of a better way to solve this easily. I made a [repo with an example code](https://github.com/RocAlayo/example-video-wavesurfer.js) and a [test website](http://development.example-video-wavesurfer.divshot.io).

### DIFF
--- a/src/mediaelement.js
+++ b/src/mediaelement.js
@@ -1,8 +1,8 @@
 'use strict';
 
-WaveSurfer.AudioElement = Object.create(WaveSurfer.WebAudio);
+WaveSurfer.MediaElement = Object.create(WaveSurfer.WebAudio);
 
-WaveSurfer.util.extend(WaveSurfer.AudioElement, {
+WaveSurfer.util.extend(WaveSurfer.MediaElement, {
     init: function (params) {
         this.params = params;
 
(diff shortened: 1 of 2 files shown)


katspaugh:
So, the main difference from the Audio Element is the node name, i.e. `audio` is replaced with `video`?

RocAlayo:
Yes, and the selection of the position of where to put it. 

If you want I can create a function that gets the DOM element and this way the load method will be the same with both `AudioElement` and `VideoElement` and then add the logic of the selection of the position in `AudioElement`. The method can be called something like `getDomElement`.

katspaugh:
I think we can rename the `AudioElement` backend to `MediaElement` and add a parameter of the media node type.

As for the the selection of the position of where to put the waveform, I don't see why you can't specify a container that is already located in the desired place on a webpage.

RocAlayo:
Good idea but I forgot to mention that the video element init function needs to initialize the variable `ac` with the `AudioContext` and also that the video load function only have one argument and `this.peaks` is not initialized as in `AudioElement` does. With your solution I can add this functionalities by putting two ifs that depend of the media node type parameter.

katspaugh:
What you did with `AudioContext` there is interesting. You're making a hybrid of media element for playing and Web Audio for decoding. I think `AudioElement` can have that, too.

RocAlayo:
Ok. I will modify the PR to reflect this changes then.

RocAlayo:
Ok. Changes done. This solution is better than the last I think. I added two parameters `mediaType` which is the name of the tag to add in the DOM and `mediaContainer` which specifies where to put this tag. I also added backwards compatibility for who uses `AudioElement`.

katspaugh:
Great, thanks!

PR #335, opened 2015-01-20
Minimap new functionalities (overview and regions support)

RocAlayo:
The regions are shown in the minimap and the minimap shows clearly which part of the wave is showing above. You can see an example [here](http://development.wavesurfer-bitbucket.divshot.io/) and the example code  [here](https://github.com/RocAlayo/example-wavesurfer.js-minimap)


katspaugh:
Great work, thanks a lot!

Do you think we could later refactor the overview and regions functionality in such way that it reuses the same functionality from the regions plugin?

RocAlayo:
I suppose we could but I don't see why we would. The region functionalities are developed to have a wavesurfer element as a base and the minimap is not a full wavesurfer element, it's just the wavesurfer drawer enhanced. In other words, I think if you want the functionalities of a wavesurfer element as a minimap just make another wavesurfer element and interconnect it with the main wavesurfer element.

katspaugh:
Fair enough. Thank you!

RocAlayo:
Thank you to you too. This library is great! :D

PR #342, opened 2015-02-03
Drag and seek

jaredmcateer:
I wanted the progress to update while holding down the mouse button rather than only on click so I refactored the drawer to update as you moved the mouse.


jaredmcateer:
I'm going to recreate this PR I messed up a few of the commits and it's getting messy.

katspaugh:
Please, keep in mind that the [regions plugin](http://www.wavesurfer.fm/example/annotation/) uses drag to create and resize regions. We need a way to differentiate these types of mouse moves.

jaredmcateer:
I hadn't thought of that so I did some testing just now. It seems to behave the same as what's in master currently. It gives precedence to the plugin but I'll do some more extensive testing later before I submit another PR.

PR #344, opened 2015-02-04
Region container width depends on the behavior chose in the params of wa...

RocAlayo:
With this change, regions will resize correctly when using something like:
```javascript
window.removeEventListener("resize", _.debounce(function() {
  wavesurfer.empty();
  wavesurfer.drawBuffer();
  region.updateRender();
}, 500);
```


katspaugh:
Thanks!

PR #346, opened 2015-02-08
Corrects the behaviour of the cursor (centered in the waveform container) when setPlaybackRate=>2.5

Kage1989:
Corrects the behaviour of the cursor (centered in the waveform container) when setPlaybackRate=>2.5

### DIFF
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -162,7 +162,7 @@ WaveSurfer.Drawer = {
                 this.recenterOnPosition(newPos);
             }
 
-            this.updateProgress(progress);
+            this.updateProgress(progress, true);
         }
     },
 


src/drawer.js
```diff
@@ -162,7 +162,7 @@ WaveSurfer.Drawer = {
                 this.recenterOnPosition(newPos);
             }
 
-            this.updateProgress(progress);
+            this.updateProgress(progress, true);
```
katspaugh: The `updateProgress` method takes only one argument. https://github.com/katspaugh/wavesurfer.js/blob/master/src/drawer.canvas.js#L103

katspaugh:
Please, try decreasing the re-center rate from 5 px to a smaller value (see here: https://github.com/katspaugh/wavesurfer.js/blob/master/src/drawer.js#L117), so that it re-centers more frequently.

Kage1989:
Hi katspaugh. I've tested your recommended solution. I think you meant increasing the re-center rate. For playback rate up to 4x, re-center rate set to 10 is fine. However, any playback rate higher than that, it's best to set a higher value (I've to set to more than 10 in order for the cursor to stay center at 6x speed. And yes, it distorts music to being almost indiscernible). There isn't any one value that fits all playback rate though.

katspaugh:
Then the re-center rate must be multiplied by the playback rate.

Kage1989:
Good idea. I'll try it out probably tomorrow... A bit busy with other work... Sorry! >.<

Kage1989:
I've tried multiplying the re-center rate with the playback rate... It works... But it is still limited to a certain playback rate before the cursor starts advancing again. However, I think there shouldn't be any need to push the boundaries any further. It's a good enough solution. A value of 20*getPlaybackRate() should be sufficient for playbackRate up to 10x.
I'll send another pull request to resolve the issue. Thanks for your suggestion.

Kage1989:
Very sorry for the numerous number of pull request made. I'm still new to github. >.<

thijstriemstra:
No problem, just make sure to close old ones :)

Kage1989:
hopefully resolved ...

PR #350, opened 2015-02-11
Added getPlaybackRate() method

Kage1989:
Unable to use this.setPlaybackRate(1) in this.getPlaybackRate(). 

An "Uncaught TypeError: Cannot read property '0' of undefined" error would occur at isPaused() method.


src/mediaelement.js
```diff
@@ -86,6 +86,16 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
...
     },
+    
+    /**
+     * Get the audio source playback rate.
+     */
+    getPlaybackRate: function () {
+    	if (this.media.playbackRate == null) {
+    		this.media.playbackRate = 1;
+    	}
+        return this.media.playbackRate;
```
katspaugh: I'd rather not set a value in the getter. How about just `return this.media.playbackRate || 1`?

src/mediaelement.js
```diff
@@ -86,6 +86,16 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
         this.playbackRate = value || 1;
         this.media.playbackRate = this.playbackRate;
     },
+    
+    /**
+     * Get the audio source playback rate.
+     */
+    getPlaybackRate: function () {
```
katspaugh: Actually, there's no need for this method, as it's inherited from the Web Audio backend.

Kage1989:
Ok. I'll change accordingly.

katspaugh:
Thanks! By the way, you don't need to create new PRs, you can just commit and push to the same branch of yours.

Kage1989:
oh. i didn't know that. ok. last pr from me regarding this issue. closing the original request as well.

PR #352, opened 2015-02-12
Correct cursor behaviour at high playback rates

Kage1989:
Playback rate is still limited to 10x. Any higher rates require one to setting higher value than 20


katspaugh:
@Kage1989 thanks for the patch. I think we can remove the re-center limit completely, which I did in the commit above.

Kage1989:
After playing with the ui with the new commit, i kinda realised why you had the recenter rate. It's actually a bit much visually appealing to have the canvas scroll back to the cursor's location than abruptly jumping .

katspaugh:
You're right. I've rolled that commit back.

Kage1989:
I guess we can include a clause in the readme specifying the maximum playback limit allowable. And should the user decide to use a playback rate higher than the limit, we could direct him to the constant in the recenter rate and tweak to his/her own needs.

katspaugh:
@Kage1989 good idea, thank you!

Maybe, later we will replace DOM scroll bar with Canvas-based scrolling, so this will get sorted out automatically.

PR #356, opened 2015-02-16
Multichannel display

Kage1989:
I've added methods that would allow for multichannel display. The waveform canvas would change it's height accordingly to the number of channels in the audio file. I made separate methods for getPeaks as I realised that the minimap uses the same getPeaks method.


src/drawer.canvas.js
```diff
@@ -48,6 +48,25 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.Canvas, {
 
         this.clearWave();
     },
+    
+    updateWidthnHeight: function () {
```
thijstriemstra: First of all, the method name is weird, that n looks like a typo.
Also, could you factor out the height changes into a new `updateHeight` method or something? There's also a different ticket that asks for a updateHeight method (when going fullscreen for example) so this change could then benefit both tickets.

src/drawer.js
```diff
@@ -149,6 +164,27 @@ WaveSurfer.Drawer = {
 
         this.updateWidth();
     },
+    
+    setWidthnHeight: function (width, height) {
+        if (width == this.width && height == this.height) { return; }
```
thijstriemstra: Same issue with the method name..

src/wavesurfer.js
```diff
@@ -256,8 +256,10 @@ var WaveSurfer = {
             width = parentWidth;
         }
 
-        var peaks = this.backend.getPeaks(width);
-        this.drawer.drawPeaks(peaks, width);
+//        var peaks = this.backend.getPeaks(width);
+//        this.drawer.drawPeaks(peaks, width);
```
thijstriemstra: Don't leave commented out code in there.

src/webaudio.js
```diff
@@ -196,6 +196,45 @@ WaveSurfer.WebAudio = {
...
+                    if (value > max) {
+                        max = value;
+                    // faster than Math.abs
+                    } else if (-value > max) {
+                        max = -value;
+                    }
+                }
+                
+				//Sequentially
+				peaks[c*length+i] = max;
```
katspaugh: I don't like the idea of push all channels into the same flat array. I'd rather go for an array of arrays:

    [ [ 0.5, ..., 0.8 ], [ 0.1, ..., 0.4 ] ]

To avoid breaking things, one could still accept a flat array (for mono) *and* an array of arrays (for stereo and higher).

src/drawer.canvas.js
```diff
@@ -48,6 +48,25 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.Canvas, {
...
+        var width = Math.round(this.width / this.params.pixelRatio);
+        var height = this.height;
+
+        this.waveCc.canvas.width = this.width;
+        this.waveCc.canvas.height = this.height;
+        this.style(this.waveCc.canvas, { width: width + 'px'});
+        this.style(this.waveCc.canvas, { height: height + 'px'});
```
katspaugh: this.style(this.waveCc.canvas, {
    width: width + 'px',
    height: height + 'px'
});

src/drawer.canvas.js
```diff
@@ -100,6 +119,61 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.Canvas, {
         }, this);
     },
 
+    drawWaveMainWaveform: function (peaks, length, max) {
+    // A half-pixel offset makes lines crisp
+    var $ = 0.5 / this.params.pixelRatio;
+
+    var vscale = 0.95; // scaled values of height of waveform
```
katspaugh: Could you explain what 0.95 means?

src/drawer.canvas.js
Kage1989: Yeap. Sure. I'll work on it.

src/drawer.canvas.js
Kage1989: For stereo channels, the waveform would stack on top of each other. If both channels are on full blast, I thought it would have been better to add some spacing between the waveforms. So I scaled down the waveform height to 0.95 times of its original height.

src/webaudio.js
Kage1989: I'll try out array of arrays soon. I didn't do that because you were using typed arrays previously and I wasn't sure of the repercussions of using the normal arrays.

src/wavesurfer.js
Kage1989: OK. I'll take that out.

src/drawer.canvas.js
X-Raym: Hi! That's nice that you work on this!
Can you please show us a HTML exemple of how it look like ? Thanks!

thijstriemstra: Yes, including an example would be nice as well (don't forget to include it at the bottom of index.html, the frontpage).

katspaugh:
@Kage1989, I've committed the same changes with the exception of the format of the channel-split peaks (as I said I wanted them an array of arrays). Thanks for the pull request!

Kage1989:
Sorry. wasn't of much help for the second half. was too busy with my official work. Glad to be of help though. Thanks for the code!

Issue #358, opened 2015-02-20
Calling destroy twice throws error


thijstriemstra:
When you call `destroy` twice, in the demo for example:

```diff
diff --git a/example/main.js b/example/main.js
index 098a25b..2617ec9 100644
--- a/example/main.js
+++ b/example/main.js
@@ -39,6 +39,8 @@ document.addEventListener('DOMContentLoaded', function () {
 // Won't work on iOS until you touch the page
 wavesurfer.on('ready', function () {
     //wavesurfer.play();
+    wavesurfer.destroy();
+    wavesurfer.destroy();
 });
 
 // Report errors
```
It'll throw this error:

```
TypeError: Argument 1 of Node.removeChild is not an object.
http://localhost:8080/wavesurfer.js/src/drawer.js
Line 173
```

It's trying to remove a child that doesn't exist the second time. I'd agree it doesn't make much sense to call it twice, but when your code does then it shouldn't throw an error imo.

PR with fix will follow.

PR #359, opened 2015-02-20
fix issue when calling destroy() twice

thijstriemstra:
refs #358


thijstriemstra:
ping @katspaugh

katspaugh:
Great, thanks!

Issue #308, opened 2014-12-04
TypeError: this.buffer is undefined


thijstriemstra:
If you comment out the `load` call in the example (`main.js`):

``` javascript
// Load audio from URL
//wavesurfer.load('example/media/demo.wav');
```

and click anywhere on the empty waveform, it will throw this error:

```
TypeError: this.buffer is undefined
http://localhost:8080/wavesurfer.js/src/webaudio.js
Line 249
```

`getDuration` in `webaudio.js` should probably not assume there's a good buffer available? Current code:

``` javascript
getDuration: function () {
        return this.buffer.duration || 0;
    },
```

thijstriemstra:
Well, turns out buffer can be `null` as well, throwing `TypeError: this.buffer is null` under certain circumstances. PR with fix will follow.

PR #361, opened 2015-02-21
return 0 when buffer is undefined or null

thijstriemstra:
refs #308


thijstriemstra:
@katspaugh, this fixes an issue when destroying an instance.

katspaugh:
OK, thanks!

PR #370, opened 2015-02-26
microphone plugin: loadDecodedBuffer doesn't empty anymore

thijstriemstra:
Do it in plugin instead.


katspaugh:
Thanks! I had to remove the `empty` call because `loadDecodedBuffer` is also used by other internal  loaders, which caused emptying twice.

PR #372, opened 2015-02-27
locally viewed fix

tomByrer:
Can't find Bootstrap without the `http:` when opened in local HD.

### DIFF
--- a/example/annotation/index.html
+++ b/example/annotation/index.html
@@ -7,7 +7,7 @@
         <link href="data:image/gif;" rel="icon" type="image/x-icon" />
 
         <!-- Bootstrap -->
-        <link href="//maxcdn.bootstrapcdn.com/bootstrap/3.3.1/css/bootstrap.min.css" rel="stylesheet">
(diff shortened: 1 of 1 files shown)


katspaugh:
Won't wavesurfer.js itself work when opened via file://? Last time I checked it wouldn't.

tomByrer:
`file://` works fine in my Firefoxes.
I tried a local web server, [Fenix](http://fenixwebserver.com/), which didn't work at all.

katspaugh:
OK, then. However, I encourage you to try a different server. The easiest option I know is `ruby -run -e httpd . -p 8080` if you are on a mac.

tomByrer:
@thijstriemstra sorry about my laziness, but it has been accepted already :worried: 

> ruby -run -e httpd . -p 8080 if you are on a mac

On Windows, & I don't have Ruby installed.
node/connect works:
```js
var express = require('express');
var app = express();
var port = process.argv[2] || 8000;

app.use(express.static(__dirname + '/'));
app.listen(port);

console.log("express server started port "+ port +" CTRL+C to stop)");
```
Let me know if you want a PR for the README.

Issue #315, opened 2014-12-17
Minified plugins


thijstriemstra:
It would be nice to have minified plugins, next to the normal build of wavesurfer. I'd imagine a `plugins` directory in the `build` directory that contains the plugins with `min` in the filename of the plugin.

tomByrer:
+1
Or just put put all miniified files in the same folder.

thijstriemstra:
@katspaugh, do you want to use `grunt` for this, and remove the 'manual' `Makefile` build script?

katspaugh:
I'm OK with grunt or gulp.

thijstriemstra:
I could use a little feedback before I submit the pull request:

- added a `Gruntfile.js`, got rid of the `Makefile`
- running `grunt` concencates everything into a single file, which is written to `dist/wavesurfer.dev.js`. This file is useful for debugging, but do you want to include this in a release as wel?
- running `grunt` scans the `plugin` folder, minifies each plugin, and places the result in `dist/plugin/wavesurfer.microphone.min.js` for example

Any thoughts?

katspaugh:
Hi Thijs,

`wavesurfer.dev.js` is unnecessary, I think. Source map is better for debugging.

As for the plugins, all sounds good. By the way, I'm planning to promote the regions plugin back to the main code.

thijstriemstra:
Ok. What about the current `build` folder, do we want to:

- keep it, and have a `dist` folder only locally
- replace it with `dist` (and break existing script links out on the web that link to the `build/wavesurfer.min.js` vs `dist/wavesurfer.min.js`
- have both a `dist` and `build` folder in the repository..

katspaugh:
I'd have only `dist`, since it seems to be the npm standard. To break the links they firstly would have to update the package, at which point they could also change the link, I think.

tomByrer:
> promote the regions plugin back to the main code

One of my main uses of wavesurfer, though I'm not everyone will use it.

thijstriemstra:
I think this is ready for a pull request. After running `grunt` from root directory you'll end up with a `dist` that looks like this:

![schermafbeelding 2015-03-02 om 20 28 08](https://cloud.githubusercontent.com/assets/305679/6448514/cb85e8b8-c11a-11e4-91fa-4776b1a28ce0.png)

katspaugh:
Cool, looking forward to the PR!

thijstriemstra:
[...]

PR #377, opened 2015-03-02
Use grunt in build process

thijstriemstra:
- replace `Makefile` with `Gruntfile.js`
- `build` is now `dist`
- updated examples and updated CSS to Bootstrap 3.3.2


thijstriemstra:
@katspaugh :) refs #315

katspaugh:
Thank you very much!

PR #386, opened 2015-03-05
Correct param assignments on region update

### LINKED ISSUE #384: regions.update updates wrong variables
loop, resize and drag in the regions.update function update the wrong variables. instead of color and data they should instantiate the right variables

shirshendu:
Fixes #384


katspaugh:
OMG, what a mistake! Thanks so much!

PR #387, opened 2015-03-05
Region now won't oversize the wave element

RocAlayo:
With this changes now the start and end of the region won't oversize the wave element in any case.


katspaugh:
Thanks!

RocAlayo:
Don't mention it. Thanks to you too for making this library :)

PR #394, opened 2015-03-10
Added listener on mouseup to body for region usability improvement.

RocAlayo:
With this code i easier  to use the dragging and resizing of region because if the mouseup event happens outside the wavesurfer element this actions are stopped and you don't get stuck with it until you click again. This way is easier to adjust a region to the star or end of the wave.

### DIFF
--- a/plugin/wavesurfer.regions.js
+++ b/plugin/wavesurfer.regions.js
@@ -329,12 +329,16 @@ WaveSurfer.Region = {
             };
 
             my.element.addEventListener('mousedown', onDown);
-            my.wrapper.addEventListener('mouseup', onUp);
             my.wrapper.addEventListener('mousemove', onMove);
+            document.body.addEventListener('mouseup', onUp);
 
             my.on('remove', function () {
-                my.wrapper.removeEventListener('mouseup', onUp);
+                document.body.removeEventListener('mouseup', onUp);
                 my.wrapper.removeEventListener('mousemove', onMove);
+            });
+
+            my.wavesurfer.on('destroy', function () {
+                document.body.removeEventListener('mouseup', onUp);
             });
         }());
     },


plugin/wavesurfer.regions.js
```diff
@@ -331,6 +331,9 @@ WaveSurfer.Region = {
             my.element.addEventListener('mousedown', onDown);
             my.wrapper.addEventListener('mouseup', onUp);
             my.wrapper.addEventListener('mousemove', onMove);
+            document.querySelector("body").addEventListener("mouseup", function () {
```
katspaugh: `document.body.addEventListener('mouseup', onUp)` — shorter, the same meaning.

Please, also use single quotes.

plugin/wavesurfer.regions.js
```diff
@@ -331,6 +331,9 @@ WaveSurfer.Region = {
             my.element.addEventListener('mousedown', onDown);
             my.wrapper.addEventListener('mouseup', onUp);
```
katspaugh: This can be removed, as we don't need two identical events.

plugin/wavesurfer.regions.js
katspaugh: The event handler on body should be removed on destroy. Subscribe to the wavesurfer's destroy event.

RocAlayo: In function `onUp` you do `e.stopPropagation();` and `e.preventDefault();`. This applied to the body can make some functionality of the web using this library to brake, that's why I put it this way. But If we remove `my.wrapper.addEventListener('mouseup', onUp);` we can delete the two lines that stop event bubbling.

RocAlayo:
Ok, I have update the pull request and I put single quotes too. In the last update I forgot to change the single quotes :P

katspaugh:
Is it working as before without `preventDefault`? Most importantly, no text selection should occur.

And you still aren't unsubscribing mouseup on wavesurfer destroy. If you add this last bit, we can merge it.

RocAlayo:
I'm sorry maybe I'm undestanding wrong the code but nothing in regions plugins is subscribed to wavesurfer destroy. All is subscribed to remove of a region witch now also remove the event listener mouseup of body. Am I wrong?

katspaugh:
We need to unsubscribe both on region remove and on global destroy. Previously, nothing was unsubscribed on destroy, because the subscribed DOM elements themselves were removed from DOM on destroy. Since `document.body` isn't going to be removed, it needs to be unsubscribed.

RocAlayo:
Ok, got it, thanks for the explanation. I will change this and test if everything goes well without the `preventDefault` and then update the PR.

katspaugh:
I guess it's working OK without `preventDefault`. Thanks!

Issue #395, opened 2015-03-11
Equalizer example not working?


normington:
Firstly, thanks for your awesome work on this, wavesurfer is amazing!

I'm trying to get something going with multiple filter nodes in the chain that wavesurfer uses. I've read about the setFilters() method implemented to do this and take an array of nodes, and I'm looking at the [Equalizer example](http://www.wavesurfer.fm/example/equalizer/) that seems to use this, but the example doesn't seem to work for me..

Other web audio API things are working fine in my browser, including the wavesurfer [panner example](http://www.wavesurfer.fm/example/panner/).

Is there something wrong my end, or is there a problem with the equalizer example as it is at the moment?

Cheers!

normington:
Found the problem, pull request #396 should solve it.

PR #396, opened 2015-03-11
Fixing equalizer example

nmgtn:
Fix for #395. After submitting the issue, realised what the problem was. Filter type was not being mapped and set correctly from the EQ array, so was just defaulting to 'lowpass'.

I'll be honest - I'm not sure why the capitalisation of the filter type stopped it working, but it did. If you're any wiser than me on that, I'd love to find out!

Tested on Chrome 40.0.2214 and Firefox Developer Edition 38.0a2

### DIFF
--- a/example/equalizer/main.js
+++ b/example/equalizer/main.js
@@ -20,41 +20,41 @@ document.addEventListener('DOMContentLoaded', function () {
         var EQ = [
             {
                 f: 32,
-                type: 'LOWSHELF'
+                type: 'lowshelf'
             }, {
                 f: 64,
-                type: 'PEAKING'
+                type: 'peaking'
             }, {
                 f: 125,
-                type: 'PEAKING'
+                type: 'peaking'
             }, {
                 f: 250,
-                type: 'PEAKING'
+                type: 'peaking'
             }, {
                 f: 500,
-                type: 'PEAKING'
+                type: 'peaking'
             }, {
                 f: 1000,
-                type: 'PEAKING'
+                type: 'peaking'
             }, {
                 f: 2000,
-                type: 'PEAKING'
+                type: 'peaking'
             }, {
                 f: 4000,
-                type: 'PEAKING'
+                type: 'peaking'
             }, {
                 f: 8000,
-                type: 'PEAKING'
+                type: 'peaking'
             }, {
                 f: 16000,
-                type: 'HIGHSHELF'
+                type: 'highshelf'
             }
         ];
 
         // Create filters
         var filters = EQ.map(function (band) {
             var filter = wavesurfer.backend.ac.createBiquadFilter();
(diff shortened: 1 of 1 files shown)


katspaugh:
Maybe let's just call `toUpperCase` for the property name?

Also, I think the name of the modes exist as constants in the Web Audio nose API. We should have used these instead of strings.

normington:
Do you mean call toLowerCase within the EQ.map function, so that any changes to the EQ array params won't mess it up in future? That sounds like a good shout, I'll do that.

The filter type is spec'd as a string in the [standard](http://webaudio.github.io/web-audio-api/#the-biquadfilternode-interface) I think, have you seen otherwise?

katspaugh:
I think I saw the possible filter types listed on the filter nodes as constants which were just properties in capital case (containing strings of course). I don't have time to check if it's true.

What I'm worried about is that some browsers may support only all-caps spelling.

normington:
I'm not sure I understand what you're getting at. FWIW, I've only ever seen the filter types as lowercase, as [here](http://webaudio.github.io/web-audio-api/#idl-def-BiquadFilterType). However, I bow to your experience here, as I'm very new to this!

I'll leave it up to you where to go from here, but please do let me know if there's some other way I should do it, or something I should test more carefully.

Cheers!

katspaugh:
I checked (http://codepen.io/anon/pen/OPBxQG) and, yeah, in Chrome the filter type is lowercase and there's no any constants to use. However, I think some version of Safari and Mobile Safari wants the type in upper case, so if you could test in Safari, it would be cool. Thanks!

normington:
That's cool, I will check it out on Safari and have a play with uppercase and lowercase properties when I can get to a Mac.

thijstriemstra:
I'm on a Mac with Safari, what am I supposed to check?

nmgtn:
Thanks for offering, @thijstriemstra . If you could load up `/example/equalizer/index.html` and check that nothing appears in your JavaScript console, that would be great. you should hear the cut and boost applied to each frequency band when you move the sliders (the original fault was that the sliders had no effect, because the underlying filters hadn't been initialised as expected).

Make sure you're using the files from this commit - easiest way is probably to clone from [my repo (click)](https://github.com/nmgtn/wavesurfer.js).

thijstriemstra:
Tested with `Mozilla/5.0 (Macintosh; Intel Mac OS X 10.7; rv:36.0) Gecko/20100101 Firefox/36.0` and `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_7_5) AppleWebKit/537.78.2 (KHTML, like Gecko) Version/6.1.6 Safari/537.78.2`, both worked fine.

katspaugh:
OK, thanks guys! Merging.

PR #406, opened 2015-03-18
make sure the element exists before styling it

thijstriemstra:
Sometimes right after destroying a wavesurfer, an additional `drawer.style` call is made somewhere but since the element does not exist, it throws an error in `drawer.js`:

```
TypeError: el is null

if (el.style[prop] != styles[prop]) {
```

This PR makes sure the element exists before styling it.

### DIFF
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -75,11 +75,13 @@ WaveSurfer.Drawer = {
     },
 
     style: function (el, styles) {
-        Object.keys(styles).forEach(function (prop) {
-            if (el && el.style[prop] != styles[prop]) {
-                el.style[prop] = styles[prop];
-            }
-        });
+        if (el) {
+            Object.keys(styles).forEach(function (prop) {
+                if (el.style[prop] != styles[prop]) {
+                    el.style[prop] = styles[prop];
+                }
+            });
+        }
         return el;
     },
 
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -76,7 +76,7 @@ WaveSurfer.Drawer = {
 
     style: function (el, styles) {
         Object.keys(styles).forEach(function (prop) {
-            if (el.style[prop] != styles[prop]) {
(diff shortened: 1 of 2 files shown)


thijstriemstra:
@katspaugh could you also release a new version after merging this, it's been a while.. :)

src/drawer.js
```diff
@@ -76,7 +76,7 @@ WaveSurfer.Drawer = {
 
     style: function (el, styles) {
```
katspaugh: It would be better to check if element exists here, and exit early if not.

katspaugh:
Have you figured out why is that extra call to style is made? The must be async callback somewhere.

thijstriemstra:
Ok I'm using the microphone plugin in a Bootstrap modal. When you open the modal the microphone plugin is displayed, press the play button, and your microphone will start emitting a signal that is drawn by wavesurfer. When you press the close button on the modal I destroy the wavesurfer instance.

What the microphone plugin does, is basically continously calling `reloadBuffer`:

```javascript
/**
 * Redraw the waveform.
 */
reloadBuffer: function(event) {
    if (!this.paused) {
        this.wavesurfer.empty();
        this.wavesurfer.loadDecodedBuffer(event.inputBuffer);
    }
},
```
So I expect that that wavesurfer `empty` and `loadDecodedBuffer` are still being called after `destroy` is called on it. So I guess making sure to set `this.paused` to `true` just before destroying the microphone plugin (and it's associated wavesurfer instance) should be enough to fix it. Let me check.

katspaugh:
So 881b8fc (checking if element exists) is not really needed? I'd prefer not to check and let it throw then, because otherwise we won't know of possible bugs.

thijstriemstra:
@katspaugh, yeah that's right, fine with me.

katspaugh:
OK, I rolled that commit back. Thanks!

Issue #400, opened 2015-03-13
Straight-through connection retained when filters added with setFilters(filters)


nmgtn:
When you use `setFilters(filters)` to add some filters in the 'FX loop' kind of structure that wavesurfer seems to use, the analyser node at the start of the chain of filters remains directly connected to the gain node at the other end of the chain. This means that what you hear is the sum of the unprocessed sound and the sound going through the filter chain. Please excuse me if have misunderstood, and this is by design, but it looks to me like a bug!

## Audio graph, without filters
Without any filters set, analyser is connected directly to gain node, which is sensible:

![wavesurfer-fx-bypass_2](https://cloud.githubusercontent.com/assets/9980206/6648149/3edac49e-c9ce-11e4-8680-0178980b82b3.png)
(picture is Firefox visualisation of the audio graph on the [wavesurfer.fm home page](http://www.wavesurfer.fm/))

## Audio graph, with filters
Adding filters using the `setFilters` method (or `setFilter`)

### Expectation
Analyser node should be disconnected from gain node, and connected to first filter in the chain. Last filter in the chain should be connected to gain node. This means that audio is routed 100% wet through the filter chain.

### Observation
Analyser node retains its direct connection to gain node, and a connection is added to first filter in the chain. Last filter in the chain is connected to gain node, so you get 100% wet + 100% dry signal:

![wavesurfer-fx-bypass](https://cloud.githubusercontent.com/assets/9980206/6648148/3cd4e92c-c9ce-11e4-846f-5ce07301b9e7.png)
(picture is Firefox visualisation of the audio graph on the [panner filter example](http://www.wavesurfer.fm/example/panner/))

## A fix?

I can't see anything in the definition of `setFilters` which is supposed to disconnect analyser node from gain node before connecting each filter in turn. I considered adding a line to `disconnectFilters` to do this, but that doesn't really make sense. Perhaps it's best to add `this.analyser.disconnect()` immediately after [line 79 of webaudio.js](https://github.com/katspaugh/wavesurfer.js/blob/master/src/webaudio.js#L79).

I wanted to check with you that I've understood properly what is supposed to be happening, before I see about fixing the issue - please do let me know if I have misunderstood! If you agree that this is a bug, let me know any pointers you have for solving it and I'll put a pull request together.

katspaugh:
Hi Andy,

Your analysis is correct and a fix is very welcome!

Thanks!

nmgtn:
[...]

PR #423, opened 2015-03-30
Fix for #400: disconnect direct path in 'FX loop'

nmgtn:
This fixes the bug in #400. Direct path from analyser to gain node is
disconnected before connecting filters in turn, so that all audio goes
exclusively through the inserted filters now.

### DIFF
--- a/src/webaudio.js
+++ b/src/webaudio.js
@@ -75,7 +75,10 @@ WaveSurfer.WebAudio = {
      * @param {Array} filters Packed ilters array
      */
     setFilters: function (filters) {
+        // Remove existing filters
         this.disconnectFilters();
+        // Disconnect direct path before inserting filters
+        this.analyser.disconnect();
 
         if (filters && filters.length) {
             this.filters = filters;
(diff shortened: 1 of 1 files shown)


thijstriemstra:
Cool, the first PR build: https://travis-ci.org/katspaugh/wavesurfer.js/builds/56459591

katspaugh:
Thanks Andy!

So now if someone first calls `setFilter` and then calls `disconnectFilters` directly, the analyser node is left disconnected from the gain node?

Maybe we should streamline the API a bit and not allow empty arguments for the `setFilters` so that one would have to call `disconnectFilters` to detach filters? And then we should connect the analyser node back to the gain node in `disconnectFilters`, so that it's never left disconnected from everything.

What do you think?

nmgtn:
I think it makes sense to remove the filters from the signal path by calling `setFilters` with no arguments, rather than calling `disconnectFilters` directly. We should probably make sure that either option works, though, since both make sense semantically.

I've made some changes so that the direct path through is reconnected after `disconnectFilters`, rather than in the `else` statement of `setFilters` that was executed if there were no filters specified. This means that the analyser node is never left disconnected from everything, even if `disconnectFilters` is called directly.

This also means that we only manually disconnect analyser node from gain node once we have checked that there are filters to put inline to replace the direct connection.

katspaugh:
Excellent, thank you!

PR #439, opened 2015-04-20
Set microphone and timeline plugins to work with require

dbunker:


katspaugh:
@thijstriemstra review please, this PR removes AMD definition from the microphone plugin.

katspaugh:
@dbunker what are the other changes?

dbunker:
Github has ordered the changes strangely, the only change was to match these two plugins up to the others. If [AMD](http://requirejs.org/docs/whyamd.html) is preferable, is there a reason the other plugins are not set up this way, and examples don't seem to use require + AMD? Thanks.

katspaugh:
Oh, I see. The diff is better viewed like this: https://github.com/katspaugh/wavesurfer.js/pull/439/files?w=1

I would prefer plain files, but the microphone, timeline and spectrogram were contributed by different people, and I just went with their preferences.

thijstriemstra:
I quickly tested this with my own app which uses `dist/wavesurfer.amd.js`. With this change that AMD version no longer works (throws a `ReferenceError: cant find WaveSurfer`) but the plugins look like they do work with the `dist/wavesurfer.min.js`. So whether this change is correct or not, I can't say for sure, AMD puzzles me. When I added the microphone plugin I made a copy of the timeline plugin, hence the AMD header in both plugins. I also prefer plain files but I wonder what @dbunker thinks of that wavesurfer.amd.js, do you also use this one or the minified version?

katspaugh:
Maybe we could add plugins to the AMD task in the gruntfile (https://github.com/katspaugh/wavesurfer.js/blob/master/Gruntfile.js#L38)?

thijstriemstra:
I guess all we need is a unit test that proves it works or not

dbunker:
@thijstriemstra I actually just use the raw unminified files. These and all my other raw js get loaded using require.js, then webpack spits them into a single minified bundle.js for me. It looks like the only change is that wrapper for amd files, would it make sense to have a separate amd-dist directory where all the plugins have this wrapper? This would be pretty easy to add as a grunt task.

katspaugh:
@dbunker could you update this PR?

PR #448, opened 2015-05-06
limit regions/callback on region update/create

glebcha:
Regions count limit patch
--------
Use option 'count: integer'

Regions callback
--------
Use option regionsCallback
Call wavesurfer.regionsCallback

Example
--------
    //Save time to input and localStorage.
    var saveRegion = function() {
        var regionHolder = document.getElementById('region-data'),
            regions = wavesurfer.regions.list,
            regionsObj = Object.keys(regions).map(function (id) {
                var region = regions[id];
                return {
                    start: region.start,
                    end: region.end
                };
            });

        for (var i = 0; i < regionsObj.length; i++) {
            var audioRegion = regionsObj[i],
                start = Math.round(audioRegion.start),
                end = Math.round(audioRegion.end);
            startInput.value = start;
            endInput.value = end;
        }

        localStorage.regions = JSON.stringify(regionsObj);
    };

    if (wavesurfer.enableDragSelection) {
        wavesurfer.enableDragSelection({
            color: 'rgba(0, 255, 0, 0.1)',
            count: 1 
        });
        wavesurfer.regionsCallback = saveRegion;
    }

### DIFF
--- a/plugin/wavesurfer.regions.js
+++ b/plugin/wavesurfer.regions.js
@@ -12,16 +12,29 @@ WaveSurfer.Regions = {
 
     /* Remove a region. */
     add: function (params) {
+        //Limit ranges count
         var region = Object.create(WaveSurfer.Region);
-        region.init(params, this.wavesurfer);
 
-        this.list[region.id] = region;
-
-        region.on('remove', (function () {
-            delete this.list[region.id];
-        }).bind(this));
-
-        return region;
+        if(Object(params).hasOwnProperty('count')) {
+            if( Object.keys(this.list).length < params.count){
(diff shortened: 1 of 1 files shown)


katspaugh:
No, sorry.

 1. Counting can be easily done in the application code, so I don't see any reasons to include it into the plugin.
 2. We don't provide callbacks in the API. We emit events.

glebcha:
Just a try without a chance to be approved. 
Thanks for a great tool!

katspaugh:
Thank you too! If you are solving a specific problem with the callback, let me know and I'm sure we can find a way to do the same using events.
Cheers!

PR #453, opened 2015-05-22
Added Zoom example to examples directory

maxehnert:
I have created an example of implementing a zoom with slider feature with wavesurfer. I copied the annotations example and appended my zoom slider. I used the annotations example because this will also work with annotations. They also redraw and scale correctly with the zoom. 

Please look over everything and make sure it's good to go. I built it out quickly in a separate application and copied the code over into the example/zoom directory.


katspaugh:
Hi Max!

I liked your simple and elegant solution so much that I added it as a new method `zoom` in cc2ad8f. Will update the readme soon.

Thanks!

maxehnert:
That's awesome! Glad I could give back to the project.

PR #454, opened 2015-05-25
Updating readme description for backend parameter

zschmid:
I noticed when testing fallback that setting AudioElement doesn't seem to do anything, but MediaElement does.


katspaugh:
Thanks!

PR #456, opened 2015-05-26
region update end is called when a region is created; callback now receives the region; updated readme

### LINKED ISSUE #382: region-update-end hook does not return affected region
it would be useful if the region-update-end hook would supply the region that was updated

### LINKED ISSUE #403: wavesurfer.regions when adding events should also set drag, resize and time
when creating a new region from dragging, during the bind of the mouse onUp/down part the drag, resize and time should be fetched from the region object. Without this the on-update-end event is not triggered during creation.

kakai248:
When creating a region by dragging, region-created is fired on the press of the mouse button and then region-updated is continuously called when the region is actually being created. But when we finish creating it, ie no more region-updated events, region-update-end wasn't being fired. This commit fixes this behaviour.

It also fixes #382. And I think it also fixes #403 but i'm not sure I understand what the problem really is on this one.


thijstriemstra:
Thanks. Could you add a comment explaining why it should be there and not inside the loop? I'm a fan of comments.

katspaugh:
@kakai248 this will fire `region-update-end` for every region on every click anywhere on the page.

kakai248:
`region-update-end` will indeed fire on other ocasions that I did not catch in my testing yesterday. I was just happy that it was working as I wanted in my application (due to the fact that I'm only allowing one region). I'm sorry.

On further analysis, `onDown` is not called when we create a new region by dragging because the element is not defined yet. However `onUp` is called when we release the button. Since `onDown` is not called neither `drag` or `resize` evaluate to true.

I suggest adding a create flag on the `Region`, where `init` sets it to true. `onUp` will be like this:

```javascript
var onUp = function (e) {
	if (drag || resize || my.create) {
		console.log("create " + my.create);
		drag = false;
		resize = false;
		my.create = false;
		e.stopPropagation();
		e.preventDefault();
		
		my.fireEvent('update-end', e);
		my.wavesurfer.fireEvent('region-update-end', my, e);
	}
};
```

It's not pretty, but it *seems* to be working. Tell me what you think.

katspaugh:
Ah, I see. Good that you found the reason.

Alternatively, we could fire the event from the `mouseup` event here: https://github.com/kakai248/wavesurfer.js/blob/master/plugin/wavesurfer.regions.js#L47

The `mousemove` event handler nearby in the code creates the region object, so it can fire the event as well.

kakai248:
That seems to work.

```javascript
this.wrapper.addEventListener('mouseup', function (e) {
	drag = false;
	
	if(region) {
		region.fireEvent('update-end', e);
		my.wavesurfer.fireEvent('region-update-end', region, e);
	}
	
	region = null;
});
```

katspaugh:
Yeah, looks good.

plugin/wavesurfer.regions.js
```diff
@@ -304,9 +310,9 @@ WaveSurfer.Region = {
                     resize = false;
                     e.stopPropagation();
                     e.preventDefault();
-
-                    my.fireEvent('update-end');
-                    my.wavesurfer.fireEvent('region-update-end');
+
```
katspaugh: Trailing whitespace.

plugin/wavesurfer.regions.js
```diff
@@ -304,9 +310,9 @@ WaveSurfer.Region = {
                     resize = false;
                     e.stopPropagation();
                     e.preventDefault();
-
-                    my.fireEvent('update-end');
-                    my.wavesurfer.fireEvent('region-update-end');
+					
+                    my.fireEvent('update-end', e);
+                    my.wavesurfer.fireEvent('region-update-end', my, e);
```
katspaugh: `my` → `region`

plugin/wavesurfer.regions.js
```diff
@@ -42,8 +42,14 @@ WaveSurfer.Regions = {
             start = my.wavesurfer.drawer.handleEvent(e);
             region = null;
         });
-        this.wrapper.addEventListener('mouseup', function () {
+        this.wrapper.addEventListener('mouseup', function (e) {
             drag = false;
+			
+            if(region) {
```
katspaugh: `if(` → `if (`

katspaugh:
Small whitespace comments. When you are finished, please squash commits into one if you can (if not, never mind).

kakai248:
Fixed the whitespaces and squashed the commits.

katspaugh:
Thanks!

PR #457, opened 2015-05-27
package.json js location causing bower/wiredep issues

zschmid:
…n js file

I'm not sure of the backstory here, but it looks like there has been some conversation around bower in the past. 

I'm noticing that when trying to use wiredep in grunt to automatically pull in bower_components, it fails for wavesurfer.js because main is not correctly set in package.json . 

Can we change main to what's proposed in the PR? Or if not, create a bower.json file that defines main as the .js in the dist folder? 

Currently have to use this workaround in our bower.json file for wiredep to work:

```
"overrides": {
    "wavesurfer.js": {
      "main": "dist/wavesurfer.min.js"
    }
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "wavesurfer.js",
   "version": "1.0.28",
   "description": "Interactive navigable audio visualization using Web Audio and Canvas",
-  "main": "dist/wavesurfer.js",
+  "main": "dist/wavesurfer.min.js",
   "directories": {
     "example": "example"
   },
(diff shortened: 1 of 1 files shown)


katspaugh:
I guess this PR can be merged safely and won't break any npm/browserify installations because the `dist/wavesurfer.cjs.js` file (the current `main`) doesn't even exist in the repo.

@thijstriemstra what do you think?

thijstriemstra:
:+1:  I also never understood why it was changed and glad someone put in a PR (and take responsibility for the change :tongue:

PR #462, opened 2015-06-08
Individual channel gain control

Kage1989:
Added connection and disconnection methods for bulk nodes connection/disconnection.
Also added a new page for stereo panning to resolve the odd panning in the panning example. I've noted that there is still some error in playing mono files. Stereo files are playing fine though.

I would like to check if you would only like restriction of insertion of nodes only to between gainNode and analyserNode.


example/panner/mainStereo.js
```diff
@@ -0,0 +1,121 @@
...
+
+        wavesurfer.backend.analyser.disconnect();
+
+        wavesurfer.backend.analyser.connect(wavesurfer.splitter);
+        for (var channel = 0; channel < wavesurfer.backend.ac.destination.channelCount; channel++) {
+			wavesurfer.channelVolumeNode.push(wavesurfer.backend.ac.createGain());
+			wavesurfer.splitter.connect(wavesurfer.channelVolumeNode[channel],channel);
+			wavesurfer.channelVolumeNode[channel].connect(wavesurfer.merger,0,channel);
+		};
+        wavesurfer.backend.connect(wavesurfer.merger,wavesurfer.backend.gainNode,wavesurfer.backend.ac.destination,'series');
```
Kage1989: This whole section above is horribly lengthy for connection. But it seems that this is only the case because I've to use the channelMerger and channelSplitter nodes. It'll look much cleaner after I use the connectionCore method for the connection between the splitter and merger.

example/panner/mainStereo.js
```diff
@@ -0,0 +1,121 @@
...
+        monoSlider.addEventListener('input', onMono);
+        monoSlider.addEventListener('change', onMono);
+        onMono();
+        wavesurfer.on('ready', function () {
+        	if (wavesurfer.backend.buffer.numberOfChannels == 1) {
+        		mono(-1);
+        		monoSlider.value = -1;
+        		monoSlider.disabled = true;
+        	};
+        });
```
Kage1989: Also added a mono function that's kind of missing if we're playing stereo files

src/webaudio.js
```diff
@@ -242,6 +242,128 @@ WaveSurfer.WebAudio = {
...
+			case 'miso':
+				/**
+				 * Connects the AudioNodes in the argument from multiple inputs to single output
+				 * 	@param arguments[0,...,N-2]: AudioNodes to be connected to destination
+				 * 	@param arguments[N-1]: destination AudioNode
+				 */
+				var dest = args.pop();
+				this.connectionCore(source,args,dest,mode);
+				break;
+		};
```
Kage1989: Just in case you wonder why I called it SIMO and MISO, [ SIMO = single input multiple output ] [ MISO = multiple input single output ]

katspaugh:
I was thinking we could just add a small method that takes 2 parameters: the 1st node (to connect to the source) and the last node (to connect to the destination).

That way you could create an arbitrary subgraph in the application code (no matter of which kind – parallel, serial) and insert it into the wavesurfer's graph.

---

Initially we have this:
```
[ wavesurfer source ] → [ wavesurfer destination ] 
```

In the app code one creates a parallel graph:

```
                            ↗︎  [ channel1 ] → [ gain ] ↘︎
[ channelSplitter ]                                               [ channelMerger ]
                            ↘︎
                               [ channel2 ] → [ gain ] ↗︎
```

Then inserts into wavesurfer like this (`wavesurfer.setGraph(channelSplitter, channelMerger)`:

```
[ wavesurfer source ] → [ channelSplitter ] → [ ... ] → [ channelMerger ] → [ wavesurfer destination ] 
```

Sorry for shitty charts :smile:

Kage1989:
Oh. We are not restricting them to insert their app audio nodes between the analyser node and gain node?

katspaugh:
We do. Do you think it may be a hindrance in some cases?

Kage1989:
I see. I was a little confused for a while when I saw your audio graph chart because there wasn't the analyser node and gain node. Ok. There's this github repo (http://ykifle.github.io/audiograph/) that enables users to draw audio nodes dynamically by dragging and dropping UI elements. I'll try and see if our arrangement could pose any restrictions to users.

Kage1989:
Instead of drawing arbitrary graphs myself, I searched online and realized that if there is only one source and one destination, then there isn't much issue with forcing the app to insert the nodes between analyser and gain node. However, if eventually, we're heading towards multiple sources (meaning playing multiple files simultaneously), we could hit a stone wall with this arrangement. One of the complex audio graph that I've seen is as shown below.
modular-routing2
So, I guess it's a decision that we've to make. Going for multiple sources would be awesome though. :)

katspaugh:
Going for multiple sources would mean refactoring of the whole `webaudio.js`, wouldn't it?

To be honest, I want to keep this project simple. After all, it's just a waveform player, not an all-around Web Audio wrapper providing doubtful convenience over direct Web Audio API. So I'd go for extensibility and staying out of the way, rather than trying to embrace all use cases.

src/webaudio.js
```diff
@@ -242,6 +242,134 @@ WaveSurfer.WebAudio = {
         this.createSource();
     },
 
+    setGraph: function (source, destination) {
+    	this.analyser.disconnect();
+    	this.analyser.connect(source);
+    	destination.connect(this.gainNode);
+    },
+
```
Kage1989: Right. Here's the method that will add user-defined audio nodes into the audio graph. :smile:

Kage1989:
This should be all for coding the panner node using the gainNode control on individual channel. Mono function is set to disabled for mono audio files.

katspaugh:
Closing, sorry. I think different types of nodes connections should be implemented in the application or plugin code. We should let users insert the first and the last node of their graph into the wavesurfer graph, that's all.

PR #466, opened 2015-06-12
Added id to the rendered region

berio:
I find very useful to add an id to the rendered region, so I can modify it by interacting with other DOM elements.


katspaugh:
As you see, TravisCI build failed because of double quotes.

berio:
Sorry about that. I have changed them to simple quotes.
Best!

katspaugh:
Thanks!

berio:
:D

PR #467, opened 2015-06-23
Added null checks to fix 'Uncaught TypeError: Cannot read property ..…

oroshnivskyy:
Added null checks to fix 'Uncaught TypeError: Cannot read property of null' errors when destroy() is called

### DIFF
--- a/src/mediaelement.js
+++ b/src/mediaelement.js
@@ -71,7 +71,7 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
     },
 
     getCurrentTime: function () {
-        return this.media.currentTime;
+        return this.media && this.media.currentTime;
     },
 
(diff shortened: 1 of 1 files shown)


src/mediaelement.js
```diff
@@ -71,7 +71,7 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
     },
 
     getCurrentTime: function () {
-        return this.media.currentTime;
+        return this.media && this.media.currentTime;
```
katspaugh: Is `getCurrentTime` called from `destroy`?

oroshnivskyy: It wass called from audioprocess handler that was handled from clearPlayEnd that was called from pause.

katspaugh: OK, thanks!

PR #469, opened 2015-06-24
Added check if media exists in isPaused function

oroshnivskyy:
Added check if media exists in isPaused mediaelement function, it's useful when we trying stop audio when it was destriyed or not set

### DIFF
--- a/src/mediaelement.js
+++ b/src/mediaelement.js
@@ -59,7 +59,7 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
     },
 
     isPaused: function () {
-        return this.media.paused;
+        return this.media && this.media.paused;
     },
 
     getDuration: function () {
--- a/src/mediaelement.js
+++ b/src/mediaelement.js
@@ -59,7 +59,7 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
     },
 
     isPaused: function () {
-        return this.media && this.media.paused;
+        return !this.media || this.media.paused;
     },
 
     getDuration: function () {


.idea/misc.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ProjectRootManager" version="2" />
+</project>
```
thijstriemstra: this shouldn't be part of the PR.

oroshnivskyy: Sorry, i updated PR.

src/mediaelement.js
```diff
@@ -59,7 +59,7 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
     },
 
     isPaused: function () {
-        return this.media.paused;
+        return this.media && this.media.paused;
```
oroshnivskyy: @katspaugh  Is it ok to return false when media not exists?

katspaugh: @oroshnivskyy I'm not even sure we are taking the right approach hiding all these possible errors in user's application. Why can't the user destroy the instance cleanly in their code? And not call `isPaused` when they know it's destroyed. Or is it called internally?

oroshnivskyy: @katspaugh It was called internally. How do you think if we will create [Null Object](https://sourcemaking.com/design_patterns/null_object) and set it to media when media it is destroued. Player will work like a cassete player when you turn it on without cassete.

katspaugh: If it's called internally, then I guess a check is enough. As for returning `false`, I would expect `isPaused` to return `true` if it's not playing.

katspaugh:
@oroshnivskyy can you update this PR so that `isPaused` should return `true` when there's no `media`?

oroshnivskyy:
@katspaugh yes sorry for delay

oroshnivskyy:
@katspaugh Hello i updated PR

katspaugh:
Thanks! Could you add spaces around the `||` operator? Like this: `!this.media || this.media.paused;`

oroshnivskyy:
@katspaugh Updated PR

katspaugh:
:ok_hand:

Issue #471, opened 2015-06-29
Pause event not fired when playing a range of audio


kjantzer:
```javascript
var wavesurfer = Object.create(WaveSurfer);

wavesurfer.init({
    container: '#wave',
    waveColor: 'violet',
    progressColor: 'purple'
});

wavesurfer.load('example/media/demo.wav');

wavesurfer.play(); // play event fired
wavesurfer.pause(); // pause event fired

// will play the first 10 seconds of the audio clip
wavesurfer.play(0,10); // play event fired, but NO pause event when finished
```

The issue lies with how the `backend` plays a range of audio.

katspaugh:
Kevin, can you verify that your fix is applied and resolves this issue?

kjantzer:
@katspaugh Looks good!

katspaugh:
Cool, thanks!

Issue #398, opened 2015-03-13
Region's play() does not fire pause/finish event after whole region is played


AndKiel:
I'm using wavesurfer together with html video tag and trying to achieve full sync of video with waveform.
Unfortunately when I use region's `play()` function, it's impossible to stop video from playing because it does not fire any events beside `region-play` (undocumented in readme, found in code) and `play`.
Could backend fire `pause` or `finish` after whole region is played?
Also when region starts playing it does not fire `seek` event when setting audio position to region's `start` value.

katspaugh:
Hi Andrzej,

There's a region event called `region-out`. Both `region-play` and `region-out` events are listed in the README under the Region events section.

AndKiel:
`region-play` is not listed in README.md file.
But `region-out` works for me, thank you very much. I assumed that `region-out` will fire even with normal playback, not only with `region.play()` function, my bad.

AndKiel:
`region-play` is *not* listed in README.
And `region-out` does not work for me. It fires even when normal playback leaves region, not only when clicked region finished playing (`region.play()` plays specific part from beginning to end and after it finished there is no `pause` event). So right now for workaround I would probably have to bind my `region-out` callback on `region-click` and unbind on `region-out`.

katspaugh:
You are right, thanks for clarifying. So the following should be done:

 * [ ] Region should fire the global `seek` event when it starts playing from start to end.
 * [ ] Region should fire `region-pause` when it finishes playing from start to end.
 * [ ] Document `region-play` in the README. Also make it clearer that region events correspond to global events prefixed with `region-`.

AndKiel:
Yes.
But probably backend* should fire global events instead of region having its own.
Also it appears that when we use wavesurfer's `play` after region's `play`, it rewinds back to the region's start time and plays from there (shouldn't it start playing when it last stopped?).

*MediaElement and WebAudio - they are the ones with `seekTo`, `play` and `pause` not firing events, region just uses wavesurfer's `play` which uses backend's `play` (so it's probably impossible for region to have its own `region-pause` event, but global `play` and `pause` should be fine).

kjantzer:
I'm looking for this feature as well. Seems inconsistent that when the region stops/pauses that the global `pause` event is not fired since the audio is indeed pausing.

katspaugh:
I'll raise the priority of this issue. A PR would also be welcomed.

kjantzer:
[...]

PR #473, opened 2015-06-30
Fixes inconsistent events with backend when playing ranges/regions

kjantzer:
Addresses #398 and #471: backend player not fires events for play and pause and wavesurfer forwards these events on; region now properly fires it's `out` event


katspaugh:
Sorry, `Math.round(time*100)/100 ` isn't a good solution. It will make other cases fail.

kjantzer:
@katspaugh – Which cases?

katspaugh:
Cases where precision matters.

Besides, I don't get why in your case it wasn't firing. The original code was like this:

`if (!my.firedOut && my.firedIn && my.end <= time) {`

So if `audioprocess` gives you `19.9999` at some point, and your end mark is `20`, `region-out` would still be fired in the next frame (when the time would be, say, 20.00000001`), wouldn't it?

The rest of the PR (moving `play`, `pause` and `finished` events to the audio backends) was good, I'd merge these changes if you would be kind enough to make a separate PR (don't include the build, I'll update the version and rebuild anyway).

kjantzer:
@katspaugh - that was the problem, the time stopped at 19.999 and never reached 20.000001. I tested regions in the very beginning and at the end and neither fired.

From my testing, `Math.round(time*100)/100` is still quite accurate. For example:

`19.994` = `19.99`
`19.995` = `20`

So **at most five-thousandths (.005) of a second is lost**. I can't imagine how this would be a problem when the function is just calling `out` events:

```
my.fireEvent('out');
my.wavesurfer.fireEvent('region-out', my);
```

To me, this would be better than not having the events fire at all, but its your call.

If you still don't think the round is a good thing, I can make another PR without that part.

katspaugh:
I may be wrong. Let's push it.

katspaugh:
@kjantzer I've merged your patch manually, thanks!

PR #477, opened 2015-07-07
Resize minimap when browser resize occurs

dbunker:


dbunker:
Example of issue can be seen [here](http://www.wavesurfer.fm/example/annotation/). First, reload the tab when the browser is small, then expand the browser to full size. You can see the minimap remains at the original size. This pull checks when a resize occurred based on the window resize event and the container size, then re-renders if needed.

plugin/wavesurfer.minimap.js
```diff
@@ -117,6 +117,14 @@ WaveSurfer.Minimap = WaveSurfer.util.extend({}, WaveSurfer.Drawer, WaveSurfer.Dr
         }
 
         this.wavesurfer.on('destroy', this.destroy.bind(this));
+
+        my.check_width = 0;
```
katspaugh: Please change this to `var prevWidth = 0;`.

plugin/wavesurfer.minimap.js
```diff
@@ -117,6 +117,14 @@ WaveSurfer.Minimap = WaveSurfer.util.extend({}, WaveSurfer.Drawer, WaveSurfer.Dr
...
+        window.addEventListener('resize', function() {
```
katspaugh: Please add a space after `function`.

katspaugh: You need to also unsubscribe from the resize event on destroy.

```
this.on('destroy', function () {
    window.removeEventListener(onResize);
});
```

katspaugh:
Mörged! :metal:

Issue #449, opened 2015-05-07
Option to Draw Raw Waveform (which may be asymmetric)


nickarmstrongcrews:
In order to annotate certain signals (really, accelerometer data masquerading as .wav audio files), I need to be able to see if the signal transitions from high to low or low to high.

As I understand it, this is currently impossible with current framework, but I'm pretty sure a simple option to init() and a slight change to drawing code could provide this functionality.

Happy to write / test it, if you give me a hint as to where to look!

Thanks!

katspaugh:
If I understand right, you want to achieve image like this:

![raw waveform](http://web.mit.edu/9.29/www/neville_jen/hmm/MEA_Report_files/image002.jpg)

You would need to change one or both of these functions:

 * [`getPeaks`](https://github.com/katspaugh/wavesurfer.js/blob/a96738c6a1112252fe64289643ab3806a40643e2/src/webaudio.js#L180) – extracts sampled maximum values from raw PCM data.

 * [`drawWave`](https://github.com/katspaugh/wavesurfer.js/blob/a96738c6a1112252fe64289643ab3806a40643e2/src/drawer.canvas.js#L68`) – the function that draws the waveform on Canvas.

nickarmstrongcrews:
Great, thanks! Have been playing with these.

Current struggle is that library seems to assume 44 kHz sampling rate...
resulting in drawing a width zero image for my ~100 Hz signal. Can you
suggest best place to make this change?

Thanks!

katspaugh:
@nickarmstrongcrews the library isn't setting the sample rate anywhere. There might be an issue with low sample rates in Web Audio implementations (see a similar issue #353).

PR #478, opened 2015-07-09
Canvas waveforms show true min and max peaks

dpwe:
This allows waveforms to be asymmetric around zero, and to see the actual wave shape when you are zoomed in far enough (or if the wave shape has low enough frequency). 

See the comparison screenshot of a zoomed-in rumble.  Top is current code, bottom is with this patch.  

This resolves https://github.com/katspaugh/wavesurfer.js/issues/449 .

waveform_comparison


katspaugh:
Thanks Dan! Sorry for the delay!

alimatinfar:
hi 
i want to draw my wav file In the form of bottom picture but drawn in the form of top picture
  
image

What changes should I make?

PR #496, opened 2015-08-04
Added support for maxWidth and minWidth options for regions

gapipro:


katspaugh:
Why do you want them? What's the use case?

gapipro:
When you are using wavesurfer for something like audio trimming. Region indicates are that will remain. And in that case you can limit user to minimum and maximum length of trimmed audio.

plugin/wavesurfer.regions.js
```diff
@@ -89,6 +89,9 @@ WaveSurfer.Region = {
         this.color = params.color || 'rgba(0, 0, 0, 0.1)';
         this.data = params.data || {};
 
+        this.maxWidth = params.maxWidth === undefined ? null : Number(params.maxWidth);
+        this.minWidth = params.minWidth === undefined ? null : Number(params.minWidth);
```
katspaugh: * Rename `*Width` to `*Length`.
 * Just `this.*Length = params.*Length` would be enough.

plugin/wavesurfer.regions.js
```diff
@@ -210,6 +220,15 @@ WaveSurfer.Region = {
           this.end = dur;
           this.start = dur - (this.end - this.start);
         }
+
+        if (this.maxWidth && this.end - this.start > this.maxWidth){
```
katspaugh:
```
if (this.minLength != null) {
    this.end = Math.max(this.start + this.minLength, this.end);
}

if (this.maxLength != null) {
    this.end = Math.min(this.start + this.maxLength, this.end);
}
```

katspaugh:
Merged, thanks!

Issue #495, opened 2015-08-03
Support for only region part of waveform to be selected with darker color


gapipro:
So it would like to have something like this:

http://postimg.org/image/mz0llz6ab/

katspaugh:
That's exactly what the regions plugin does.

gapipro:
Actually it doesn't. Check your coloring and my image example. Difference is that only selected part of waveform is darker not the whole waveform from 0 to selected region.

katspaugh:
@gapipro I see what you mean. I've checked your PR again and still think it's a bit of an overkill in terms of how much code is needed for a very particular feature. Thanks for your contribution, anyway!

PR #501, opened 2015-08-10
Added support for waveform region highlighting 

gapipro:
When enabled only waveform inside region is highlighted.


katspaugh:
Why can't you just use a darker color for the region's background?

gapipro:
Because that is not the desired look of the waveform. 

Check image at: https://github.com/katspaugh/wavesurfer.js/issues/495

katspaugh:
Sorry, I don't want to merge this. The feature is too specific.

Issue #529, opened 2015-09-21
Examples use outdated links


thijstriemstra:
Most example index pages still refer to http://wavesurfer.fm (which seems to be dead). The correct link is http://wavesurfer-js.org

thijstriemstra:
```
$ grep -Rn 'wavesurfer.fm' .
```

```
./example/angular/index.html:116:            ga('create', 'UA-50026819-1', 'wavesurfer.fm');
./example/annotation/index.html:129:            ga('create', 'UA-50026819-1', 'wavesurfer.fm');
./example/audio-element/index.html:131:            ga('create', 'UA-50026819-1', 'wavesurfer.fm');
./example/elan/index.html:41:                <h1 itemprop="name"><a href="http://wavesurfer.fm">wavesurfer.js</a><noindex> + <a rel="nofollow" href="http://spokencorpora.ru/showelan.py">ELAN</a></noindex></h1>
./example/elan/index.html:101:            ga('create', 'UA-50026819-1', 'wavesurfer.fm');
./example/equalizer/index.html:88:            ga('create', 'UA-50026819-1', 'wavesurfer.fm');
./example/microphone/index.html:28:                <h1 itemprop="name"><a href="http://wavesurfer.fm">wavesurfer.js</a><noindex> + Microphone</noindex></h1>
./example/microphone/index.html:191:            ga('create', 'UA-50026819-1', 'wavesurfer.fm');
./example/panner/index.html:164:            ga('create', 'UA-50026819-1', 'wavesurfer.fm');
./example/playlist/index.html:108:            ga('create', 'UA-50026819-1', 'wavesurfer.fm');
./example/spectrogram/index.html:36:                <h1 itemprop="name"><a href="http://wavesurfer.fm">wavesurfer.js</a><noindex> + Spectrogram</noindex></h1>
./example/spectrogram/index.html:139:            ga('create', 'UA-50026819-1', 'wavesurfer.fm');
./example/split-channels/index.html:104:            ga('create', 'UA-50026819-1', 'wavesurfer.fm');
./example/timeline/index.html:38:                <h1 itemprop="name"><a href="http://wavesurfer.fm">wavesurfer.js</a><noindex> + Timeline</noindex></h1>
./example/timeline/index.html:147:            ga('create', 'UA-50026819-1', 'wavesurfer.fm');
./example/zoom/index.html:118:            ga('create', 'UA-50026819-1', 'wavesurfer.fm');
[...]

PR #530, opened 2015-09-22
Fix broken links and update bootstrap in examples

thijstriemstra:
refs #529


katspaugh:
Thanks! :pray:

katspaugh:
By the way, if I knew earlier that https://dns.js.org/ existed, I would go for `wavesurfer.js.org`.

thijstriemstra:
Ah, well next time please open a ticket so we can get a heads up when you intend to change the domain because it went unnoticed until the examples I use broke due to a link to wavesurfer.fm/dist/ etc.

PR #534, opened 2015-10-02
Updating README formatting and resource links

CWSites:
I noticed that **Mouse Events** wasn't formatted correctly under **Region Events**. I also added a link to the browser support for Web Audio to http://caniuse.com/audio-api


katspaugh:
👍

PR #545, opened 2015-10-08
use protocol relative urls for https in examples and documentation

thijstriemstra:
All examples now also work on https (http://katspaugh.github.io/wavesurfer.js/ for example).

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,7 +254,7 @@ thanks to
 
 ## License
 
-![cc-by](http://i.creativecommons.org/l/by/3.0/88x31.png)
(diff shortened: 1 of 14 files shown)


thijstriemstra:
Also, linking to `//wavesurfer-js.org/dist/wavesurfer.min.js` is not option because the https certificate for that domain isn't matching. I guess this will cost some money to get a proper certificate for that domain?

katspaugh:
Thanks!

Hot-linking to a project homepage isn't a good idea. People should use proper CDNs instead, e.g. https://cdnjs.com/libraries/wavesurfer.js

Issue #339, opened 2015-01-29
New region creation - getDuration() ?


sg3510:
When creating a new object after initialisation, with this:
```javascript
au_opt = {
	start: 0.704,
	end: 1.018 };
wavesurfer.addRegion(au_opt);
```

The region will fail to initialise correctly as in update render
```javascript
var dur = this.wavesurfer.getDuration();
```
will return 0 and cause start = -0.314 and end = 0 as (below) resolves as true
```javascript
 if (this.end > dur) {
    this.end = dur;
    this.start = dur - seconds;
}
```

I can succesfully create regions using this same code above  - is this an issue with the getDuration() or the way I am using it?

Full code below:
```javascript
'use strict';

// Create an instance
var wavesurfer = Object.create(WaveSurfer);
var au_opt = {};

// Init & load audio file
document.addEventListener('DOMContentLoaded', function () {
	var region_count = 0;
    var options = {
        container     : document.querySelector('#waveform'),
        waveColor     : 'violet',
        progressColor : 'purple',
        loaderColor   : 'purple',
        cursorColor   : 'navy',
        normalize : true
    };
    // Init
    wavesurfer.init(options);
    // Load audio from URL
    wavesurfer.load('{{audio_file}}');

    wavesurfer.enableDragSelection({
        color: 'rgba(0, 255, 0, 0.2)',
        loop: true
    });
	wavesurfer.on('region-created', function (region, e) {
               // Limit to one region hacky way
		if (region_count >= 1) {
			region_count = 0;
			au_opt = {
				start: 	region.start,
				end: 	region.end
			};
			wavesurfer.clearRegions();
			wavesurfer.addRegion(au_opt); //THIS WORKS

		}
		console.log(region);
		region_count += 1;
	});

	au_opt = {
	   start: 0.4,
	   end: 3.2 };
	wavesurfer.addRegion(au_opt); // THIS DOESNT WORK

	wavesurfer.on('ready', function () {
		wavesurfer.play();
	});
});
```

katspaugh:
You can't add a region until the audio is loaded. So, add the first region inside the `ready` event callback.

sg3510:
Ah yes, makes sense and it works - thanks!

PR #546, opened 2015-10-08
Updated README.md

russianryebread:
Added some clarification on the addition of regions, as answered in the issue: https://github.com/katspaugh/wavesurfer.js/issues/339

I didn't find any direct documentation on this, and thought it might save some time, as I ran into this issue, and noticed that I wasn't the only one.


katspaugh:
Thanks!

PR #549, opened 2015-10-14
add support for touch devices

jonykalavera:
adds drag support for touch devices for the Regions Plug in


plugin/wavesurfer.regions.js
```diff
@@ -37,12 +37,17 @@ WaveSurfer.Regions = {
         var start;
         var region;
 
-        this.wrapper.addEventListener('mousedown', function (e) {
+        function eventDown(e) {
             drag = true;
+            if(typeof e.targetTouches != 'undefined'){
```
katspaugh: `if (e.targetTouches) {` (needs spaces).

katspaugh: I think we also need to check if it's a single touch, otherwise we would prevent pinch zoom and other multitouch gestures from working properly.

plugin/wavesurfer.regions.js
```diff
@@ -60,12 +67,17 @@ WaveSurfer.Regions = {
             }
 
             var duration = my.wavesurfer.getDuration();
+            if(typeof e.targetTouches != 'undefined'){
```
katspaugh: The same.

plugin/wavesurfer.regions.js
jonykalavera: will do

plugin/wavesurfer.regions.js
```diff
@@ -60,12 +67,17 @@ WaveSurfer.Regions = {
             }
 
             var duration = my.wavesurfer.getDuration();
+            if (typeof e.targetTouches !== 'undefined' && e.targetTouches.length === 1) {
```
katspaugh: Here we could also check if the id of the ended touch is the same as the started touch, but I think it's good enough for now.

katspaugh:
Merged as a0464534070c4e6aaaad07883c98b89f114f41f5.

PR #552, opened 2015-10-16
add a parameter called attributes to the regions

rikrd:
the function of this parameter is to add attributes to the underlying HTML element to allow declarative style modifications using CSS


katspaugh:
Interesting, thanks!

PR #561, opened 2015-10-27
Add zoom event

rozap:
I noticed that on zooming the regions do not redraw themselves, which results in them being displayed incorrectly. 

* Make the region call updateRender on zoom so its
rendering matches the wavesurfer's scale
* update readme


katspaugh:
Great, thanks!

PR #564, opened 2015-10-29
Add autoCenter option

rozap:
This seems like a common use case. At least for the thing I'm building, it's pretty distracting to have the waveform move while playing, as I'm often zoomed in really far, so it would be nice to disable it.

The default behavior is the same as the old behavior. 

Thanks for the great library btw


katspaugh:
Thanks!

Issue #429, opened 2015-04-10
Ready event not triggered at the right time


pkeuter:
First off: what a nice tool this is. 

I do have a little problem with events though. It looks like the ready event not triggers like it says on the site:  "When audio is loaded, decoded and the waveform drawn." It triggers after I start play playback. Maybe I'm doing something wrong here, could be. I'm using it inside of AngularJS

thijstriemstra:
Can you reproduce it with the AngularJS example?

katspaugh:
@pkeuter the example Thijs is referring to is this: https://github.com/katspaugh/wavesurfer.js/blob/master/example/angular/app.js

Make sure your AngularJS directive isn't turned off with `ng-if` until you press the play button.

pkeuter:
I can reproduce it with the example. http://plnkr.co/edit/LRKeAglgUxwHBLzW1veN?p=preview

As you can see, the text only dissapears when you press play.

liamdv4:
I'm using MediaElement mode (without autoplay) and loading an hour-long MP3 which takes a little while. The event fires for me as soon as the player instantiates (before I press play). Again this is not quite how the event is described because it takes 10-15s before the waveform is drawn.

I was hoping to be able to implement a spinner or "Generating waveform..." message during that time which I could remove when the "ready" event fired.

katspaugh:
@liamdv4 thanks, looks like a bug.

liamdv4:
I think it would be ideal to leave the ready event as it is (since the player is, indeed, ready to play at that point), but add a new event for when the waveform has been drawn. Thanks.

katspaugh:
Yes, makes sense.

hungryforcodes:
+1 for new event. I had this problem with "ready" as well, but have since used it to know when the player is ready as suggested.

chinshr:
@pkeuter check if this #565 works for you?

katspaugh:
Solved in ##856.

PR #565, opened 2015-10-29
Configure autoplay for mediaelment

chinshr:
Enable autoplay for backend "mediaelement". Background is the `canplay` event on an `<audio>` element, which is mapped to `ready` in wavesurfer is not triggered automatically on iOS. The [workaround](http://stackoverflow.com/questions/9920297/html5-audio-canplay-event-doesnt-fire-on-safari-mac-desktop) is to set `autoplay` to `true` and pause the audio as soon as the `ready` event is fired.

Note: This could provide a workaround for #429.

```javascript
wavesurfer.init({autoplay: true});
wavesurfer.on('ready', function() { wavesurfer.pause(); })
```

### DIFF
--- a/src/mediaelement.js
+++ b/src/mediaelement.js
@@ -25,7 +25,7 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
 
         var media = document.createElement(this.mediaType);
         media.controls = this.params.mediaControls;
-        media.autoplay = false;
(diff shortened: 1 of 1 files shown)


katspaugh:
Have you checked if it works on iOS?

chinshr:
Yes, tested and works for me.

katspaugh:
Cool, thanks a lot!

Issue #425, opened 2015-04-02
region-out not triggered when seeking out of current region


stromnet:
Hi,

first of all, thanks for wavesurfer, it's awesome!
I'm using 1.0.24, and I'm having some issue with region-in/out and seeking:
1. playing inside a region (region-in has fired earlier)
2. seeking to somewhere before this region
   - Expected: region-out is fired
   - Actual: nothing.

Seeking to somewhere BEHIND the current region end seems to work better though..
1. playing inside a region  (region-in has fired earlier)
2. region end is reached (region-out is fired). playing continues for a second
3. seek to a second before the region starts
   - Expected: region-in is triggered again
   - Actual: nothing

I currently do not have any standalone test-case, but it should be quite easy to reproduce.

PR #574, opened 2015-11-06
Fixes issue #425 where region-out is not triggered when seeking out of current region

jbeyer05:


plugin/wavesurfer.regions.js
```diff
@@ -269,19 +269,22 @@ WaveSurfer.Region = {
     bindInOut: function () {
         var my = this;
 
+        my.firedIn = false;
+        my.firedOut = false;
+
         var onPlay = function () {
-            my.firedIn = false;
-            my.firedOut = false;
```
katspaugh: So `onPlay` is now empty? Why keep it?

jbeyer05: Would probably be cleaner with onPlay not there.  For those users who have custom code in onPlay, but are trying to stay current with the latest version of wavesurfer, it's easier for them if it's left there.
You want me to remove and re-submit the PR?

katspaugh: You don't have to re-submit the PR, just remove the function and push to the same branch, and this PR will be updated.

jbeyer05: done

katspaugh:
Thanks!

PR #583, opened 2015-11-11
Replace depreciated AudioElement with MediaElement in examples

SeanBannister:
There were a few cases where AudioElement was still being used in the example code.


katspaugh:
Thanks!

PR #586, opened 2015-11-19
Added New Example using Angular Material

logbon72:
I just added a new example using Angular Material.


katspaugh:
Very nice, thanks! Could you squash your commits into 1?

PR #593, opened 2015-12-03
Updation to chrome >=45 stop function deprecated resolved

hardeepmehta:


katspaugh:
Why not just check for existence of `getTracks`?

hardeepmehta:
Yes, you are right. It's the better way.

thijstriemstra:
Can the Chrome-related code be moved to wavesurfer.util, I'd rather not have it in the plugin and that way it can be re-used in other plugins.

plugin/wavesurfer.microphone.js
```diff
@@ -122,8 +122,32 @@
             this.disconnect();
 
             // stop stream from device
-            if (this.stream) {
-                this.stream.stop();
+<<<<<<< HEAD
```
katspaugh: Looks like a failed merge.

plugin/wavesurfer.microphone.js
```diff
@@ -122,8 +122,32 @@
...
+            if (this.stream && this.stream.getTracks()) {
```
katspaugh: The check of the stream being not null should be outside of the `getTracks` change.

```
if (this.stream) {
    if (this.stream.getTracks) { // note that this should not be a call
        this.stream.getTracks().forEach(function (stream) {
            stream.stop();
        });
    } else {
        this.stream.stop():
    }
}
```

plugin/wavesurfer.microphone.js
```diff
@@ -122,8 +122,32 @@
...
+                
+                this.stream.getTracks().forEach(function( stream ) {
```
katspaugh: Wrong spaces:
`function( stream ) {` → `function (stream) {`

thijstriemstra:
See #657

Issue #598, opened 2015-12-11
clearRegions() after zoom causes error 


nabiulin:
Hi. 
Using zoom and region plugin simultaneously and trying to redraw regions after zooming.
In that case get error "TypeError: Cannot read property 'style' of null" in drawer.js:82:19 
```javascript
80.    style: function (el, styles) {
81.        Object.keys(styles).forEach(function (prop) {
82.            if (el.style[prop] !== styles[prop]) {
83.                el.style[prop] = styles[prop];
84.            }
85.        });
86.        return el;
86.    }
```
Any idea how to fix it?

nabiulin:
Fix regarding issue I found https://github.com/katspaugh/wavesurfer.js/issues/598.
Please take a look on it. Not sure that this is the best solution but it work for me

PR #600, opened 2015-12-14
Fixed region updation error after zoom

nabiulin:
Fix regarding issue I found https://github.com/katspaugh/wavesurfer.js/issues/598.
Please take a look at it. Not sure that this is the best solution but it works for me


katspaugh:
Nice, thanks!

PR #602, opened 2015-12-14
Migrate from legacy to container-based Travis infrastructure

thijstriemstra:
See https://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade


katspaugh:
Thanks!

Issue #608, opened 2016-01-06
Bring Back CommonJS Support?


skorlir:
It looks like WaveSurfer still has an NPM module, but there's (seemingly) no longer a CommonJS module being exported in the code. Which makes me sad, since I'm using Browserify and it would be supremely convenient to be able to `require` wavesurfer.js from NPM.

Is there a reason why #139 got rolled back? It looks like you made a move from Makefile to Grunt, so could that be why?

If there's anything I can contribute to make this possible, let me know.

katspaugh:
There's a task in the grunt file to build a CommonJS package: https://github.com/katspaugh/wavesurfer.js/blob/master/Gruntfile.js#L40

I'm not sure why we stopped saving it into the repository. Let's add it back.

skorlir:
:+1:

Awesome, thanks!

thijstriemstra:
@skorlir yes thanks! This would also eliminate the 'amd build' target in the grunt file. We should also look at how this affects the plugins, perhaps they should also need a universal umd/amd etc header.

skorlir:
As best I can tell from my most recent `npm i wavesurfer.js` install (just now), `wavesurfer.cjs.js` is tracked, but the package.json `main` entry isn't right for NPM to resolve the module.

That's all I have so far. https://github.com/skorlir/wavesurfer.js

Plugins are currently inconsistent. `wavesurfer.microphone.js` and `wavesurfer.timeline.js` have an AMD header, but not UMD/CJS. `wavesurfer.minimap.js` is written entirely inside of a single call to `WaveSurfer.util.extend`, and then directly modifies the `WaveSurfer` object at the end. The others add fields to the `WaveSurfer` object directly, then use `WaveSurfer.util.extend` to implement `WaveSurfer.Observer`. I don't quite understand everything going on with plugins.

In a Node/CommonJS context, it makes sense to me that WaveSurfer plugins would be registered programmatically. `WaveSurfer.registerPlugin(require('wavesurfer.js-microphone-plugin'))`, or some such. The existing system could be adapted without such a major refactor, I am sure.

[This](https://github.com/umdjs/umd/blob/master/templates/returnExports.js) appears to me to be the relevant UMD approach for adapting plugins. I'm looking into that now.

mspae:
+1 for this issue. I'm currently working on a wrapper for React (https://github.com/mspae/react-wavesurfer) but it's a bit difficult without proper UMD support.

[...]

PR #613, opened 2016-01-12
Add Back CommonJS Support via grunt-umd

skorlir:
Adds back CommonJS per #608 via UMD.

Most of the work is done by grunt-umd. Run the task with `grunt umd:main`

This has been put through a very basic browserify test, and it holds up!

There's another branch for plugin umd wrapping, but I don't feel 100% finished. Will make a separate PR for review in case it's not what we want.


katspaugh:
Many thanks, merged!

PR #614, opened 2016-01-12
Plugin umd

skorlir:
Builds on the last PR for #608 by adding grunt umd tasks for wrapping plugins in umd.

Run the task with `grunt umd:plugins`

It's not very well configured. It may be desirable to manually configure the settings for the umd build for each plugin, although this feels inefficient.

Right now the AMD modules are anonymous, which is a clear net loss from the prior hard-coded AMD wrappers on a couple of the plugins.


skorlir:
Whoops. Looks like `uglify` fails on the travis build. I'll look into it tomorrow.

katspaugh:
Looks like `cjs.js` was changed to `min.js` in #457. So it's either bower or browserify but not both?

skorlir:
Huh. Looks like it.

That PR is less than a year old, but I keep hearing in the back of my mind "Bower is deprecated."

That's not quite the whole truth, but:
https://github.com/bower/bower/issues/1520

And see the most recent comment here:
https://github.com/bower/bower/issues/1626

Bower is shifting toward a merger with NPM. Webpack + Browserify use the NPM/CommonJS module system for managing dependencies for front-end projects, and they have exploded in popularity. (See the mention of React in #608 - React is very often used with Webpack. React popularity = Webpack popularity.)

Having a separate bower.json could alleviate this? But, imho, bower is going the way of the dodo bird. Ending bower support is not a big loss.

katspaugh:
Oh, we do have a separate `bower.json` file: https://github.com/katspaugh/wavesurfer.js/blob/master/bower.json – so I guess we are good.

skorlir:
Did not see that before posting all that about bower. :)

Well, cool! Anything else? Should plugin modules be configured differently / is it okay for the AMD module to be anonymous? 

As far as I can tell the plugin modules are all placed on the WaveSurfer object, so they are accessible either way. 

Also, should we remove amd and cjs build targets?

mspae:
:clap:

I tried it with webpack just now. One issue: requiring a plugin (timeline) and bundling returns an error:

```
ERROR in ../~/wavesurfer.js/plugin/wavesurfer.timeline.js
Module not found: Error: Cannot resolve module 'wavesurfer' in /home/martin/Dropbox/Sandbox/wavesurfer-webpack-test/node_modules/wavesurfer.js/plugin
 @ ../~/wavesurfer.js/plugin/wavesurfer.timeline.js 3:8-39
```

This can be fixed in two ways:
* Define an alias in webpack.config.js (`wavesurfer` resolves as `wavesurfer.js`)
* Change the UMD header in `wavsurfer.timeline.js` like so:
```javascript
    // ...
    if (typeof define === 'function' && define.amd) {
        define(['wavesurfer.js'], factory);
    // ...
```

Is there a particular reason why we call the dependency `wavesurfer` without the `.js` – is this maybe the way AMD works and I've overlooked something, or should we rename the dependency to fit the npm package name so it works out of box with webpack?

skorlir:
I'm all for making the AMD module ID consistent with the NPM/CJS module name, but I kept the names separate because that's how they were before. As far as I can tell, there's not any reason the AMD module would have a different name - except for backwards compatibility.

mspae:
I'm all for making the change. If the change is reflected in semver people will be able to see it and it probably wont be such a huge change for people's build process if they do want to roll with it. I think the positives outweigh the possible negatives. But the maintainers need to decide.

thijstriemstra:
I use the AMD version and don't see why we should change this and break a lot of code out there. It looks like `wavesurfer.js` for AMD isn't an option, I ran into this issue before where video.js couldn't be mapped with r.js, the period wasn't allowed or confused the compiler. Using `wavesurfer` for UMD/Node and `wavesurfer` for AMD sounds fine, video.js uses this as well for example.

I hope eventually we can get rid of the wavesurfer.amd.js and wavesurfer.umd.js, why not have a single wavesurfer.min.js??

mspae:
Oh dear. that is a pretty good reason not to change it. Maybe we can document that webpack needs an alias (When the PR has been merged in)

katspaugh:
So do I get it right, that we can merge this PR and it won't break anything?

And that webpack works with our modules with a special config option?

dist/plugin/wavesurfer.elan.js
```diff
@@ -0,0 +1,271 @@
...
+    });
+  } else if (typeof exports === 'object') {
+    // Node. Does not work with strict CommonJS, but
+    // only CommonJS-like environments that support module.exports,
+    // like Node.
+    module.exports = factory(require("wavesurfer.js"));
+  } else {
+    factory(WaveSurfer);
+  }
+}(this, function (wavesurfer.js) {
```
skorlir: You can't call a variable `wavesurfer.js`

All the plugin builds have this issue

skorlir:
We should most definitely not merge this just yet. I just looked over it again, and the plugin task builds invalid code (namely, it calls the variable inside of the UMD callback `wavesurfer.js`, which is invalid).

As a sidebar, according to a prior comment, you can't change the AMD module name because AMD doesn't like '.'s in its modules. Thus, the AMD module name remains 'wavesurfer', even with this change.

skorlir:
Okay; now I feel fairly confident that this can be merged without issue.

The AMD module is anonymous. It is unchanged; still 'wavesurfer' without the '.js'.
Plugins can be build with UMD via `grunt umd:plugins` and they'll end up in `dist/plugin` with a regular `.js` extension. And the closures use valid parameter variable names.

katspaugh:
Cool, thanks! Can you update the branch for merge?

mspae:
awesome :ok_hand:, thanks!

skorlir:
I think I did update the branch, @katspaugh ? Is there anything else I can/should do?

thijstriemstra:
@skorlir maybe creating a new PR solves the issue.

skorlir:
Okay, I did a quick rebase with the current master and fixed everything. I also modified the `uglify:plugins` task to minify the UMD-wrapped versions of plugins. Should be ready for merge.

skorlir:
As someone mentioned earlier in this thread, now that we have a UMD build we really don't need cjs and amd builds anymore. So I have removed them. I also added .gitignore entries for the intermediary umd build files for both wavesurfer main and for plugins, and stopped tracking `wavesurfer.umd.js` and set the package.json `main` entry to point to the new single-source-of-truth, UMD-wrapped `wavesurfer.min.js`.

Let me know if I'm overstepping. Compatibility should not change at all with these new commits. In fact, everything should actually be simpler now with single UMD-wrapped builds of all of the WaveSurfer ecosystem.

katspaugh:
Wonderful! Thank you so much!

thijstriemstra:
@skorlir when I run `grunt` files with the extension `.min.min.js` are created:

```
	dist/plugin/wavesurfer.elan.min.min.js
	dist/plugin/wavesurfer.microphone.min.min.js
	dist/plugin/wavesurfer.minimap.min.min.js
	dist/plugin/wavesurfer.regions.min.min.js
	dist/plugin/wavesurfer.spectrogram.min.min.js
	dist/plugin/wavesurfer.timeline.min.min.js
```
Are you seeing this as well? It also looks like the minified microphone plugin contains two UMD headers (one from the src file, and one appended by grunt-umd).

thijstriemstra:
Ehm, when I change a line in the microphone plugin src file for example, re-run grunt, there's nothing changed at all in the dist version (other than the timestamp).....

skorlir:
@thijstriemstra mmm, that's happening because there are already wavesurfer.*.min.js files. The glob doesn't consider that. My thought, actually, would be to write a `clean` task that deletes `dist` every time you run the default `grunt` tasks, since you're replacing everything anyway.

As for changes not propagating, that doesn't make sense to me. I definitely changed the plugins and the build correctly created changed files.

thijstriemstra:
Thanks for the update @skorlir. I actually take the previous comment back, I modified Gruntfile.js a little to prevent the .min.min.js files but that wasn't a good fix

thijstriemstra:
@skorlir also, have you seen the wavesurfer.umd.js file that's generated in dist? Can we get rid of this too? It contains some random lines...

skorlir:
It's generated, but it's ignored in .gitignore. It's just an intermediate file. We can delete it, but if we keep it, you'll always have a readable version of the last build that you can check. Also if you're debugging, it might be helpful to keep a readable version around. Up to you, although deleting it will probably require one additional build task for cleanup.

thijstriemstra:
@skorlir but it doesn't contain the full source, it only has this in it:

```javascript
(function (root, factory) {
  if (typeof define === 'function' && define.amd) {
    // AMD. Register as an anonymous module unless amdModuleId is set
    define('wavesurfer', [], function () {
      return (root['WaveSurfer'] = factory());
    });
  } else if (typeof exports === 'object') {
    // Node. Does not work with strict CommonJS, but
    // only CommonJS-like environments that support module.exports,
    // like Node.
    module.exports = factory();
  } else {
    root['WaveSurfer'] = factory();
  }
}(this, function () {

'use strict';

/* Init from HTML */
(function () {
    var init = function () {
        var containers = document.querySelectorAll('wavesurfer');

        Array.prototype.forEach.call(containers, function (el) {
            var params = WaveSurfer.util.extend({
                container: el,
                backend: 'MediaElement',
                mediaControls: true
            }, el.dataset);

            el.style.display = 'block';

            var wavesurfer = WaveSurfer.create(params);

            if (el.dataset.peaks) {
                var peaks = JSON.parse(el.dataset.peaks);
            }

            wavesurfer.load(el.dataset.url, peaks);
        });
    };

    if (document.readyState === 'complete') {
        init();
    } else {
        window.addEventListener('load', init);
    }
}());

return WaveSurfer;

}));
```

skorlir:
Ohhh kay. This is a deeper problem. The build is running tasks that need to run synchronously asynchronously. I'm getting a similar problem if, eg, I `rm -r dist` and then try to run `grunt`. The `uglify:plugins` task doesn't find any files, and `uglify:dist` doesn't seem to have the full source either.

thijstriemstra:
Here's how I fixed the .min.min.js problem:

```diff
@@ -71,12 +71,12 @@ module.exports = function(grunt) {
         dest: '<%= concat.dist.dest %>'
       },
       plugins: {
-        files: grunt.file.expandMapping(['dist/plugin/*.js'], '', {
+        files: grunt.file.expandMapping(['dist/plugin/*.js', '!dist/plugin/*.min.js'], '', {
           rename: function(destBase, destPath) {
             var newPath = destBase + destPath.replace('.js', '.min.js');
             return newPath;
           }
-       })
+        })
       }
     },
```

thijstriemstra:
I'll first open a PR that removes the old umd headers from the microphone and timeline plugins.

skorlir:
@thijstriemstra The old headers should already be removed. I'm pretty sure I did that.

thijstriemstra:
@skorlir sorry mate, https://github.com/katspaugh/wavesurfer.js/blob/master/plugin/wavesurfer.microphone.js

skorlir:
That's so weird.

PR #632, opened 2016-02-04
use absolute value for peaks

kylophone:
For proper silence detection, we should actually be taking the absolute value of the peaks. `<=` allows us to use true silence (0.0) as our detection threshold. Also, it would be better to measure RMS instead of peak values. I may add a getRMS function later.

Thanks!


katspaugh:
True, thanks!

Issue #611, opened 2016-01-11
Cursor moves jerkily


katspaugh:
As mention in #592, the cursor movement is slow and not smooth. This appears to be a regression.

Issue #592, opened 2015-11-26
Cursor Position Refreshing Rate


X-Raym:
Hi,
I was wondering
the cursor position (in HTML init) is based on audio element current time value.
Is it possible to have a higher refreshing rate, so that the cursor move more smoothly ?
Currently, it is just a bit jumpy (not as fluid as other players).

Thanks ! :)

katspaugh:
Are you talking about the Web Audio's `audioprocess` event or the Media Element's `timeupdate`? The frequency of the former is rather hight, so I would think it should be even capped down.

On the other hand, `timeupdate` might be rather slow indeed. Maybe adding CSS transition would help.

X-Raym:
@katspaugh 
I'm talking about the speed of the cursor (don't don't what event is related to it).
If you compare HTML init version and regular jQuery init version,
you will see that jQuery version has a very more fluid cursor.

On your website, the cursor is very fluid,
but if you compare HTML init version of your [codepen](http://codepen.io/katspaugh/pen/WQXzpB), you will see that the cursor is a bit "jumpy".

Do you think it can be more fluid for HTML init version ?

katspaugh:
I see what you mean. This is because Media Element's `timeupdate` event's frequency is quite low. CSS transition didn't help. HTML version uses Media Element for the sake of the built-in player UI it provides.

Since you are creating your own player, you could use the JavaScript init instead, with `WebAudio` as the backend.

X-Raym:
@katspaugh 
Do you mean adding data-backend="WebAudio" on wavesurfer HTML Tag ? this doesn't produce audio tag... How could I play the sound then ?
I think I have to use HTML init because it is how WordPress shortcodes (in post content) are render, it can't push javascript datas (as far as I know), only pure html.
That's why I had to make my own buttons.

I'm not that experienced in javascript/jQuery (this project is a way to learn a bit more), so maybe there was more optimum ways to do. Still a bit hard to put all this together.

Anyway, thanks for your support, it helps !

katspaugh:
As far as I understood you still use some JavaScript to bind the buttons. So I suppose you could insert `<div id="wavesurfer_0"></div>` in place of the shortcode and bootstrap wavesurfer.js from JavaScript (like in the readme).

X-Raym:
[...]

Issue #498, opened 2015-08-08
Choppy playhead movement/scrolling when setting backend: 'MediaElement'


stephenlaughton:
I am getting really choppy playhead scrolling on all the examples where backend is set to MediaElement. When remove any backend setting from init (using default WebAudio backend) scrolling becomes smooth.

Anyone else seeing this sort of behaviour? I am testing in the latest versions of Chrome and FireFox.

katspaugh:
I don't see any difference in scrolling smoothness between http://wavesurfer-js.org/example/annotation/ and the rest of demo pages.

If you can find out the immediate problem or a way to reproduce it consistently, please feel free to reopen this issue.

Issue #379, opened 2015-03-03
When playing a region, the cursor stops slightly after the end


jlegoff:
Hi!

I am using wavesurfer's region plugin and I want to be able to play a given selected region. That is, I want the playback to start at the beginning of the region and to stop right at the end. What I noticed is that, many times, the progress bar stops slightly after the end of the region.

After tinkering with the code for a while, I think the problem is the following: wavesurfer.play(start, end) calls backend.play(start, end) in order for the media to play; but the drawing of the progress is handled in restartAnimationLoop, which does not have the start/end parameters. What I think happens is that the requestFrame callback gets called more often than the timeupdate event is fired by the mediaelement. Therefore the progress bar gets painted slightly after the end of the region is reached and stops when timeupdate is fired with time > end.

In order to fix it, I added the start / end parameters to restartAnimationLoop and did something like:

```javascript
if (!my.backend.isPaused() && my.backend.getCurrentTime() <= end) {
  my.drawer.progress(my.backend.getPlayedPercents());
    requestFrame(frame);
}
```

If you think this is a proper way of fixing the issue, I'd be more happy to create a pull request.

katspaugh:
Hi @jlegoff!

Sorry for the delayed reaction. I think we might remove `requestAnimationFrame` completely. Can you check with the commit above if the problem still reproduces?

PR #633, opened 2016-02-04
Smoothed out the movement play bar while using a MediaElement backend

jj227:
As mentioned in #498, #592, and #611 

I smoothed out the movement of the play bar while using a MediaElement backend by polling for updates using requestAnimationFrame.  The update loop is broken on pause and ended.

You can see the improvement by comparing:
http://rawgit.com/jj227/wavesurfer.js/master/example/annotation/index.html
http://rawgit.com/katspaugh/wavesurfer.js/master/example/annotation/index.html

### DIFF
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -122,10 +122,25 @@ var WaveSurfer = {
         this.backend.on('pause', function () { my.fireEvent('pause'); });
 
         this.backend.on('audioprocess', function (time) {
-            my.drawer.progress(my.backend.getPlayedPercents());
             my.fireEvent('audioprocess', time);
         });
     },
+
+    startAnimationLoop: function () {        
+        var my = this;        
+        var requestFrame = window.requestAnimationFrame ||
+                           window.webkitRequestAnimationFrame ||
+                           window.mozRequestAnimationFrame;
+        var frame = function () {
+            if (!my.backend.isPaused()) {
+                var percent = my.backend.getPlayedPercents();
+                my.drawer.progress(percent);
+                my.fireEvent('audioprocess', percent);
+                requestFrame(frame);
+            }
+        };
+        frame();
+     },
 
     getDuration: function () {
         return this.backend.getDuration();
@@ -137,6 +152,7 @@ var WaveSurfer = {
 
     play: function (start, end) {
         this.backend.play(start, end);
+        this.startAnimationLoop();
     },
 
     pause: function () {


src/mediaelement.js
```diff
@@ -105,6 +106,14 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
...
         this.media.play();
         end && this.setPlayEnd(end);
+        var my = this;
+        function updateProgress(){
+            my.fireEvent('audioprocess', my.getCurrentTime());
+            if (my.updateProgressAnimating) {
+                requestAnimationFrame(updateProgress);
+            }
+        }
+        this.updateProgressAnimating = requestAnimationFrame(updateProgress);
```
katspaugh: It shouldn't be part of the drawer. It should be used in the drawer code. I'm OK for having `requestAnimationFrame` for both Web Audio and Media Element.

src/mediaelement.js
```diff
@@ -105,6 +106,14 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
         this.seekTo(start);
         this.media.play();
         end && this.setPlayEnd(end);
+        var my = this;
+        function updateProgress(){
+            my.fireEvent('audioprocess', my.getCurrentTime());
+            if (my.updateProgressAnimating) {
+                requestAnimationFrame(updateProgress);
```
katspaugh: Doesn't it still need vendor prefixes in some browsers?

katspaugh:
We used to have `requestAnimationFrame`:
https://github.com/katspaugh/wavesurfer.js/commit/786ac62d8d8f3ff9cc91db84edb860c4c30ea6cf

The reason was that I thought it would solve #379, but now I realize `requestAnimationFrame` had nothing to do with that problem.

src/mediaelement.js
katspaugh: Sorry, not in the drawer code, but in `wavesurfer.js` (since it has access to both the backend and the drawer).

jj227:
the old requestAnimationFrame code you had was a little more concise with how it stops the loop so i merged that back in with two small changes:

- i added in a firing of the audioprocess event because my use case involves keeping a separate play bar in sync using that event.  i suppose i could use two animation loops to keep the updated separately if you don't think this is a good idea.
- most browsers stopped using the prefixes mid-2013 but i left in the webkit prefix and added a second prefix for old versions of firefox

katspaugh:
> i added in a firing of the audioprocess event

No harm in that, I guess.

> most browsers stopped using the prefixes for requestAnimationFrame mid-2013

Cool! Good to know.

Can you [squash](https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/) all commits into one, please? You can do that with `git rebase -i HEAD~13`.

Thanks!

katspaugh:
Thank you! :thumbsup:

PR #640, opened 2016-02-17
Polyfill for IE 9/10 for window.devicePixelRatio

jbeyer05:
IE 9/10 does not support window.devicePixelRatio.  Wavesurfer.js doesn't seem to officially support IE 9/10, but a number of members of the community are interested in supporting IE 9/10.  Our project is using wavesurfer and is supporting IE 9/10 by generating PCM on server side, and using MediaElement fallback.


katspaugh:
Thanks!

PR #644, opened 2016-02-18
Fix ScriptProcessor's buffer size

fzyukio:
ScriptProcessor is created with parameter 'bufferSize' set to 0, it'll select a default value (4096) which is too big for short audio recording. Hence the spectrogram is rendered with very wide pixel blocks. This commit will determine the best buffer size depending on the length of the recording and the width of the canvas.

Before

After the fix:
After

### DIFF
--- a/plugin/wavesurfer.spectrogram.js
+++ b/plugin/wavesurfer.spectrogram.js
@@ -135,12 +135,19 @@ WaveSurfer.Spectrogram = {
             this.fireEvent('error', 'Web Audio buffer is not available');
             return;
         }
+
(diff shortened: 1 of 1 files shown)


katspaugh:
Cool, thanks! cc @akreal

fzyukio:
@katspaugh Thanks! This is my first pull request!.
And actually I'm fixing the spectrogram plugin all over again to fix the annoying problem with OfflineAudioContext ... stay tune :smile_cat:

katspaugh:
@fzyukio ごくろさま :thumbsup:

akreal:
Wow! Thank you!

PR #645, opened 2016-02-18
New Styling Options

gjsmith66:
Add options add background-url to Wave-Canvas, and change overFlow of
Wave container.  Also add option to override style of Region container.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,7 @@ where you can also add your own project.
 | `audioContext` | object | _none_ | Use your own previously initialized `AudioContext` or leave blank. |
 | `audioRate` | float | `1` | Speed at which to play audio.  Lower number is slower. |
 | `backend` | string | `WebAudio` | `WebAudio` or `MediaElement`. In most cases you don't have to set this manually. `MediaElement` is a fallback for unsupported browsers. |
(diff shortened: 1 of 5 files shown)


chinshr:
Nice! The background image could be used to show a thumb image at the current position. Would be interesting to see how to align multiple image thumbs like in video authoring software on desktops.

gjsmith66:
Thanks, I use it to display a grid behind the wave form, like so.

slack-imgs com

katspaugh:
Can't you just set these styles via CSS on the container? The container is supposed to be controlled by the application, isn't it?

gjsmith66:
For the background-url, yes I suppose that it could be set using CSS style.  However I believe the overflow will be over-ridden by the element style.  Same for the Region style I think.

gjsmith66:
Yes, I was able to make all changes I needed via CSS and using !important.  Adding the ability to add programmatically may be useful, otherwise it does appear that using CSS alone will work.

Thanks.

katspaugh:
> Yes, I was able to make all changes I needed via CSS and using !important.

Exactly. Besides, `overflow: hidden` is already supported as the [`hideScrollbar` option](https://github.com/katspaugh/wavesurfer.js/blob/master/src/wavesurfer.js#L23).

Let's keep the API compact. We can be encourage using CSS directly in the documentation.

PR #646, opened 2016-02-18
Fix Spectrogram problem because of webkit's OfflineAudioContext

fzyukio:
Use my own FFT calculater instead of relying on the broken webkit. Now all browsers are supported and the calculation is much much faster!.

To use:

`     ws.Spectrogram.init ( (

      wavesurfer: self.ws,
      fftSamples: 512,
      noverlap: 256, // But it's better to leave this param unset - the program will choose the best value based on canvas width to maximise resolution.
      pixelRatio: 1,
      windowFunc: 'hann',
      container: "#waveform-spectrogram"
});
`

Provide 10 different window functions to choose from.
FFT calculator is based on @corbanbrook library (https://github.com/corbanbrook/dsp.js)


plugin/wavesurfer.spectrogram.js
```diff
@@ -230,7 +247,184 @@ WaveSurfer.Spectrogram = {
         }
 
         return newMatrix;
-    }
+    },
```
katspaugh: A trailing comma (not critical).

plugin/wavesurfer.spectrogram.js
```diff
@@ -230,7 +247,184 @@ WaveSurfer.Spectrogram = {
...
+
 };
 
 WaveSurfer.util.extend(WaveSurfer.Spectrogram, WaveSurfer.Observer);
+
+function FFT(bufferSize, sampleRate, windowFunc, alpha) {
```
katspaugh: Maybe store this function in `WaveSurfer.FFT`?

katspaugh: Could you also add a comment about the code being based on https://github.com/corbanbrook/dsp.js?

katspaugh:
すごいです。:bow:

screen shot 2016-02-19 at 10 36 50
screen shot 2016-02-19 at 10 37 53

plugin/wavesurfer.spectrogram.js
fzyukio: ( ͡° ͜ʖ ͡°)

fzyukio:
@katspaugh ありがとう　:+1:

katspaugh:
@akreal I hope you don't mind :smile:

akreal:
@katspaugh Sure!

PR #648, opened 2016-02-21
Fix region's width when zoom out

fzyukio:
When zooming out, `this.wrapper.scrollWidth` somehow always returns a wrong value. Using pxPerSec is more robust to get the width of the waveform.


katspaugh:
Good catch, thanks!

PR #652, opened 2016-02-22
critical: fix broken build

skorlir:
The current default build task does *not* correctly build WaveSurfer.

Moreover, the current `dist/wavesurfer.min.js` in master is very very incomplete. @thijstriemstra caught this, but I don't think he realized how far it went.

The reason this wasn't caught originally was because I hadn't done an `rm -r dist`. Since there was a pre-existing "good" intermediary build the first several times I tested the `umd` tasks, I didn't realize that the build broke if dist did not have pre-existing files in it.

In other words, mea culpa. :(

This PR also encapsulates #651, because I felt dumb I rebased out the original commit removing the old headers. :/


thijstriemstra:
There's some weird spacing added to Gruntfile.js. Can you make sure your editor is set to 2 spaces?

skorlir:
My editor is ordinarily set to two spaces - not sure what happened, but should be fixed.

thijstriemstra:
I see there's now a `dist/wavesurfer.js` that contains the unminified build. There's already a src/wavesurfer.js source file and to prevent confusion I'd rather call the dist version `dist/wavesurfer.dev.js`. Thoughts?

skorlir:
That's true, although `dist/wavesurfer.js` is not tracked. We can rename it, but unless someone builds from source they won't see it anyway. Doesn't matter to me. It's just intermediate output from the `concat` task. Give me a sec, I'll push an update.

thijstriemstra:
Can you also doublecheck the .gitignore file, there are probably some entries that are not needed anymore.

skorlir:
I double-checked the .gitignore. It's all needed stuff.

Gruntfile.js
```diff
@@ -40,7 +40,7 @@ module.exports = function(grunt) {
     umd: {
       main: {
         options: {
-          src: '<%= concat.dist.src %>',
+          src: '<%= concat.dist.dest %>',
           dest: 'dist/wavesurfer.umd.js',
```
thijstriemstra: why is wavesurfer.umd.js still created? Shouldn't this be wavesurfer.min.js? If not, can we get rid of it? wavesurfer.dev.js is our non-minified reference build file now isn't it?

plugin/wavesurfer.microphone.js
```diff
@@ -1,218 +1,210 @@
...
+deviceError: function(code) {
+    // notify listeners
+    this.fireEvent('deviceError', code);
+}
+
+};
+
+WaveSurfer.util.extend(WaveSurfer.Microphone, WaveSurfer.Observer);
+
+return WaveSurfer.Microphone;
```
thijstriemstra: are these returns necessary?

skorlir:
Those are intermediary files.

The build has three steps: concatenation (`wavesurfer.dev.js`), UMD wrapping (`wavesurfer.umd.js`), and uglification (`wavesurfer.min.js`). Each step prior to uglification generates an intermediate output file. The intermediary build files are not tracked in git, but they *are* necessary for proper building at each step. If two tasks try to output to the same file, wonky stuff happens.

We could set up a cleanup task for deleting them at the end, but it'll just be an extra step.

katspaugh:
Thanks!

thijstriemstra:
@skorlir @katspaugh guys, the `dist` folder is filled with .min.min.js files, wavesurfer.amd.js is included in dist, this is not good.

thijstriemstra:
@katspaugh if you could check #653 and merge it then I'll take a look at this Gruntfile mess. I mean running grunt keeps adding on files.
```
        dist/plugin/wavesurfer.elan.min.min.min.js
	dist/plugin/wavesurfer.microphone.min.min.min.js
	dist/plugin/wavesurfer.minimap.min.min.min.js
	dist/plugin/wavesurfer.regions.min.min.min.js
	dist/plugin/wavesurfer.spectrogram.min.min.min.js
	dist/plugin/wavesurfer.timeline.min.min.min.js
```

thijstriemstra:
I've fixed the build process (and in the process also got rid of the intermidate wavesurfer.umd.js). Will merge PR when build passes.

PR #653, opened 2016-02-23
microphone plugin: navigator.getUserMedia is deprecated, use navigator.mediaDevices.getUserMedia instead

thijstriemstra:


katspaugh:
Thanks, @thijstriemstra!

Issue #631, opened 2016-02-03
Microphone plugin crashes on Chrome v47+


AlexEdgcomb:
Line 133 crashes on Chrome v47+:
https://github.com/katspaugh/wavesurfer.js/blob/master/plugin/wavesurfer.microphone.js#L133

Chrome removed support for MediaStream.stop() in v47 (deprecated in v45):
https://developers.google.com/web/updates/2015/07/mediastream-deprecations?hl=en

A solution may be to use:
```
this.stream.getAudioTracks().forEach(function(audioTrack) {
    audioTrack.stop();
});
```

AlexEdgcomb:
I just noticed PR #593 is attempting to fix this issue.

PR #655, opened 2016-02-23
use grunt-jscs for code-style

thijstriemstra:
This detects trailing whitespace, use of tabs etc.


thijstriemstra:
@katspaugh if you're happy with this, could you also push a new release?

thijstriemstra:
@katspaugh actually I also want to fix #631 first so I'll ping you again for that new release then.

katspaugh:
Sure, I'm fine with this. I'll publish after #631 then.

PR #656, opened 2016-02-23
Add clean to build

skorlir:
We're having trouble with the build creating extra files when there are already files in `dist`, so here's a `clean` task for deleting `dist` before each build to keep that from happening. #608

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -11,6 +11,9 @@ module.exports = function(grunt) {
       '<%= pkg.homepage ? "* " + pkg.homepage + "\\n" : "" %>' +
       '* @license <%= pkg.license %> */',
     // Task configuration.
+    clean : {
+      build: ['dist']
+    },
     concat: {
       options: {
         stripBanners: true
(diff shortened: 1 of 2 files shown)


skorlir:
@thijstriemstra thanks for helping me out with all this. When you get a chance, let me know if there's anything else I should add to the PR!

Gruntfile.js
```diff
@@ -189,10 +192,11 @@ module.exports = function(grunt) {
...
   grunt.loadNpmTasks('grunt-contrib-connect');
   grunt.loadNpmTasks('grunt-umd');
-  grunt.loadNpmTasks("grunt-jscs");
+  grunt.loadNpmTasks('grunt-jscs');
+  grunt.loadNpmTasks('grunt-contrib-clean');
 
   // Default task.
-  grunt.registerTask('default', ['jshint', 'jscs', 'test', 'coverage', 'concat',
+  grunt.registerTask('default', ['clean', 'jshint', 'jscs', 'test', 'coverage', 'concat',
     'umd', 'uglify']);
```
thijstriemstra: @skorlir can you also add the `clean` to the `dev` task?

thijstriemstra:
thanks @skorlir I added one note about your PR, can you check that out and I'll merge this.

Gruntfile.js
thijstriemstra: Nevermind I'll merge it now and make the change myself so I can continue development.

skorlir: :+1:

PR #659, opened 2016-02-26
add cancelLoad method and info to readme.

pmckelvy1:
I wrote this method to help me in using wavesurfer.js with react.  Occasionally the user would navigate before an audio unit was finished loading which would cause errors.  In the case that a React componentWillUnmount and it is holding a waversurfer unit that is currently loading audio, 'cancelLoad' can be called with the XHR object returned from 'load' to perform garbage collection.


mspae:
You might find this interesting https://github.com/mspae/react-wavesurfer

pmckelvy1:
i might indeed. thank you.

cloudsounds.xyz

README.md
```diff
@@ -86,6 +86,7 @@ All methods are intentionally public, but the most readily available are the fol
  * `getDuration()` – Returns the duration of an audio clip in seconds.
  * `isPlaying()` – Returns true if currently playing, false otherwise.
  * `load(url)` – Loads audio from URL via XHR. Returns XHR object.
+ * `cancelLoad(xhrObject)` - Cancels the loading of audio from a URL. Takes the XHR object returned from `load`. Used for early navigation if loading is incomplete.
```
katspaugh: I think calling it `xhrObject` is a bit misleading. Maybe `ajaxEvent`?

pmckelvy1: makes sense to me. should i change it or will you?

src/wavesurfer.js
```diff
@@ -342,6 +342,17 @@ var WaveSurfer = {
     },
 
     /**
+     * Cancels the audio loading and destorys all necessary handlers
+     * and connections.
+    **/
+    cancelLoad: function (xhrObject) {
+        xhrObject.xhr.abort();
+        this.Observer.unAll();
```
katspaugh: This will kill all external subscribers. I think it's not what most users would want.

src/wavesurfer.js
```diff
@@ -342,6 +342,17 @@ var WaveSurfer = {
...
+        this.empty();
+        this.destroy();
```
katspaugh: Why aborting a load should destroy the whole instance?

README.md
katspaugh: It's your PR, isn't it? :)

katspaugh:
I think instead of adding a new method, we should modify `destroy` so that it aborts any ongoing XHRs. From your react app you would call `wavesurfer.destroy()` as this seems what you want–to destroy the whole player.

katspaugh:
In [this method](https://github.com/katspaugh/wavesurfer.js/blob/master/src/wavesurfer.js#L391-L407) I would subscribe to the `destroy` event and abort the XHR.

README.md
pmckelvy1: maybe...
if im going abort the request and not use the audio, i dont really feel
like storing the rest of it in memory. might as well just instantiate a new
one if that track ever comes up again.

pmckelvy1:
ok i think that makes more sense. ill update it and issue a new pull
request.

katspaugh:
Thanks! I'll close this one then.

PR #660, opened 2016-02-26
add 'destroy' listener to ajax request in getArrayBuffery.

pmckelvy1:
adds a 'destroy' listener to the ajax request in the getArrayBuffer function, which aborts the ajax request when the user calls destroy.


src/wavesurfer.js
```diff
@@ -401,6 +401,9 @@ var WaveSurfer = {
             ajax.on('success', callback),
             ajax.on('error', function (e) {
                 my.fireEvent('error', 'XHR error: ' + e.target.statusText);
+            }),
+            ajax.on('destroy', function (e) {
```
katspaugh: The `ajax` object doesn't have a `destroy` event. It should listen to the wavesurfer's `destroy` event. It should also unsubscribe from it on success or error (of Ajax).

Also use `once` instead of `on`.

src/wavesurfer.js
```diff
@@ -463,7 +466,8 @@ var WaveSurfer = {
         this.unAll();
         this.backend.destroy();
         this.drawer.destroy();
-    }
+    },
```
katspaugh: A trailing comma and a new line.

katspaugh:
I did it a bit differently, please see the commit above.

PR #661, opened 2016-02-26
add unmount and remount functions.

pmckelvy1:
added public functions to unmount and remount the canvas elements from/to the specified DOM Element on the page.


src/wavesurfer.js
```diff
@@ -441,6 +441,39 @@ var WaveSurfer = {
     },
 
     /**
+     * Unmounts the wavesurfer instance from the page, but keeps the audio
+     * player in memory
+     */
+    unmount: function (selector) {
+        if (this.container.children.length === 0) {
+            var wsNode = document.querySelectorAll(selector)[0];
```
katspaugh: The wavesufer element can be accessed as `this.drawer.wrapper`. It's better to define the unmount method in `drawer.js` and just call it from here.

src/wavesurfer.js
```diff
@@ -441,6 +441,39 @@ var WaveSurfer = {
...
+            var newNode = wsNode.cloneNode();
+            this.container.appendChild(newNode.children[0]);
+            // $(this.container).append(newNode.children[0]);
```
katspaugh: Please, remove this line.

src/wavesurfer.js
```diff
@@ -441,6 +441,39 @@ var WaveSurfer = {
...
 
     /**
+     * Unmounts the wavesurfer instance from the page, but keeps the audio
+     * player in memory
+     */
+    unmount: function (selector) {
+        if (this.container.children.length === 0) {
+            var wsNode = document.querySelectorAll(selector)[0];
+            var newNode = wsNode.cloneNode();
+            this.container.appendChild(newNode.children[0]);
```
katspaugh: I don't quite get this logic with the clone and append. If you want to remove the element, wouldn't simply calling `this.container.removeChild(this.drawer.wrapper)` be enough?

src/wavesurfer.js
```diff
@@ -441,6 +441,39 @@ var WaveSurfer = {
     },
 
     /**
+     * Unmounts the wavesurfer instance from the page, but keeps the audio
+     * player in memory
+     */
+    unmount: function (selector) {
+        if (this.container.children.length === 0) {
```
katspaugh: This check is not correct. The container can have other elements besides wavesurfer.

src/wavesurfer.js
```diff
@@ -441,6 +441,39 @@ var WaveSurfer = {
...
+     * with specified height
+     */
+    remount: function (selector, height) {
+        var container = document.querySelectorAll(selector)[0];
+        var child = this.container.children[0];
+
+        container.appendChild(child);
+
+        this.container = container;
+        this.mediaContainer = container;
```
katspaugh: `mediaContainer` can be a separate element. In the unmount method this should be taken into consideration.

src/wavesurfer.js
```diff
@@ -441,6 +441,39 @@ var WaveSurfer = {
...
+        this.drawer.container = container;
+
+        this.drawer.height = height;
+        this.drawer.params.height = height;
```
katspaugh: Setting a new height seems to me a separate functionality that should be a separate method (`setHeight`). It shouldn't be part of remount.

src/wavesurfer.js
```diff
@@ -441,6 +441,39 @@ var WaveSurfer = {
...
+        }
+    },
+
+    /**
+     * Remounts the wavesurfer object onto selected DOM Element
+     * with specified height
+     */
+    remount: function (selector, height) {
+        var container = document.querySelectorAll(selector)[0];
+        var child = this.container.children[0];
```
katspaugh: As said above, use `this.drawer.wrapper`.

src/wavesurfer.js
```diff
@@ -441,6 +441,39 @@ var WaveSurfer = {
     },
 
     /**
+     * Unmounts the wavesurfer instance from the page, but keeps the audio
+     * player in memory
+     */
+    unmount: function (selector) {
```
katspaugh: No need to pass the selector since the current container is always referenced in `this.container`.

src/wavesurfer.js
katspaugh: Or I don't get what this `if` does. Why is it checking whether the container is empty?

PR #663, opened 2016-02-27
pass seconds instead of percents to audioprocess event 

akreal:


thijstriemstra:
Can you also explain this PR @akreal? What's broken exactly?

akreal:
`audioprocess` passes seconds in `createBackend`, so it should do here.

katspaugh:
Thank you very much!

PR #674, opened 2016-03-16
fix plugin path in audio-element example

therewasaguy:


katspaugh:
Thanks! Is this the only file with the wrong path?

Btw, it could be `../../dist/plugin/wavesurfer.regions.min.js` (with `min`), but the source file is also OK.

PR #678, opened 2016-03-22
Cursor plugin

greegus:
A proposal for simple plugin which renders mouse line-cursor when hovering above the wave


greegus:
image

katspaugh:
Cool, thank you!

Issue #413, opened 2015-03-23
Long mp3 file with peaks


RocAlayo:
I'm using wavesurfer with this options:
```javascript
{
      container: document.querySelector("div.wave-container"),
      scrollParent: true,
      fillParent: false,
      hideScrollbar: false,
      cursorWidth: 2,
      waveColor     : '#cccccc',
      progressColor : '#0189BE',
      loaderColor   : 'purple',
      cursorColor   : '#ff0000',
      backend: "MediaElement",
      mediaType: "audio"
}
```
 
Then I try to load a long mp3 file (around an hour) and I specify the peaks arra (consequently It's a huge array of about 18000 positions). Then the wavesurfer canvas renders to the needed width with scroll but the wave didn't get printed in the canvas. I debugged the problem and the method `drawWave` seems to work as expected (makes the lines correctly as far as I can tell).

As a test I tried to load the same array of peaks but with an mp3 file of about 3 minutes and this way the wave does get printed.

Any idea of what could be the problem?

katspaugh:
Looks like the same as #47, i.e. canvas size limit.

Try to decrease the number of square pixels on the canvas by setting a lower minPxPerSecond and smaller height.

RocAlayo:
I tried it and it didn't work. This is weird because I'm updating a webpage that already uses wavesurfer with peaks arrays of this size and works fine, but the version of the wavesurfer used is 9 month old.

katspaugh:
Also with a huge waveform it would be necessary to set pixelRatio to 1 so that it doesn't blow Retina up.

Issue #47, opened 2013-09-22
minPxPerSec can not be set to anything above 59 (Canvas size limitation)


muggy8:
So i want to set my minPxPerSec to 100 because I'm making horizontal music thing and it would be nice if the waveform matched everything else but as i have discovered that if minPxPerSec is set to anything more than 59, the waveform it outputs is transparent. Thanks :)

tomByrer:
Just taking a stab at a fast-fix:

1. Look to see if long sample
2. If true, [increase divisor of `sampleSize`](https://github.com/katspaugh/wavesurfer.js/blob/master/src/webaudio.js#L171)

chrisparton1991:
I've created a pull request that should allow greater minPxPerSec values: #679

chrisparton1991:
Can this be closed, @katspaugh?

katspaugh:
Sure, thanks!

PR #679, opened 2016-03-24
Multi-Canvas Renderer

chrisparton1991:
I need very large canvases for some personal light sequencing software I'm working on (for an interactive xmas light display), and I quickly hit Chrome's canvas address limitations, as detailed in the following issues:
 - #47
 - #413

I have implemented a Drawer that uses multiple adjacent canvases, and draws seamlessly between them. I've attached a zip file containing the following examples:
 - A 50000px waveform
 - Multiple canvases with bars
 - Multiple canvases with split channels
 - A waveform with pixel ratio of 10
 - Lots of tiny canvases

To see the examples, run multicanvas-tests/src/app-test.html.

I'm no JavaScript expert, so it's possible I'm not always following best practices. I figure there are a few options for this drawer:
 
1. Replace the current CanvasDrawer with this one (rename it from MultiCanvasDrawer to CanvasDrawer)
2. Add MultiCanvasDrawer as its own file so users can select which drawer they would like to use (which is how it's set up now)
3. I can put MultiCanvasDrawer into a repo under my name and submit it to Bower.

Thanks for your consideration,
Chris

[multicanvas-tests.zip](https://github.com/katspaugh/wavesurfer.js/files/187276/multicanvas-tests.zip)

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -10,6 +10,28 @@ WaveSurfer.util = {
         return dest;
     },
 
+    min: function(values) {
+        var min = +Infinity;
+        for (var i in values) {
+            if (values[i] < min) {
+                min = values[i];
+            }
+        }
+
+        return min;
+    },
+
+    max: function(values) {
+        var max = -Infinity;
+        for (var i in values) {
+            if (values[i] > max) {
+                max = values[i];
+            }
+        }
+
+        return max;
+    },
+
     getId: function () {
         return 'wavesurfer_' + Math.random().toString(32).substring(2);
     },
--- a/spec/util.spec.js
+++ b/spec/util.spec.js
@@ -4,4 +4,19 @@ describe('util', function() {
         expect(WaveSurfer.util.getId()).toStartWith('wavesurfer_');
     });
 
+    it('min returns the smallest number in the provided array', function() {
+        expect(WaveSurfer.util.min([0, 1, 1.1, 100, -1])).toEqual(-1);
+    });
+
+    it('min returns +Infinity for an empty array', function() {
+        expect(WaveSurfer.util.min([])).toEqual(+Infinity);
+    });
+
+    it('max returns the largest number in the provided array', function() {
+        expect(WaveSurfer.util.max([0, 1, 1.1, 100, -1])).toEqual(100);
+    });
+
+    it('max returns -Infinity for an empty array', function() {
+        expect(WaveSurfer.util.max([])).toEqual(-Infinity);
+    });
 });
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -10,6 +10,7 @@ WaveSurfer.Drawer = {
 
         this.lastPos = 0;
 
+        this.initDrawer(params);
         this.createWrapper();
         this.createElements();
     },
@@ -182,6 +183,8 @@ WaveSurfer.Drawer = {
     },
 
     /* Renderer-specific methods */
+    initDrawer: function () {},
+
     createElements: function () {},
 
     updateSize: function () {},
--- a/src/drawer.multicanvas.js
+++ b/src/drawer.multicanvas.js
@@ -1,0 +1,309 @@
+'use strict';
+
+WaveSurfer.Drawer.MultiCanvas = Object.create(WaveSurfer.Drawer);
+
+WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
+
+    initDrawer: function (params) {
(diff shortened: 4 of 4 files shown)


src/util.js
```diff
@@ -10,6 +10,28 @@ WaveSurfer.util = {
         return dest;
     },
 
+    min: function(values) {
+        var min = null;
```
katspaugh: The initial value could be `Infinity`, then you don't need to check if it's null.

src/util.js
```diff
@@ -10,6 +10,28 @@ WaveSurfer.util = {
...
+            if (min == null || values[i] < min) {
+                min = values[i];
+            }
+        }
+
+        return min;
+    },
+
+    max: function(values) {
+        var max = null;
```
katspaugh: The same, but `-Infinity`.

katspaugh:
Very nice, great job!

Let's have it as a separate renderer for now. Then after a while, I think, we can use this by default instead of the single canvas.

It's not a big deal, but could you [squash](https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/) the commits into one?

chrisparton1991:
Thanks :) I hope others will find it a useful addition to the library.

I'm a bit busy today, but I'll make those changes tomorrow and squash the commits. I was going to squash them initially, but I figured it'd be easier to review as separate commits first.

Would you like me to add tests for the min() and max() functions? The `maxCanvasSize` parameter will also need to be documented.

katspaugh:
> Would you like me to add tests for the min() and max() functions?
Only if you want. :)

> The maxCanvasSize parameter will also need to be documented.

As well as the `drawer` parameter (see here: https://github.com/katspaugh/wavesurfer.js/blob/master/src/wavesurfer.js#L85). We'll also need to include `src/drawer.multicanvas.js` into the build in the Gulpfile.

chinshr:
This is super useful. I too run into canvas limitations for large files. Does this work with zoom? Where is this going to live?

chrisparton1991:
Hi @chinshr,
I have to fix up a couple of things in the pull request. Once my code has been merged in, this renderer will be available in the latest version of the library using the `drawer: 'MultiCanvas'` parameter (which I will add documentation for sometime in the next few days, I'm a bit busy with Easter stuff at the moment).

I believe this renderer will work with zoom. I did some basic testing, but I look forward to feedback from other users :)

As for when the next release will be made, that will be something for @katspaugh to answer.

chrisparton1991:
@katspaugh, I don't think I need to touch the `Gruntfile`, it already has the following inclusion: `src/drawer.*.js`.

chinshr:
Awesome, I will test your PR with my app.

chrisparton1991:
Thanks @chinshr, I look forward to your feedback.

@katspaugh After a few failed builds, I've fixed up the tests and satisfied JsLint. I just need to raise a separate PR for the documentation update, which I'll do tonight or tomorrow.

katspaugh:
Perfect!

chinshr:
Hey @chrisparton1991, finally had a chance to take a look at the MultiCanvas. Great job, I really appreciate your effort. 

I ran into a problem. I am using timeline and region plugins and I am loading my peaks from a json file (see configuration params below). Using the default `maxCanvasWidth`. The issue appears that the waveform (see screenshot) beyond 2:11 is rendered with a "gap" after 3:20. Around 2:10 the new canvas is starting.

canvas
multi-canvas

```js
    var options = {
      container     : '#waveform',
      height        : 40,
      waveColor     : '#ddd',
      progressColor : '#fff',
      loaderColor   : '#555',
      cursorColor   : '#5492ce',
      cursorWidth   : 1,
      audioRate     : 1,
      scrollParent  : true,
      normalize     : true,
      minimap       : true,
      minPxPerSec   : 10,
      pixelRatio    : 2,
      backend       : 'AudioElement',
      fillParent    : true,
      hideScrollbar : false, 
      dragSelection : false,
      loopSelection : false,
      interact      : true,
      splitChannels : false,
      skipLength    : 2,
      mediaType     : 'audio',
      mediaControls : false,
      barWidth      : 0,
      autoplay      : true,
      renderer      : "MultiCanvas"
    };
```

whenov:
Hi @chrisparton1991 

The `clearWave()` functions provided by `Canvas` and `MultiCanvas` don't have an identical interface (the `clearWave` function in `MultiCanvas` has a parameter), which may cause problem if called from `WaveSurfer` class. What do you think?

chrisparton1991:
Hi @whenov, nice find. I'll fix this up and submit a PR.

chrisparton1991:
New PR created @whenov @katspaugh: #706

PR #682, opened 2016-03-26
Change getPeaks baseline to zero

nathansnider:
The getPeaks function was using the first sample in each channel to initialize the min and max values, causing files that start above or below zero to be drawn with persistent offsets. Initializing the min and max to 0 fixes this.


katspaugh:
_Very_ good catch! Silence in the beginning was a bad assumption. Thanks!

PR #684, opened 2016-03-27
Region not required when removing it

h0lyalg0rithm:


plugin/wavesurfer.regions.js
```diff
@@ -461,6 +461,10 @@ WaveSurfer.clearRegions = function () {
     this.regions && this.regions.clear();
 };
 
+WaveSurfer.clearRegion = function (region) {
+    this.regions && region.remove();
+};
```
katspaugh: I don't think such a method is useful. One can simply call `region.remove()`.

plugin/wavesurfer.regions.js
```diff
@@ -169,7 +169,7 @@ WaveSurfer.Region = {
     },
 
     /* Remove a single region. */
-    remove: function (region) {
+    remove: function () {
```
katspaugh: 👍

h0lyalg0rithm:
I will push a commit without the helper method.

katspaugh:
Thank you, merged.

Issue #686, opened 2016-03-31
Click-to-seek doesn't work appropriately for short audio when fillParent=false


whenov:
If `fillParent` is `false`, and the audio is not long enough to span the entire canvas, then click-to-seek won't work as expected.

PR #687, opened 2016-03-31
Fix click-to-seek for audios which are not spanning the entire canvas

whenov:
Solves #686

### DIFF
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -41,8 +41,25 @@ WaveSurfer.Drawer = {
 
     handleEvent: function (e) {
         e.preventDefault();
+
         var bbox = this.wrapper.getBoundingClientRect();
-        return ((e.clientX - bbox.left + this.wrapper.scrollLeft) / this.wrapper.scrollWidth) || 0;
+
+        var nominalWidth = this.width;
+        var parentWidth = this.getWidth();
+
+        var progress;
+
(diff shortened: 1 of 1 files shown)


src/wavesurfer.js
```diff
@@ -92,6 +92,20 @@ var WaveSurfer = {
...
+            var parentWidth = my.drawer.getWidth();
+
+            if (!my.params.fillParent && nominalWidth < parentWidth) {
+                var bbox = my.drawer.wrapper.getBoundingClientRect();
+                progress = (e.clientX - bbox.left) * my.params.pixelRatio / nominalWidth;
+
+                if (progress > 1) {
+                    progress = 1;
+                }
+            }
```
katspaugh: This logic should be inserted here: https://github.com/katspaugh/wavesurfer.js/blob/master/src/drawer.js#L44 or https://github.com/katspaugh/wavesurfer.js/blob/master/src/drawer.js#L63, in the drawer code.

whenov: I was trying to do that, but getDuration() is not accessible in drawer

katspaugh: True, but I think you can take `this.drawer.width` (or `this.width` in the drawer code) instead of calculating `nominalWidth`.

whenov: Reasonable.

katspaugh:
Nice, thanks!

PR #688, opened 2016-03-31
Selection plugin

greegus:
Selection plugin allows to create mouse click-and-drag selections over wave timeline. This feature is potentially useful for some sort of audio editor, in which user needs to select a part of song to work with.

### DIFF
--- a/example/selection/app.js
+++ b/example/selection/app.js
@@ -1,0 +1,56 @@
+'use strict';
+
+// Create an instance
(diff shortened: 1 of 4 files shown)


greegus:
image

greegus:
image

katspaugh:
Thanks, but how is this different from the existing drag-to-select functionality in the [regions plugin](http://wavesurfer-js.org/plugins/regions.html)?

PR #689, opened 2016-04-01
Allow to play/draw one channel specified by parameter

whenov:


src/webaudio.js
```diff
@@ -128,6 +128,30 @@ WaveSurfer.WebAudio = {
...
+    },
+
+    setChannel: function (channel) {
+        var channels = this.buffer.numberOfChannels;
+
+        this.splitter.disconnect();
+
+        for (var c = 0; c < channels; c++) {
+            this.splitter.connect(this.merger, channel === -1 ? c : channel, c);
+        }
```
katspaugh:
```
if (channel > -1) {
    this.splitter.connect(this.merger, channel);
} else {
    for (var c = 0; c < channels; c++) {
        this.splitter.connect(this.merger, c);
    }
}
```

katspaugh:
Nice! Does it add much overhead in terms of performance, though?

src/webaudio.js
whenov: I tried to omit the third parameter also, but the output of splitter is only connected to the first channel of the merger (tested on newest Chrome/Firefox).

Although the last two parameters are marked as optional in the document of [AudioNode.connect](https://webaudio.github.io/web-audio-api/#widl-AudioNode-connect-AudioNode-AudioNode-destination-unsigned-long-output-unsigned-long-input), the default behaviour is not described.

whenov:
@katspaugh Although I haven't tested its performance (and I haven't considered how to do it), the added logic seems not to include costly overhead.

But I'm not sure about the splitter/merger nodes, whose performance are dependent on browser implementations. If this channel-switch function is not required by most users, we can directly connect `analyser` to `gainNode` when `channel === -1`, which however decreases a little readability.

katspaugh:
Could you measure CPU load/memory usage with and without your changes? You can simply open the default OS activity monitor and watch the browser process.

whenov:
@katspaugh Good idea, will do it tomorrow.

whenov:
@katspaugh I tested its performance using Chrome's profiling tool, the increased execution time is negligible.

katspaugh:
Cool, merged! Thanks!

PR #693, opened 2016-04-07
Change MultiCanvas line drawing logic to handle prerendered peaks

chrisparton1991:
Implemented fix proposed by @chinshr, with a minor change. I calculate the `length` variable in the `drawWave` method instead of `drawLine` to avoid redundant calculations.

Please don't accept until @chinshr has had a look.


chinshr:
This works. Great job, thanks!

PR #706, opened 2016-04-25
Remove argument from MultiCanvas clearWave() method

chrisparton1991:
Renamed `clearWave()` to `clearWaveForEntry()`. Added `clearWave()` method that calls `clearWaveForEntry()` on all canvases.

Question: The only thing calling `clearWave()` is the `setChannel()` method in `wavesurfer.js`. This method isn't in the API documentation as far as I can tell. Is it still used?


whenov:
The `setChannel()` method is added by me recently ([https://github.com/katspaugh/wavesurfer.js/pull/689](https://github.com/katspaugh/wavesurfer.js/pull/689)), and I was too lazy to update the documentation...

katspaugh:
Merged, thanks!

Issue #698, opened 2016-04-10
Control video tag/videojs with wavesurfer?


chinshr:
Is it possible to create a video tag created by your logic and then have it controlled by wavesurfer? I want to playback HLS content with wavesurfer.

katspaugh:
It's the `mediaType: 'video'` parameter. I've added docs for it on the website.

chinshr:
How would one specify the video tag in the options? In the docs you are specifying to create an audio/video tag, but what if it already exists?

katspaugh:
Oh, an existing video element... Maybe https://github.com/collab-project/videojs-wavesurfer can give you some hints?

@thijstriemstra created it, so maybe he can advice.

chinshr:
Got it, thanks. I will check @thijstriemstra
<https://github.com/thijstriemstra> code.

therewasaguy:
I'm also interested in using wavesurfer with HLS, and I've been working on this in a fork [here](https://github.com/therewasaguy/wavesurfer.js/tree/load-existing-media-elt).

I needed the ability to load an existing audio or video element. So I added a `loadExistingMediaElement` method to wavesurfer.js, and a loadElt method to mediaelement.js.

To use this in my project, I init wavesurfer with `backend: 'MediaElement' `. I have not gotten HLS to work with Web Audio to render the waveform, so I pre-analyze the waveform data, use waveform.util.wjax to get the waveform data, then loadExistingMediaElement like this:

```
  wavesurfer.util.ajax({
    responseType: 'json',
    url: waveformPath // path to waveform.json data
  }).
  on('success', function (data) {
      wavesurfer.loadExistingMediaElement( htmlElement, data);
      wavesurfer.zoom(4);
  });
```

Here are the changes I made so far https://github.com/therewasaguy/wavesurfer.js/commit/49717b5dafb5e6cfd84eb0bd81ee95cb45bec85c

I wonder if this is a good approach? If it's helpful I can clean it up and send a PR. But maybe there's a better way, and I'd be very open to suggestions!

katspaugh:
@therewasaguy sounds like a good approach indeed. A PR would be awesome!

chinshr:
@therewasaguy do you mind creating a PR?

therewasaguy:
@katspaugh @chinshr here's a PR https://github.com/katspaugh/wavesurfer.js/pull/707

chinshr:
@therewasaguy awesome! Thanks so much!

[...]

PR #707, opened 2016-04-25
Load existing media element

therewasaguy:
Adds the ability to load an existing HTML5 <audio> or <video> element with `loadExistingMediaElement` method in wavesurfer.js. Also adds a `loadElt` method to wavesurfer's mediaelement.js file.

Usage:

init wavesurfer with ```backend: 'MediaElement'```.

```
 wavesurfer.util.ajax({
    responseType: 'json',
    url: waveformPath // path to .json file with precomputed waveform data
  }).
  on('success', function (data) {
      wavesurfer.loadExistingMediaElement( htmlElement, data);
  });
```

What is the best way to document this? Should I add a line to  `docs/methods.html` in the gh-pages branch and an example?

If anyone has suggestions let me know and I'll happily update the PR.

( re https://github.com/katspaugh/wavesurfer.js/issues/698 )


src/mediaelement.js
```diff
@@ -60,6 +60,43 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
         this.setPlaybackRate(this.playbackRate);
     },
 
+    loadElt: function (elt, container, peaks) {
+        var my = this;
+
+        var media = elt;
+        media.controls = this.params.mediaControls;
+        media.autoplay = this.params.autoplay || false;
+        media.style.width = '100%';
```
katspaugh: Maybe we shouldn't change the way an existing element looks? It can be done in the app's CSS.

src/mediaelement.js
```diff
@@ -60,6 +60,43 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
...
+
+        media.addEventListener('timeupdate', function () {
+            my.fireEvent('audioprocess', my.getCurrentTime());
+        });
+
+        var prevMedia = container.querySelector(this.mediaType);
+        if (prevMedia) {
+            container.removeChild(prevMedia);
+        }
+        container.appendChild(media);
```
katspaugh: Maybe just keep the media element where it was initially?

src/mediaelement.js
```diff
@@ -60,6 +60,43 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
         this.setPlaybackRate(this.playbackRate);
     },
 
+    loadElt: function (elt, container, peaks) {
```
katspaugh: This method almost entirely duplicates the regular `load` method. Can you share the code between the two methods somehow? Maybe create another function and call it from both methods.

src/mediaelement.js
```diff
@@ -89,7 +126,8 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
     },
 
     seekTo: function (start) {
-        if (start != null) {
+        var seekAmount = Math.abs(start - this.getCurrentTime() );
+        if (start != null && !isNaN(start) && (start === 0 || seekAmount > 1) ) {
```
katspaugh: Rationale for this change?

src/wavesurfer.js
```diff
@@ -382,6 +382,22 @@ var WaveSurfer = {
         }
     },
 
+    loadExistingMediaElement: function (elt, peaks) {
```
katspaugh: Again, almost a 1-to-1 copy of the `loadMediaElement`. The only difference is which backend method it calls.

katspaugh:
> What is the best way to document this? Should I add a line to docs/methods.html in the gh-pages branch and an example?

Yeah, sure. An example with an existing video element would be nice, since it's asked about a lot.

Thank you very much!

therewasaguy:
Thanks for your quick feedback, @katspaugh! All good points, and I'm sorry I should have updated some of these before sending the PR.

I will update this all soon, and I'll add a demo with video.

therewasaguy:
Updated and added an example using a `<video>`.

While consolidating the code, I was thinking... why not could just use `load` for existing media elements, rather than create a separate method? If the first argument is a media element, load handles it correctly. Does that sound like a good approach?

katspaugh:
Thanks!

> Does that sound like a good approach?

Yep, sounds like a good idea!

src/mediaelement.js
katspaugh: Just to clarify: this will move the existing media element from its initial position in DOM to the wavesurfer.js' container. Are you sure we want this?

src/mediaelement.js
```diff
@@ -89,7 +124,7 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
     },
 
     seekTo: function (start) {
-        if (start != null) {
+        if (start) {
```
katspaugh: Why did you change this? If start is `0` we want to go to the beginning, not to the current position.

src/mediaelement.js
```diff
@@ -104,7 +139,7 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
      * relative to the beginning of a clip.
      */
     play: function (start, end) {
-        this.seekTo(start);
+        start && this.seekTo(start);
```
katspaugh: The same comment.

src/wavesurfer.js
```diff
@@ -200,8 +200,8 @@ var WaveSurfer = {
         if (paused) {
             this.params.scrollParent = false;
         }
+        this.drawer.progress(progress);
         this.backend.seekTo(progress * this.getDuration());
-        this.drawer.progress(this.backend.getPlayedPercents());
```
katspaugh: Again, an unrelated and risky change. The backend is supposed to be the data source for the drawer.

src/wavesurfer.js
```diff
@@ -363,9 +363,31 @@ var WaveSurfer = {
...
+    /**
+     *  Either create a media element, or load
+     *  an existing media element.
+     *  @param  {String|HTMLElement} url_or_elt Either a path to a media file,
+     *                                          or an existing HTML5 Audio/Video
+     *                                          Element
+     *  @param  {Array}            [peaks]     Array of peaks. Required to bypass
+     *                                          web audio dependency
+     */
+    loadMediaElement: function (url_or_elt, peaks) {
```
katspaugh: Please use camelCase for variable names.

src/mediaelement.js
therewasaguy: Good call, I will revert this. While I was testing, setting media element .currentTime was always going back to zero, and I made this while troubleshooting before I realized it was due to the server (python simple server) and had nothing to do with the library. I should have reverted this and remembered that from [your stack overflow post](http://stackoverflow.com/questions/9563887/setting-html5-audio-position#comment12125431_9564020)

src/mediaelement.js
therewasaguy: With the new changes, the `mediaElement.load` method is only called if a media element doesn't already exist, while `mediaElement.loadElt` is called if an element does exist, and doesn't mess with the DOM.

katspaugh:
Thanks a lot, @therewasaguy!

PR #710, opened 2016-04-27
Flip the in/out regions processing conditionals

chinshr:
This will fix a bug where out/region-out events are fired once the in/region-in processes.


katspaugh:
Thanks!

PR #714, opened 2016-05-05
allow the WebAudio backend to accept prerendered peaks just like the MediaElement backend

cdelguercio:
I added the ability to pass peaks from the wavesurfer.load() method to the WebAudio media backend. I haven't done much real testing on anything other than the site that I'm currently using wavesurfer on.


katspaugh:
Thanks for the contribution, but what is the point of passing the peaks if the audio is decoded anyway?

cdelguercio:
Performance, it take around 5ms to run getPeaks on my (fast) laptop. If you have 20+ songs on a page and you're on a phone the time saving could easily reach a few hundred ms. I'm prerendering the peaks serverside and then passing them to the client with the songs. The peaks themselves are so small that the added network time is easily sub-millisecond.

thijstriemstra:
@cdelguercio there are some build errors..

dist/wavesurfer.js
```diff
@@ -847,6 +851,10 @@ WaveSurfer.WebAudio = {
      * of peaks consisting of (max, min) values for each subrange.
      */
     getPeaks: function (length) {
+        if( this.peaks ) {
```
katspaugh: Please change to `if (this.peaks) {`.

dist/wavesurfer.js
```diff
@@ -333,18 +333,13 @@ var WaveSurfer = {
...
 
-    /**
-     * Loads audio using Web Audio buffer backend.
-     */
-    loadBuffer: function (url) {
-        this.empty();
-        // load via XHR and render all at once
-        return this.getArrayBuffer(url, this.loadArrayBuffer.bind(this));
+    loadWebAudio: function(url, peaks) {
+        return this.loadBuffer(url, peaks);
```
katspaugh: What is the point of this function if it just calls another function with the same signature?

src/webaudio.js
```diff
@@ -184,6 +184,10 @@ WaveSurfer.WebAudio = {
      * of peaks consisting of (max, min) values for each subrange.
      */
     getPeaks: function (length) {
+        if( this.peaks ) {
```
katspaugh: `if (this.peaks) {`

dist/wavesurfer.js
thijstriemstra: was thinking the same!

src/wavesurfer.js
```diff
@@ -283,17 +283,17 @@ var WaveSurfer = {
     /**
      * Internal method.
      */
-    loadArrayBuffer: function (arraybuffer) {
+    loadArrayBuffer: function (arraybuffer, peaks) {
         this.decodeArrayBuffer(arraybuffer, function (data) {
-            this.loadDecodedBuffer(data);
+            this.loadDecodedBuffer(data, peaks);
```
katspaugh: I wonder if we could make it work in the same way as Media Element. I.e. draw the waveform first, and then decode the buffer.

cdelguercio:
@katspaugh you read my mind. I'm actually working on a PR right now where we allow the user to JUST load the peaks, and then load the audio afterwards. This would copy what soundcloud does where the audio isn't loaded until the user actually hits play. The benefit would be a massive decrease in load time of pages that have many songs on them. I plan to submit a new PR later today.

Regarding the extra function call: I put that in for readability. When I first read through the code there it was a little confusing because the 'loadMediaElement' case called this.loadMediaElement but the 'webAudio' case called this.loadBuffer. (loadBuffer has a very generic meaning until you understand that buffer is the audio info that if owned by the audio backend) I'm from the school of thought where it's okay to write a little bit of extra code in order to make the semantics work out better. I can change this if it bothers everyone.

katspaugh:
Awesome! Then I'll wait for the next PR.

Regarding the function name, we could rename the `loadBuffer` to `loadWebAudio`. I don't think it was called directly from app code.

thijstriemstra:
-1 on renaming `loadBuffer`, it'll break existing code out there.

katspaugh:
Closing in favour of #733 or #734.

PR #720, opened 2016-05-11
MediaElement#getDuration throwing TypeError fixed

### LINKED ISSUE #719: Safari 9.1 throws `TypeError: Not enough arguments` in `#getDuration`
To be honest, I am not sure how to reproduce this issue. We use WaveSurfer on our project (still not public) and I get this error on local environment (staging and production are fine). However, the issue is that [`#getDuration`](https://github.com/katspaugh/wavesurfer.js/blob/master/src/mediaelement.js#L67) calls [`#end`](https://developers.whatwg.org/the-video-element.html#dom-timeranges-end) on [`media.seekable`](https://developers.whatwg.org/the-video-element.html#dom-media-seekable) and it expects to receive on parameter. Since there is nothing passed now, I expect that `0` is what should be in `#getDuration`.

frzng:
WaveSurfer.MediaElement#getDuration throwed `TypeError: Not enough
arguments` in Safari 9.1. It also violated contract specified in [HTML Time ranges](https://developers.whatwg.org/the-video-element.html#dom-timeranges-end).

This should fix #719.

### DIFF
--- a/src/mediaelement.js
+++ b/src/mediaelement.js
@@ -67,7 +67,7 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
     getDuration: function () {
         var duration = this.media.duration;
         if (duration >= Infinity) { // streaming audio
-            duration = this.media.seekable.end();
+            duration = this.media.seekable.end(0);
         }
         return duration;
     },


katspaugh:
Thanks! :thumbsup:

PR #733, opened 2016-05-28
Load peaks only

cdelguercio:
So I created a new function called loadPeaksOnly, which is poorly named, but what it does is save the url string temporarily (but doesn't download the audio) and then draw the peaks that the user supplied. The audio is only downloaded in the event of a seekTo or play action. This is very rough hacky code, and I'd appreciate some input on how to design it better. My javascript skills are only so-so.

wavesurfer.loadPeaksOnly('/api/v1/song/1/', [0.5, 0.6, 0.7, 1.0, 0.1, 0.3])
//then the waveform is visible
// once the user clicks on the waveform or initiates a play action, the audio will be downloaded from the given url ('/api/v1/song/1/' in this case)


src/wavesurfer.js
```diff
@@ -36,6 +36,7 @@ var WaveSurfer = {
         mediaType     : 'audio',
         autoCenter    : true
     },
+    tempUrl           : null,
```
katspaugh: This property will be defined on the prototype. Every time you assign a value, it will be shared across all wavesurfer instances.

dist/wavesurfer.js
```diff
@@ -149,6 +150,17 @@ var WaveSurfer = {
     },
 
     play: function (start, end) {
+        if (!this.backend.buffer) {
+            this.load(this.tempUrl, null);
+            this.once('ready', (function() {
+                this._play(start, end);
+            }).bind(this));
```
katspaugh:
```
this.once('ready', this._play(this, start, end));
```

dist/wavesurfer.js
```diff
@@ -149,6 +150,17 @@ var WaveSurfer = {
     },
 
     play: function (start, end) {
+        if (!this.backend.buffer) {
+            this.load(this.tempUrl, null);
```
katspaugh: You don't need to pass `null`, simply don't pass anything.

katspaugh:
Thanks for the contribution! I've used your ideas in #734.

Issue #727, opened 2016-05-17
Wrong file plays after loading multiple times in short succession


rijk:
Great library, thanks so much for putting this out.

I have a list of tracks the user can loop through, each time triggering a `WaveSurfer.load()` call. I've noticed a bug that I can reproduce by pressing the next button at just the 'right' intervals — the song that ends up playing is not the last one loaded, but the one loaded before it.

Unfortunately, this is deeply embedded in an application, so I cannot share a quick reproduction repo, but I've been plowing through the source code and adding `console.log` statements to track down this issue.

I've found out that the problem occurs when the `load` call is executed, while the `decodeArrayBuffer` is still running. The `tmpEvents` are cleared, which is why [the callback](https://github.com/katspaugh/wavesurfer.js/blob/master/dist/wavesurfer.js#L405-L407) is not executed. So far, so good, but when the aforementioned callback is re-scheduled and executed after loading the *last* file, [the buffer that's loaded](https://github.com/katspaugh/wavesurfer.js/blob/master/dist/wavesurfer.js#L296) is actually the data from the *second-to-last file*! This is really strange, and I am not sure why this happens yet. Probably as a result of [the `this` binding](https://github.com/katspaugh/wavesurfer.js/blob/master/dist/wavesurfer.js#L295-L297) that's going on [everywhere](https://github.com/katspaugh/wavesurfer.js/blob/master/dist/wavesurfer.js#L347).

What I've also found, is that when I change [this code](https://github.com/katspaugh/wavesurfer.js/blob/master/dist/wavesurfer.js#L400-L407) from:

        this.backend.decodeArrayBuffer(
            arraybuffer,
            this.fireEvent.bind(this, 'decoded'),
            this.fireEvent.bind(this, 'error', 'Error decoding audiobuffer')
        );
        this.tmpEvents.push(
            this.once('decoded', callback)
        );

to:

        this.backend.decodeArrayBuffer(
            arraybuffer,
            callback,
            this.fireEvent.bind(this, 'error', 'Error decoding audiobuffer')
        );

(making it call the callback directly, instead of via an event), the loading issue is fixed and the audio correctly plays the last loaded version. But now, the waveform shows both waves at the same time.

So, I'm submitting this ticket, because I think someone who has a little more insight into the events approach chosen has to look at what the best fix for this issue would be.

katspaugh:
Thanks for the report, I'll take a look.

katspaugh:
[...]

PR #734, opened 2016-05-29
Allow pre-decoded peaks with Web Audio

katspaugh:
@cdelguercio I like your idea to load the audio on demand (from #733). I suggest we do it every time peaks are passed with the Web Audio backend.

### DIFF
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -141,6 +141,7 @@ var WaveSurfer = {
     },
 
     play: function (start, end) {
+        this.fireEvent('interaction', this.play.bind(this, start, end));
         this.backend.play(start, end);
     },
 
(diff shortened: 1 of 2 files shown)


src/wavesurfer.js
```diff
@@ -389,14 +404,18 @@ var WaveSurfer = {
     },
 
     decodeArrayBuffer: function (arraybuffer, callback) {
+        this.arraybuffer = arraybuffer;
+
         this.backend.decodeArrayBuffer(
             arraybuffer,
-            this.fireEvent.bind(this, 'decoded'),
+            (function (data) {
+                if (this.arraybuffer == arraybuffer) {
```
katspaugh: This fixes #727.

cdelguercio:
Why not do it for media element as well? or allow a flag to be passed to activate the feature? Really it should be an options parameter when setting up the wavesurfer object.

Issue #741, opened 2016-06-06
Scheduled pause (play range) doesn't update start position


cwleonard:
I'm not sure if this is intentional or not, but when playing a range of the audio the start position is not updated when the scheduled pause takes place. This results in the next call to `play` (without a range specified) picking up at a previous start position.

I fixed it in my fork by calling `pause()` in webaudio.js when the time is >= scheduledPause instead of calling `setState` and `fireEvent` directly.

If the original behavior is intentional, I would like to request an option to change that behavior. If it was not intentional, I can submit a pull request.

Thanks.

katspaugh:
Looks like a bug. A PR would be nice! Calling `pause` in this case is fine, but I think the same code also runs when the audio is finished. I might be wrong though. Let's talk in the PR.

PR #742, opened 2016-06-06
Call pause() when scheduled pause time is reached

cwleonard:
When the scheduled pause time is reached (because only a range of the sound is playing), call `pause()` instead of setting the state and firing the pause event. This ensures that `startPosition` will be recalculated to the time of the pause event.

Addresses issue #741.


katspaugh:
Looks good to me! Thanks!

Issue #543, opened 2015-10-06
Media Element and iOS


arian1123:
Fantastic library btw! 
So Im using AJAX to create instances of the wavesurfer class through an AJAX request and loading the file as a MediaElement. This is works extremely well on computers and Android phones, however it does not fully work on iOS. When loading the page in iOS the waveforms dont display, and then only fade in when I click my play button. I tried locating where things are going wrong and it seems to me that the function basically pauses as it hits the .init() function. For whatever reason it responds properly when clicking play or trying to play a different track that I have already loaded. Is there a known bug for something like this? I know that on wavesurfer-js.org the waveform initializes fine, and in trying to reverse engineer I cannot find why my instances dont seem to initialize without clicking play. 
Any help here would be really appreciated! Thanks

EDIT: So I tried changing the MediaElement to WebAudio and it initialized as needed. There are other issues with WebAudio working properly, but is there any history of issues like this being tied to MediaElement?

katspaugh:
Duplicate of #509

arian1123:
@katspaugh read that thread already, my issue relates directly to rendering of the waveforms, not actually playing them. As mentions renders with WebAudio not MediaElement. Or am I missing something here.

katspaugh:
It's something different then, sorry.

thijstriemstra:
Ok. What browser and version of ios/wavesurfer were you using?

arian1123:
iOS 9, also tried on 8, with the newest version of Safari. Im using the last version uploaded of the wavesurfer.min.js file. Thanks

arian1123:
As mentioned before the strange part is that the wavesurfer object only seems to initalize (or trigger the 'ready' event) when I click play when using MediaElement, which makes little sense to me. It renders fine without needing to click play using WebAudio, however this is suboptimal since obviously it wont load files as quick as using MediaElement.

thijstriemstra:
The need to click to play is expected on iOS  isn't it?

arian1123:
From reading previous threads that what Im not sure about. I dont want the track to start playing when I load the page. I simply want it to display the waveform. Is clicking play needed to do that? Also if it is, would you know a workaround for example to would allow me to create a mimicked image of the waveform as placeholder? Thanks

thijstriemstra:
No clicking isn't needed to do that, true.

arian1123:
[...]

PR #744, opened 2016-06-09
fix media element ios issue #543 where peaks didnt draw

therewasaguy:
fix for https://github.com/katspaugh/wavesurfer.js/issues/543


katspaugh:
Cool, thanks!

Issue #696, opened 2016-04-08
Multi canvas approach for timeline plugin?


chinshr:
On longer waveforms with timelines it can happen that the timeline is not rendered due to limitations on the canvas size. It would be great to get some of the excellent multi-canvas magic by @chrisparton1991 to work in the timeline plugin. 

BTW I integrated MultiCanvas into the minimap plugin and it works well, of course it is less prone to run out of "canvas space". But if anyone is interested, I can share my code in a PR?

katspaugh:
Sure, a PR is always appreciated!

chrisparton1991:
I'll take a look at this at some point, but it may be a while before I can get around to it due to other commitments.

chrisparton1991:
I've taken a look at the code for the timeline plugin. I think it'll be easier to convert it over to MultiCanvas than the main renderer was.

The only potentially tricky bit is the text that's rendered. I've done some tests and it looks like rendering text partially past the right end of a canvas, or using negative coordinates to render partially before the left end of a canvas works fine, so it shouldn't be a huge issue to render text that falls between two canvases.

chinshr:
Oh, that sounds great. It would take me forever to figur this out. I can help test and code review if you like?

chrisparton1991:
Sure, that would be great when I get around to it @chinshr.

I think you'd be able to implement the MultiCanvas renderer too, I'm definitely no mathematical genius or anything. I had to discard my first attempt because it was garbage, but the second attempt went a lot more smoothly because of it.

I just hope nobody asks for some kind of MultiCanvas thing that requires drawing lines more complex than the current waveforms, I'd be lost on how to implement that!

chinshr:
Ok, will take a look at it. Thanks.

chrisparton1991:
Sure thing @chinshr. I'm happy to review and test your code. I'm also happy to have a go at implementing it if the need arises.

chinshr:
You mentioned that you had looked at how to implement drawing labels. If there is any approach worth posting here, I would pick it up. Thanks.

chrisparton1991:
Hey @chinshr, I just wrote a Codepen (http://codepen.io/chrisparton1991/pen/bpKoqm) that shows how to render text across canvases. I've added comments to (hopefully) make the code clearer. Let me know if it makes sense and if you still want to have a go at implementing this feature.

chinshr:
[...]

PR #765, opened 2016-07-06
Enhance timeline plugin to support multi-canvas rendering

chrisparton1991:
A merge request to cover #696. I have endeavoured to maintain backwards compatibility with the `Canvas` renderer. I have attached a [multicanvas-timeline-tests.zip](https://github.com/katspaugh/wavesurfer.js/files/349198/multicanvas-timeline-tests.zip) file containing my changes, along with a test html and js file in the `src` directory that can be run to visually observe the results.

I don't personally use the timeline plugin, so there may be some edge cases that I'm not aware of. I've changed some function names, e.g. `drawTimeCanvas` -> `drawTimeCanvases`, so that's a potential breaking change.

@chinshr and @jeremylach, it would be great if you could download the attached zip file and see if the changes meet your needs before the changes are merged (assuming they are ☺).

### DIFF
--- a/plugin/wavesurfer.timeline.js
+++ b/plugin/wavesurfer.timeline.js
@@ -19,6 +19,9 @@ WaveSurfer.Timeline = {
         }
 
         this.width = drawer.width;
+        this.pixelRatio = this.drawer.params.pixelRatio;
+        this.maxCanvasWidth = drawer.maxCanvasWidth || this.width;
+        this.maxCanvasElementWidth = drawer.maxCanvasElementWidth || Math.round(this.maxCanvasWidth / this.pixelRatio);
         this.height = this.params.height || 20;
         this.notchPercentHeight = this.params.notchPercentHeight || 90;
         this.primaryColor = this.params.primaryColor || '#000';
@@ -31,9 +34,10 @@ WaveSurfer.Timeline = {
         this.primaryLabelInterval = this.params.primaryLabelInterval;
         this.secondaryLabelInterval = this.params.secondaryLabelInterval;
         this.formatTimeCallback = this.params.formatTimeCallback;
+        this.canvases = [];
 
         this.createWrapper();
-        this.createCanvas();
+        this.createCanvases();
         this.render();
 
         wavesurfer.drawer.wrapper.onscroll = this.updateScroll.bind(this);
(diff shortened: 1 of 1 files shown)


chrisparton1991:
@katspaugh looks like there's an issue with the CI job that's unrelated to my changes.

katspaugh:
Looks good to me. Thanks a lot!

chrisparton1991:
No problem :) I've only tested on a Retina display. It would be good for something to test it on a regular DPI monitor to ensure the pixelRatio maths is correct.

jeremylach:
This fixes the issue of the timeline disappearing when zoomed in too far... but it breaks the scrolling of the timeline.  This was not happening with the older version of the plugin.  The time markers should move with the waveform area but currently, they stay in place while the waveform area scrolls by.   @chrisparton1991 Thanks for your efforts

chrisparton1991:
Thanks for your feedback @jeremylach. Did you test this using my test case, or with your own code? The timeline scrolls correctly for me in the `Multicanvas Timeline: Scrolling` test. See this [Youtube video](https://www.youtube.com/watch?v=XZqxA119DRs) of how it looks for me on Chrome.

Note that the timeline doesn't start scrolling until the time indicator reaches the middle of the visible waveform (around 5 seconds in).

Questions:
- Do you not see this behaviour, or is this behaviour incorrect?
- What browser are you using?
- Are you using a high DPI monitor (e.g. Retina display)?

jeremylach:
1. I do not see this behavior
2. I tested with my own code.  I have since tried with your test case and it worked...
3. Chrome
4. Retina, yes. 

I'll need to troubleshoot some more to see where my test case differs from yours.  I'm also using the Regions plugin.

jeremylach:
Realized my mistake.  Did not have the scrollParent option set to True.  That fixed it.

katspaugh:
OK, I guess good to merge then. Thanks!

chrisparton1991:
Thanks for testing it for me @jeremylach, glad you figured out the issue!

jeremylach:
One strange behavior I'm still seeing is that when scrollParent is initially set to false, then the user zooms, the timeline will not scroll with the waveform.  Even if the params.scrollParent is explicitly set to true at that point with code.  
According to the documentation:
> zoom(pxPerSec) – Horizontally zooms the waveform in and out. The parameter is a number of horizontal pixels per second of audio. It also changes the parameter minPxPerSec and enables the scrollParent option.

It would be nice to be able to see the entire waveform (scrollParent = false) initially, then have the timeline be scrollable along with the waveform when zooming.

chrisparton1991:
@jeremylach I agree that this sounds like something in `wavesurfer.js` itself. Do you know if this occured before this PR went in? If so, I'd recommend raising a new issue.

jeremylach:
@chrisparton1991  I did some testing.  Seems that this issue did not occur before your updates to the timeline plugin code.  When I use the older timeline plugin the timeline scrolls as expected when zoomed in.

chrisparton1991:
@jeremylach do you mind raising a new issue and referencing this PR? I'll try to take a look at this sometime soon, but I'm working on lots of stuff at the moment :)

jeremylach:
@chrisparton1991 thanks https://github.com/katspaugh/wavesurfer.js/issues/778

kylophone:
Thanks! This is great.

chrisparton1991:
No worries @kylophone, all I have to do now is fix the bugs I've caused :)

PR #769, opened 2016-07-07
Update Gruntfile.js

pitrackster:
WaveSurfer plugins currently can not be required by a tool like webpack because their dependence on wavesurfer module uses an absolute id instead of an id on the "plugin" directory


katspaugh:
Thanks!

Issue #717, opened 2016-05-11
Region drag changes its size


ghost:
I want to add some fixed regions that you can drag but not resize.

But when you drag a region to the beginning or ending of the wave it gets resized and becomes smaller, even if it's not "resizable".

katspaugh:
I think it was made on purpose as a feature. I assume, you would expect the region to simply stop moving when dragged to the edge?

Syrenne:
Yes, that's exactly what I'd like. Is it possible to do it?

katspaugh:
I think it's makes more sense than resizing, too. It should be possible by adding a check here:
https://github.com/katspaugh/wavesurfer.js/blob/0a0e6757174919f3c03bc1f7798f8f9247b04dcf/plugin/wavesurfer.regions.js#L343

```
var maxEnd = this.wavesurfer.getDuration();
var excess = this.end + delta - maxEnd;
if (excess > 0) {
    delta -= excess;
}
```

Can you test it and create a PR?

thijstriemstra:
> Can you test it and create a PR?

No feedback, closing.

PR #775, opened 2016-07-11
adding controls to ensure region size does not change and sliding stops when the sides are reached

pyros-endgame:
With regards to issue #717 I was looking into it as it was something I wanted to change too so I just went ahead with it. Hopefully that's alright.

Overall, I feel it's much more intuitive now as regions don't magically resize. This is especially true when you set resizing to false but end up dragging the region to the side as well.

### DIFF
--- a/plugin/wavesurfer.regions.js
+++ b/plugin/wavesurfer.regions.js
@@ -425,16 +425,14 @@ WaveSurfer.Region = {
 
     onDrag: function (delta) {
         var maxEnd = this.wavesurfer.getDuration();
-        
         if ((this.end + delta) > maxEnd || (this.start + delta) < 0) {
             delta = 0;
         }        
-
+        
         this.update({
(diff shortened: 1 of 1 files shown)


katspaugh:
Awesome, thanks!

plugin/wavesurfer.regions.js
```diff
@@ -424,6 +424,11 @@ WaveSurfer.Region = {
     },
 
     onDrag: function (delta) {
+        var maxEnd = this.wavesurfer.getDuration();
+        if ((this.end + delta) > maxEnd || (this.start + delta) < 0) {
+            delta = 0;
```
katspaugh: If the delta is 0, maybe just return at this point?

pyros-endgame:
Makes sense, I'll fix that.

katspaugh:
Thank you :thumbsup:

Issue #781, opened 2016-07-24
MultiCanvas Renderer and Zoom - Current Indicator place is changing


samih-dev:
Hi
Example can be found here:
http://codepen.io/SamihAbuawad/pen/yJjrrN

I'm using MultiCanvas as renderer (Thanks to @chrisparton1991), after setting the indicator to specific point in a wave while the audio is  paused (try the start of the first wave), and then trying to zoom, the indicator place is being changed.

Anyone faced the same issue? is there any workarouds/fix?

Thanks! 
Samih.

chrisparton1991:
Sounds similar to #778, hopefully they're caused by the same thing.

chrisparton1991:
Can reproduce using v1.1.2 using MultiCanvas renderer, which is a relatively old release. This likely isn't the same cause as #778.

chrisparton1991:
It behaves the same until multiple canvases are used by the multicanvas renderer under the hood. Something isn't taking the width of all the canvases into consideration.

This means that if you set `maxCanvasSize: 32000`, you will be able to zoom in further before this behaviour is exhibited. The tradeoff is that some browsers will not render canvases that large.

I'm working on finding a proper solution at any rate.

chrisparton1991:
Bars don't seem to be affected. Setting `barWidth: 1` is a workaround for this issue, if you can live with having bars. This makes me think this issue is in the MultiCanvas `drawWave` method.

chrisparton1991:
Found the issue, I was incorrectly dividing by the canvas size in a length calculation in the multiCanvas `drawWave` method.

chrisparton1991:
I've been trying to figure this one out for the past few hours, it's a tricky one.

The cause is essentially that the multicanvas waveform rendering code divides up the waveform into equal chunks per canvas, but the canvases aren't equal width. If the waveform width is 1000 and maxCanvasSize is 400, you'll have canvases of widths 400, 400 and 200. These three canvases will be apportioned the same amount of peak data to render, which is incorrect.

A solution still eludes me.

katspaugh:
Can we perhaps make the last canvas also 400 and have a common container that would cut off the remaining 200 px (via `overflow: hidden`)?

chrisparton1991:
I think I have a solution, but I'll have to continue looking at it after work tomorrow. Baaically, instead of having that 400/400/200 split, I divide then equally (333.3/333.3/333.3).

samih-dev:
[...]

PR #785, opened 2016-07-28
#781 MultiCanvas Renderer and Zoom - Current Indicator place is changing

chrisparton1991:


chrisparton1991:
Please hold off on this, it worked on my work laptop, but I'm having issues with my personal laptop.

src/drawer.multicanvas.js
```diff
@@ -217,33 +218,33 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
             absmax = -min > max ? -min : max;
         }
 
-        this.drawLine(length, peaks, absmax, halfH, offsetY);
+        this.drawLine(peaks, absmax, halfH, offsetY);
```
chrisparton1991: I moved `length` calculation into the `drawLineToContext` method.

src/drawer.multicanvas.js
```diff
@@ -169,11 +176,6 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
 
         var scale = length / width;
 
-        this.canvases[0].waveCtx.fillStyle = this.params.waveColor;
-        if (this.canvases[0].progressCtx) {
-            this.canvases[0].progressCtx.fillStyle = this.params.progressColor;
-        }
-
```
chrisparton1991: Dead code.

src/drawer.multicanvas.js
```diff
@@ -52,28 +52,30 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
...
 
      addCanvas: function () {
         var entry = {};
-        var leftOffset = this.maxCanvasElementWidth * this.canvases.length;
 
         entry.wave = this.wrapper.appendChild(
             this.style(document.createElement('canvas'), {
                 position: 'absolute',
                 zIndex: 1,
-                left: leftOffset + 'px',
```
chrisparton1991: I now calculate the left offset in `updateDimensions`. This is because the left offset will change as the canvases resize, whereas only the last canvas used to change size, and all of the left offsets were constant.

src/drawer.multicanvas.js
```diff
@@ -52,28 +52,30 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
             this.removeCanvas();
         }
 
+        var leftOffset = 0;
         for (var i in this.canvases) {
             // Add some overlap to prevent vertical white stripes, keep the width even for simplicity.
-            var canvasWidth = this.maxCanvasWidth + 2 * Math.ceil(this.params.pixelRatio / 2);
+            var canvasWidth = Math.floor(totalWidth / requiredCanvases * this.params.pixelRatio),
+                canvasOverlap = 2 * Math.ceil(this.params.pixelRatio / 2),
+                actualCanvasWidth = canvasWidth;
```
chrisparton1991: This is the main part of the change. Make the canvases equal width (or very close, the last one will be slightly smaller as it doesn't have an overlap).

chrisparton1991:
This proved to be a lot trickier than I thought. My solution is working, but there are slight issues for extreme zoom values or for very small `maxCanvasSize` values (e.g. 100 or less). The main issue is that the midpoint line may be a few pixels too small at the rightmost end of the last canvas, but under typical scenarios this shouldn't be an issue.

chrisparton1991:
Zip file containing a test case with the fix. Would be good for @samih-dev to download this zip and take a look.
[fix-test.zip](https://github.com/katspaugh/wavesurfer.js/files/389965/fix-test.zip)

chrisparton1991:
@katspaugh, I'm about to open another merge request. I've been having a terrible time with this approach, and I think I've found a simpler solution that solves both the indicator place issue and the seeking accuracy issue. I'll describe the solution in the new PR and attach a test case.

I'm going to close this PR for the time being. I can always reopen it if need be.

Issue #778, opened 2016-07-21
[timeline] Multicanvas Timeline Zoom Scrolling Issue


samedayjeremy:
When scrollParent is initially set to false, then the user zooms, the timeline will not scroll with the waveform. Even if the params.scrollParent is explicitly set to true at that point with code.

It would be nice to be able to see the entire waveform (scrollParent = false) initially, then have the timeline be scrollable along with the waveform when zooming.  This worked with the non-multicanvas timeline renderer but has since broken due to this pull request.
[https://github.com/katspaugh/wavesurfer.js/pull/765#issuecomment-231247869](url)

chrisparton1991:
I'll try and take a look at this tonight.

jeremylach:
It seems that this line is not doing anything

```
//From updateScroll in wavesurfer.timeline.js
this.wrapper.scrollLeft = this.drawer.wrapper.scrollLeft;
```

The drawer.wrapper.scrollLeft property updates as it scrolls but the wrapper.scrollLeft value remains 0

chrisparton1991:
That's correct @jeremylach. The (main) issue was that the canvases weren't being reallocated upon zoom, so the zoomed timeline was drawn to the first canvas, but there were no other canvases available. The scrollLeft wasn't updating because there was no scrollable content.

I also fixed a couple of other issues to do with canvas positioning and sizing. Can you please take a look at the attached 
[timeline-scroll-fix-test.zip](https://github.com/katspaugh/wavesurfer.js/files/412482/timeline-scroll-fix-test.zip) and let me know if it works for you?

If so, I'll raise a PR with the fixes.

jeremylach:
It is working great!! This also fixed another issue where timeline numbers were overlapping if the wavesurfer was zoomed in far and scrolled through for a while.  

Thanks for your work.

chrisparton1991:
Awesome, I'll raise a PR shortly. Yeah, that overlap issue was one of the other things I found. I had a dodgy length calculation for the last canvas that didn't take into account the scroll width of the container.

Apologies for causing the issues in the first place :) I think once this and the other PR are merged in, MultiCanvas will be free of known issues!

chrisparton1991:
PR raised @jeremylach, #792.

PR #789, opened 2016-08-09
781 MultiCanvas Renderer and Zoom - Current Indicator place is changing

chrisparton1991:
This PR is a replacement for #785, which caused other issues. The PR pertains to #781.

Instead of dividing the waveform up into equal sizes as I tried in #785, I keep my current apportioning algorithm the same. What changes is that I record a `start` and `end` value against each canvas entry. These values are decimals between 0 and 1, and they represent the location within the waveform.

This allows me to render the correct peak data against each canvas, whereas I previously just divided the peaks evenly amongst the canvases. This was wrong because the last canvas can be much smaller than the other canvases, which screws things up.

Attached is another runnable test with the new fix: [fix-test-2.zip](https://github.com/katspaugh/wavesurfer.js/files/408957/fix-test-2.zip). Could @samih-dev, @moyara and possibly @jeremylach take a look at the new zip file and see if it solves your issues? I hope it does, I'm looking forward to this issue being closed ;)

### DIFF
--- a/src/drawer.multicanvas.js
+++ b/src/drawer.multicanvas.js
@@ -66,8 +66,8 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
     },
 
      addCanvas: function () {
-        var entry = {};
-        var leftOffset = this.maxCanvasElementWidth * this.canvases.length;
+        var entry = {},
+            leftOffset = this.maxCanvasElementWidth * this.canvases.length;
 
         entry.wave = this.wrapper.appendChild(
             this.style(document.createElement('canvas'), {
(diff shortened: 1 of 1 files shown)


jeremylach:
I took a look at this test, though I never experienced the issue which this is supposed to fix, the test seems to work as I'd expect.  Zooming, moving the playhead around, everything plays and stays in sync.

moyara:
Thanks @chrisparton1991 It's working fine! Awesome!

chrisparton1991:
Thanks for the feedback, it's a relief to hear that it's working for you both!

chrisparton1991:
I'd like @samih-dev to take a look at this, but if he doesn't respond in a day or so, feel free to review @katspaugh (and anyone else who's interested).

Do you know if this PR fixes #778, @jeremylach?

jeremylach:
@chrisparton1991  this does not seem to fix #778 unfortunately.

chrisparton1991:
I think we can just go ahead and review this @katspaugh.

katspaugh:
Merged, thanks a lot!

Issue #728, opened 2016-05-21
Add my project to your list please


avesgit:
https://en.inettools.net/app/audio/index/1 CutMp3 Online

chrisparton1991:
Hey @avesgit, I've added a pull request (#793) to add your project to the "Commercial" section of the projects list.

Your app works well, given my limited testing. Good stuff :)

A couple of suggestions for your app:
- Hide your `Play/Pause` button initially, and show it inside a `ready` event callback, e.g.:
```
wavesurfer.on('ready', function () {
    $('#my-button').show();
});
```
  I can currently click the button before the waveform has loaded, and a JavaScript exception is thrown.
- Change `Sing In` text to `Sign In`.

chrisparton1991:
Added :)

PR #793, opened 2016-08-12
Add CutMp3 Online to Projects list

chrisparton1991:
Add project request from #728.

Project does indeed use wavesurfer, and the page comes up clean when scanned with VirusTotal.


katspaugh:
Thanks! 🙇

PR #794, opened 2016-08-12
Add Electro and Sparkled to Projects list

chrisparton1991:
Added Electro by @jspear29, and Sparkled by yours truly :)


katspaugh:
Thanks! Sparked looks awesome, btw!

chrisparton1991:
Thanks, I'm pretty happy with how it turned out. I haven't worked on it in a while though. My plan was to use it to create a sequenced Christmas light display for this year, powered by an Arduino Due.

I ran into a roadblock, however. The cable length between the Arduino and the some of the strips was too long and the signal got degraded.

In order to progress, I need to use eight Arduinos that will sit right next to the strips, and communicate/sync wirelessly. Sounds like a real pain to get working, but I'll see how I go haha.

Issue #780, opened 2016-07-22
Zooming in and out on wave makes audio stutter 


anatoliyg:
I noticed that while zooming in, this.seekAndCenter is called. Is there a real reason to do that? I would think that on zoom, no seeking needs to be done since the audio should continue playing as is, and only the waves be redrawn

Stoyvo:
@anatoliyg Please review my merge request. Should solve the stuttering.

chrisparton1991:
@anatoliyg can you please test this once a new release has been made and close the issue if it works?

Is there an ETA on a release with this fix, @katspaugh?

katspaugh:
I've published v. 1.1.12.

thijstriemstra:
No feedback, closing, considering it fixed.

PR #803, opened 2016-08-23
Only recent on zoom, removed seek

Stoyvo:
Zoom should not modify the play position. Removing seek greatly increases performance while zooming.


Stoyvo:
Related issue: https://github.com/katspaugh/wavesurfer.js/issues/780

katspaugh:
Great, thank you!

PR #806, opened 2016-08-25
Fix Play Percentage on Zoom when audio paused.

Stoyvo:
Bug possibly caused by: https://github.com/katspaugh/wavesurfer.js/pull/803

By not seeking, the play percentage is not updated. Instead, we call it directly to avoid seekTo overhead.


katspaugh:
Thanks!

Issue #814, opened 2016-08-31
Pre-generated waveform on MultiCanvas and zoom doesn't fill out each canvas


moyara:
Hi I'm using multicanvas and pre-generated waveform to deal with long audio file.
Below is the options

```
this.player = WaveSurfer.create({
      container: $element.selector,
      height: 100,
      cursorColor: 'red',
      backend: 'MediaElement',
      renderer: 'MultiCanvas'
    });

this.player.load(url, waveform);

this.player.on('ready', ()=>{
      this.player.zoom(20);
      this.setVolume(100);
    });
```

When it comes with multiple canvas, each wave doesn't stretch to its sub canvas.

<img width="1163" alt="2016-08-31 1 46 26" src="https://cloud.githubusercontent.com/assets/568994/18145761/5e0e0da8-6f81-11e6-930c-41ae775359ce.png">
<img width="1179" alt="2016-08-31 1 46 34" src="https://cloud.githubusercontent.com/assets/568994/18145762/5e118ea6-6f81-11e6-8d4b-e926090cb8f5.png">

It only happens with pre-generated waveform. It works fine if I load audio without waveform.
`this.player.load(url);`

chrisparton1991:
Hey @moyara, is there any chance I could get your pre-generated peak data for testing?

moyara:
Sure, this is the peak I used

[wave.txt](https://github.com/katspaugh/wavesurfer.js/files/448411/wave.txt)

kebers:
Hi @chrisparton1991, 
I ran into the exact same problem - when start to zoom into the waveform.
When it's initially rendered to fillParent, evereything is fine - but as soon as zooming in, it's starts to draw huge gaps between pieces of the waveform. With my peakdata (displaying 4 hours)  these gaps are wider than the browser's viewport-width, see screenshots.

Additionally I'm using splitChannels. I've attached screenshots and pre-rendered peaks in the zip-file.

[wavesurfer-issue-814.zip](https://github.com/katspaugh/wavesurfer.js/files/450211/wavesurfer-issue-814.zip)

wavesurfer.init({
	        container: document.querySelector('#waveform'),
	        backend: 'MediaElement',
	        mediaType:'video',
		splitChannels: true,
		renderer: 'MultiCanvas',
		height: 70,
		autoCenter: true
	    });

kebers:
Little add-on:
Functionality is broken in v1.1.12 but works with version 1.1.4.
[...]

PR #815, opened 2016-09-02
814: Pre-generated waveform on MultiCanvas and zoom doesn't fill out each canvas

chrisparton1991:
For #814.


katspaugh:
Thanks! I'll release asap.

PR #819, opened 2016-09-08
Fixed: JS crash occurs after audio data has been decoded, if the WaveSurfer instance has been destroyed in the meantime

ebbe-brandstrup:
Also, had to manually copy changes that had been merged into upstream/master dist files only into the src files... otherwise my change would overwrite those changes... That goes for a fix from https://github.com/katspaugh/wavesurfer.js/pull/806 and the scale calculation in src/drawer.multicanvas.js.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ dist/plugin/*.js
 /coverage
 /_SpecRunner.html
 .DS_Store
+.idea
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -77,6 +77,8 @@ var WaveSurfer = {
 
         this.createDrawer();
         this.createBackend();
+
+        this.isDestroyed = false;
     },
 
     createDrawer: function () {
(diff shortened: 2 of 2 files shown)


katspaugh:
I think you confused which changes are new and correct. The dist files contained the old stuff. The source was up-to-date. So please revert your other changes and also don't commit dist files. I do that as separate commits before publishing to npm.

ebbe-brandstrup:
Thanks for getting back so quickly!

I didn't confuse the changes for the #806 issue. If you look at the files in that merge, it is the dist/wavesurfer.js only that was changed. The src file is missing the change.

I will remove the dist files and remove the scale calculation change, so it remains what it was in the src version ASAP.

katspaugh:
True, thanks!

ebbe-brandstrup:
I've updated the PR now so only the src file is there (+a small .gitignore addition) and only included the change from #806, not the difference in the ```scale``` calculation in the multi canvas drawer.

Thanks again.

katspaugh:
Thanks a lot!

Published v. 1.1.13.

PR #832, opened 2016-09-26
Correct AMD module name for plugins

### LINKED ISSUE #831: Incorrect AMD definition in plugins
For some reason the [AMD name for wavesurfer plugins is relative](https://github.com/katspaugh/wavesurfer.js/blob/master/Gruntfile.js#L61), e.g. `'./../wavesurfer'`. You can see this in the [header of the minified plugin](https://github.com/katspaugh/wavesurfer.js/blob/master/dist/plugin/wavesurfer.microphone.min.js). This leads to AMD compilation issues, e.g. using requirejs: 
```
Tracing dependencies for: site
{ [Error: ENOENT: no such file or directory, open '/var/folders/kt/8prbnh9d6lx4sr89tn_192c40000gn/T/tmpufIiIA/js/../wavesurfer.js']
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/var/folders/kt/8prbnh9d6lx4sr89tn_192c40000gn/T/tmpufIiIA/js/../wavesurfer.js',
  fileName: '/var/folders/kt/8prbnh9d6lx4sr89tn_192c40000gn/T/tmpufIiIA/js/../wavesurfer.js',
  moduleTree: [ 'wavesurferMicrophone' ] }
```

We should simply use `wavesurfer` instead, which is also used in the UMD header for the main library: https://github.com/katspaugh/wavesurfer.js/blob/master/Gruntfile.js#L51

PR following.

thijstriemstra:
fixes #831

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,4 +4,5 @@ wavesurfer.js changelog
 x.x.x (unreleased)
 ------------------
 
+- Correct AMD module name for plugins (#831)
 - Added changelog (#824)
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -58,7 +58,7 @@ module.exports = function (grunt) {
                 dest: 'dist',
                 deps: {
                     'default': [ 'WaveSurfer' ],
-                    amd: [ './../wavesurfer' ],
+                    amd: [ 'wavesurfer' ],
       	            cjs: [ 'wavesurfer.js' ],
                     global: [ 'WaveSurfer' ]
                 }


katspaugh:
Merged, , thanks!

thijstriemstra:
Seems this was previously changed by @pitrackster with https://github.com/katspaugh/wavesurfer.js/commit/367c7cf9375a3d6cf26334317fe15886b4c08ec9.

PR #834, opened 2016-09-29
Removes occasional gaps between displayed regions.

jasminegrosso:


plugin/wavesurfer.regions.js
```diff
@@ -285,9 +285,15 @@ WaveSurfer.Region = {
...
         if (this.element != null) {
+            // Calculate the left and width values of the region such that
+            // no gaps appear between regions.
+            var left = Math.round(this.start / dur * width);
+            var regionWidth =
+                Math.round(this.end / dur * width) - left;
+
             this.style(this.element, {
-                left: ~~(this.start / dur * width) + 'px',
-                width: ~~((this.end - this.start) / dur * width) + 'px',
```
katspaugh: So your PR basically removes the `Math.floor` here. It may result in huge numbers like `25.4364564575467px`. Maybe let's just apply a `toFixed` so that the accuracy is just enough to avoid gaps? Like this:
```
left: (this.start / dur * width).toFixed(4) + 'px',
width: ~~((this.end - this.start) / dur * width).toFixed(4) + 'px',
```

I think 4 digits after comma should be enough.

jasminegrosso: I'm still finding small gaps, perhaps something like

left: Math.round((this.start / dur * width).toFixed(4)) + 'px',
width: Math.round((this.end / dur * width).toFixed(4)) - (this.start / dur * width).toFixed(4) + 'px',

?

katspaugh: Well, then let's push your PR as is.

katspaugh:
Thanks! :thumbsup:

thijstriemstra:
We should also instruct people to add info about the change to the changelog.

katspaugh:
Oh, true. @jasminegrosso could you drop a line into https://github.com/katspaugh/wavesurfer.js/blob/master/CHANGES.md?

PR #841, opened 2016-10-17
Add determistic way to mute and unmute a track

vikeen:
I wanted a way to "solo" tracks in a digital audio workstation format. With the current way muting works, I was unable to do this. I needed a deterministic way to mute or unmute a track. This PR is my solution for it. I would love to hear feedback and help contribute to this library.

As for my feedback developing with wavesurfer, I found contributing difficult without an easy way to test the api. I spent several hours trying to load a media file in the test suite without success. Is there a way to do this? 

Here is an example in a project I'm working on - https://band-maker.herokuapp.com/songs/6/


vikeen:
@katspaugh, I can open up a separate issue for this, but did you have a way to test the wavesurfer api in a test environment?

thijstriemstra:
@vikeen we have a very minimal test setup but that headless browser doesn't understand the webaudio api so we need to look for alternatives. Any help is appreciated!

vikeen:
@thijstriemstra, I figured this was the case. I spent a couple hours trying to hack in the media blob data, but it sounds like even if I got that working I would have hit the unsupported problem you hit.

My initial advice would be to move towards using karma and running the tests in an actual browser. I don't know if that works on CI environment like travis though.

thijstriemstra:
But yeah, let's continue the discussion in a new ticket.

vikeen:
@thijstriemstra / @katspaugh please refer to #849 for an example of in-browser testing with karma

swampthang:
This is an awesome addition. Thanks for this. This community is moving so fast, it would be great if some time soon these additions were added to the documentation.

mspae:
@swampthang We're currently working on a version two (next branch) which has a documentation generator (es6) and up to date jsdoc comments, aside from that I think more "getting started" information and "how do I do X" type information will be added within the near future.

@vikeen @thijstriemstra I don't know if this would work with Travis/headless browsers but I was thinking of integrating visual testing for the waveform rendering (at different scroll positions, zooming, plugin combinations etc.) to get more confident builds over time. I didn't realise that headless browsers don't support webaudio (what a bummer) – (I was thinking of something like this: https://github.com/Huddle/PhantomCSS)

thijstriemstra:
@swampthang be my guest, please open a pull request.

PR #847, opened 2016-10-25
Added HTML Variable Preload Attribute

X-Raym:
Added a way to define the the preload attribute of the HTML5 audio markup.
'Metadata' and 'None' saves bandwidth, especially if there is multiple players on the page.
'Metadata' allows to have track length.
Default is left to 'auto' to avoid unexpected change for wavesurfer users.
Note that this attribute interpretation may be different on different browsers.

### DIFF
--- a/src/mediaelement.js
+++ b/src/mediaelement.js
@@ -50,13 +50,17 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
      *  @param  {HTMLElement}   container   HTML element
      *  @param  {Array}         peaks       array of peak data
      */
-    load: function (url, container, peaks) {
+    load: function (url, container, peaks, preload) {
         var my = this;
 
         var media = document.createElement(this.mediaType);
         media.controls = this.params.mediaControls;
         media.autoplay = this.params.autoplay || false;
-        media.preload = 'auto';
+        if (preload) {
+            media.preload = preload;
+        } else {
+            media.preload = 'auto';
+        }
         media.src = url;
         media.style.width = '100%';
 
--- a/src/mediaelement.js
+++ b/src/mediaelement.js
@@ -49,6 +49,7 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
      *  @param  {String}        url         path to media file
      *  @param  {HTMLElement}   container   HTML element
      *  @param  {Array}         peaks       array of peak data
+     *  @param  {String}        preload     HTML 5 preload attribute value
      */
     load: function (url, container, peaks, preload) {
         var my = this;
--- a/src/mediaelement.js
+++ b/src/mediaelement.js
@@ -56,11 +56,7 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
         var media = document.createElement(this.mediaType);
         media.controls = this.params.mediaControls;
         media.autoplay = this.params.autoplay || false;
-        if (preload) {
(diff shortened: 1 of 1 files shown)


vikeen:
For reference - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio#attr-preload

src/mediaelement.js
```diff
@@ -50,13 +50,17 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
...
 
         var media = document.createElement(this.mediaType);
         media.controls = this.params.mediaControls;
         media.autoplay = this.params.autoplay || false;
-        media.preload = 'auto';
+        if (preload) {
+            media.preload = preload;
+        } else {
+            media.preload = 'auto';
+        }
```
katspaugh: Could you change this to a more compact syntax of

    media.preload = preload == null ? 'auto' : preload;

Thanks!

src/mediaelement.js
```diff
@@ -50,13 +50,17 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
      *  @param  {HTMLElement}   container   HTML element
      *  @param  {Array}         peaks       array of peak data
```
katspaugh: Don't forget to add the jsdoc for the new argument.

X-Raym:
@katspaugh 
is it better now ? :)

katspaugh:
@X-Raym yep, thanks!

katspaugh:
Wait a sec, isn't that the internal `load` method of the media element backend? How are you using it?

X-Raym:
I noticed that if you have it on auto, the full audio is downloaded, while having it on metadata, only few ko are downloaded.
It is especially useful when you already have peaks (loaded from a file) to load with metadata only. No need to download the whole file.
It saves a lot of traffic if you have dozens of sounds on a page.

I use this method in [WaveSurfer-WP](https://github.com/X-Raym/wavesurfer-wp/blob/master/js/wavesurfer.js#L1198).

katspaugh:
Yeah, I understand the incentive. The question is, how are you going to call this internal `load` method? It's supposed to be called from the `loadMediaElement` method in the `wavesurfer.js` file, and it's not passing any additional arguments there.

X-Raym:
Oh ok, so I guess I only have to edit the loadMediaElement method the same way too ?

katspaugh:
Ideally, create an example demonstrating how to pass this parameter from JS or HTML down to the audio element. You will have to pass it through all the methods involved in the load dataflow.

X-Raym:
Oh, this should work:
http://codepen.io/X-Raym/pen/BLXBrp
I use a moded version of wavesurfer.
https://github.com/X-Raym/wavesurfer.js/blob/master/dist/wavesurfer.js

X-Raym:
@katspaugh If you prefer, I can remake a pull request with my all forked repo, as there is several files modified.

katspaugh:
The moded version you linked is correct: you are passing the preload parameter to both `wavesurfer.load` and `wavesurfer.loadMediaElement` plus to the `load` method in `mediaelement.js`. But in this PR you are passing the parameter only in `mediaelement.js` which won't work.

So yes, you need to create a PR with both files from your fork. :)

X-Raym:
@katspaugh Ok, this https://github.com/katspaugh/wavesurfer.js/pull/854/files should be good :)

PR #849, opened 2016-10-25
Replace jasmine with karma / jasmine test suite

vikeen:
This pull request is in reference to #841 

I experimented with replacing a headless jasmine test suite with karma's chrome launcher. Now tests can actually load in a browser. This gives access to the web audio api technology and allows wavesurfer to actually load audio files. The wavesurfer api appears to be fully testable, but I was unable to actually render the graphical representation of the waveform. Also, I couldn't get sound working. I don't find these relatively important, though.

* I ditched the jasmine istanbul coverage because I wasn't sure what to do with it for now. I need some advice on that front.


vikeen:
Looks like travis is going to fail until I can figure out this error.

https://travis-ci.org/katspaugh/wavesurfer.js/builds/170590640#L320

I'm actually not sure if karma chrome launcher can even work with a CI environment

katspaugh:
Merged. Thank you very much, this is an amazing move!

PR #851, opened 2016-10-30
Fixed a bug in Firefox and scrollbar usage

### LINKED ISSUE #748: Firefox: waveform scroll and drag selection
In the latest version of firefox, when I try to scroll the waveform, the drag selection happens. In Chrome and Edge, just scroll happens.

### LINKED ISSUE #830: Error While Scrolling and Region Creation
Ive been having an issue when zoomed in and trying to scroll by grabbing the scrollbar. Each time I try to scroll it creates a new region. I want it to only create a region when clicking in the canvas area, not when trying to scroll.

I have found within the region enableDragSelection there is an eventDown click event is firing for both clicking in the waveform area and clicking on the scrollbar. I then determined when clicking on the scrollbar the tag type is Wave and when clicking within the waveform area the tag type is Canvas.

So I created a method within enableDragSelection called isCanvas

```
`isCanvas: function (evt) {
    if ( evt.target.tagName == 'CANVAS' ) {
        return true;
    } 
    return false;
}`
```

I then added in an if to check when its NOT the canvas just return out of the eventDown method

```
`if (!my.wavesurfer.Region.isCanvas(e)) { return; }`
```

Is there a better way to handle this oddity? Or is it a wanted feature?

### LINKED ISSUE #843: Firefox bug - scrolling by dragging scrollbars creates a region
See screencap below. This is only an issue in Firefox. In Chrome and Safari, it works fine.

licecap

If no one gets around to this, I might fix it first if I can figure out what's going on. If anyone has an idea, let me know!

ffxsam:
In Firefox, using the scrollbars (with the mouse pointer) would trigger a region selection. The fix for this might be a bit hacky, but it works.

Tested in Mac Chrome, Firefox, Safari, and Windows Chrome, Firefox.

Fixes #830, fixes #748, fixes #843

### DIFF
--- a/plugin/wavesurfer.regions.js
+++ b/plugin/wavesurfer.regions.js
@@ -42,6 +42,12 @@ WaveSurfer.Regions = {
 
         var eventDown = function (e) {
             if (e.touches && e.touches.length > 1) { return; }
+
+            // Check whether the click/tap is on the bottom-most DOM element
+            // Effectively prevent clicks on the scrollbar from registering as
+            // region creation.
+            if (e.target.childElementCount > 0) { return; }
+
             touchId = e.targetTouches ? e.targetTouches[0].identifier : null;
 
             drag = true;


ffxsam:
@katspaugh Could you possibly publish this promptly? 😄

ffxsam:
Rationale: I realize this *could* be a Firefox bug (it's hard to say without more research), but since this is causing problems for multiple users, I figured a workaround might be a good idea.

plugin/wavesurfer.regions.js
```diff
@@ -41,7 +41,7 @@ WaveSurfer.Regions = {
         var pxMove = 0;
 
         var eventDown = function (e) {
-            if (e.touches && e.touches.length > 1) { return; }
+            if (e.touches && e.touches.length > 1 || e.target.childElementCount > 0) { return; }
```
katspaugh: Thanks for the solution!

I think it should go a separate condition for consistency across the event callbacks:

```
if (e.touches && e.touches.length > 1) { return; }
// Check whether the click/tap is on the bottom-most DOM-element
if (e.target.childElementCount > 0) { return; }
```

Please also don't add the minified files from `dist` in your commit. I'll build & commit it before an NPM release.

ffxsam: @katspaugh Branch updated!

katspaugh:
Thanks a lot, merged!

ffxsam:
@katspaugh Any time! I'll keep a close eye out for the next npm release.

katspaugh:
@ffxsam I've just published version 1.2.2.

ffxsam:
Perfect, thank you!

PR #857, opened 2016-11-08
Update CHANGES.md

X-Raym:
1.2.3 (07.11.2016)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 wavesurfer.js changelog
 =======================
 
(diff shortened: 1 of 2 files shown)


katspaugh:
I've just noticed you forgot to pass the `preload` parameter here: https://github.com/katspaugh/wavesurfer.js/blob/master/src/wavesurfer.js#L343
Can you add it in this PR?

X-Raym:
@katspaugh should be good now :)

PR #861, opened 2016-11-12
Es6 module support & plugin system

mspae:
So this PR is probably larger than PRs normally should be, but refactoring wavesurfer to work with a module system turned out to required some extra steps.

Since I realize I've been very opinionated please consider this as a base for discussion.

### TL;DR

* bundling with webpack/grunt was removed
  * es 6 transpiling with babel
  * linting with eslint
* plugin api
  * standardised plugin definition structure
  * new api for dynamic adding, initialising and destroying plugins
  * config api for automatic plugin initialisation on wavesurfer initialisation

### There are breaking changes :pensive:

* Due to a limitation with webpack the plugins on the global object are now lowercase (`window.WaveSurfer.elan`, not `window.WaveSurfer.ELAN`)
* All module systems now use the module name `WaveSurfer` (AMD as well)
* Using the old API to add plugins (e.g. `timeline.init()`) ~~doesn't work anymore~~ mostly works, but in some cases may not anymore
* I haven't managed to get the coverage reporting to work with webpack yet

### Updated build system

Previously all parts of the library (and its plugins) relied on the global WaveSurfer object to attach themselves to. Working this way with a module bundler means exposing this object which defeats the purpose of dependency injection and module bundling. The library needed to be refactored to use dependency injection so that there was no need for a global object.

Achieving this without some kind of build tool would have be a tedious task. So I opted for webpack, which provides convenient features to build multi-part libraries. Also it has an inbuilt dev server with livereloading etc. (Currently webpack 2, which is in beta, but apparently only until the docs are finished)

jscs and jshint, which were currently being used for linting have merged as eslint, so it only made sense to change to that linter (as a webpack loader).

I also updated karma to work with webpack.

So now there was really no reason to continue using grunt at all, so I removed it. Instead all the scripts are aliased from `package.json`.

###### Wavesurfer can be accessed like this:

* script-tags approach/global object:
```html
<script src="my/path/dist/wavesurfer.js"></script>
<script src="my/path/dist/plugin/wavesurfer.regions.js"></script>
<script>
    alert(window.WaveSurfer);
    alert(window.WaveSurfer.regions);
</script>
```
* commonjs/requirejs
```javascript
const WaveSurfer = require('wavesurfer.js');
const Regions = require('wavesurfer.js/dist/plugin/wavesurfer.regions.js');
```
* ES6 imports
```javascript
import Regions from 'wavesurfer.js/dist/plugin/wavesurfer.regions.js';
```

###### Development workflow changes

The new dev tasks are:

* `npm start` – Transpile the plugins and start the dev server at `localhost:8080` (reloading only works with the core lib currently)
* `npm run build`
* `npm run test`

### Plugin system

The plugins were written in a way which basically just added to the WaveSurfer global variable. To be able to implement this in a truly modular fashion there had to be some process of registering plugins, allowing them to augment the library without having to have that globally available. To achieve this I defined a standard plugin definition format. (See an example further below)

###### Adding, initialising and destroying plugins

There are three ways to add a plugin:

1. Adding and initialising plugin immediately (this will do in 99% of the cases):
```javascript
    container: '#waveform',
    waveColor: 'violet',
    plugins: [
        WaveSurfer.timeline({
            container: '#wave-timeline'
        })
    ]
});
```
2. Adding a plugin immediately and initialising it dynamically:
```javascript
    container: '#waveform',
    plugins: [
        WaveSurfer.timeline({
            container: '#wave-timeline',
            deferInit: true // stop the plugin from initialising immediately
        })
    ]
});
wavesurfer.initPlugin('timeline');
```
3. Adding and initialising a plugin dynamically:
```javascript
    container: '#waveform',
    waveColor: 'violet'
});
// adding and initialising a plugin after initialisation
wavesurfer.addPlugin(WaveSurfer.timeline({
    container: '#wave-timeline',
})).initPlugin('timeline');
```

Also a plugin can be destroyed like this:
```javascript
wavesurfer.destroyPlugin('timeline');
```

Apart from that the public plugin API has not changed.

###### Standard plugin definition format

The definition object is being returned in a function, this allows us to pass in parameters during wavesurfer initialisation. A custom built plugin could just as well be an object passed in during initialisation.

A plugin definition now looks like this:

```javascript
export default function(params) {
    return {
        // (required) plugin name, this is the namespace under which the instance of the
        // plugin instance is added to the wavesurfer instance
        name: 'myplugin',

        // defer the initialisation until later, can either be set
        // or depend on the params like this ...

        // static properties and methods are applied to the wavesurfer instance`
        static: {
            functionAttachedToWavesurfer() {
                // this refers to the wavesurfer instance
            },
        },

        // (required)
        // 1) plugin instance is instantiated and added to the wavesurfer
        //    instance under the "name" namespace
        //
        // 2) init is run:
        // - before the ready event: plugins now need to add a listener to the
        //   ready event if they need to do something after it has been triggered)
        // - after all the plugins have been instantiated, so init can for
        //   example check for the existence of this.regions
        instance: {
            // (required)
            init(wavesurfer) {
                // this refers to the instance of plugin at wavesurfer.myplugin
                this.params = params;
                this.wavesurfer = wavesurfer;

                this.methodOfThePlugin();
            },
            // (required)
            destroy() {
                // destroy this plugin
            },
            methodOfThePlugin() {
                console.log('Plugin haz been initialized!')
            }
        }
    };
}
```

cats rolling!

### Next steps

Depending on how this is received there would/could be a few next-steps
* Updating documentation
* Adding tests for the plugin system API methods
* Adding generic tests for the plugins (if they follow the format)
* Refactoring the code base to use es6 syntax (adding arrow functions, spread operator etc.)
* We could consider refactoring multicanvas drawer into a plugin (any maybe even the mediaelement backend) since they are not part of the immediate core of the library to reduce file size.


example/annotation/app.js
```diff
@@ -1,24 +1,36 @@
...
+    wavesurfer = window.WaveSurfer.create({
         container: '#waveform',
         height: 100,
         pixelRatio: 1,
         scrollParent: true,
         normalize: true,
         minimap: true,
-        backend: 'MediaElement'
+        backend: 'MediaElement',
+        plugins: [
```
thijstriemstra: why are all these plugins enabled here?

example/microphone/app.js
```diff
@@ -1,39 +1,39 @@
 'use strict';
 
 // Create an instance
-var wavesurfer = Object.create(WaveSurfer);
+var wavesurfer;
 
 // Init & load
 document.addEventListener('DOMContentLoaded', function () {
     var options = {
-        container     : '#waveform',
```
thijstriemstra: why remove this options var and use it inline below? -1

example/spectrogram/app.js
```diff
@@ -24,6 +29,7 @@ document.addEventListener('DOMContentLoaded', function () {
 
     /* Progress bar */
     (function () {
+		console.log(wavesurfer);
```
thijstriemstra: not needed is it?

mspae: fixed.

example/microphone/app.js
mspae: fixed.

src/plugin/microphone.js
```diff
@@ -0,0 +1,327 @@
...
+ * @param  {Object} params parameters use to initialise the plugin
+ * @return {Object} an object representing the plugin
+ */
+export default function(params = {}) {
+    return {
+        name: 'microphone',
+        deferInit: params && params.deferInit ? params.deferInit : false,
+        static: {
+            enableCursor() {
+                console.warn('Deprecated enableCursor! Use wavesurfer.initPlugins("microphone") instead!');
```
thijstriemstra: enableCursor? what?

thijstriemstra:
This is a huge refactor and it's too breaking as it is now. I would have to make a lot of changes in my code to simply support ES6 (which my code does not use). -1 for now. This whole refactoring of the plugins should be done later, let's first focus on the core (and try to not break anything).

src/plugin/microphone.js
mspae: damn :hushed:

thijstriemstra:
For example, you write:
> jscs and jshint, which were currently being used for linting have merged as eslint, so it only made sense to change to that linter (as a webpack loader).

Let's start with a single PR for this task for example. If we want to merge this PR I think it have to be broken into many separate PRs, otherwise it's too much to review and too easy to break stuff.

katspaugh:
This is awesome! I welcome all of the changes. @thijstriemstra is right, however, that the plugin refactoring it's a bit scary and could be done as a separate PR.

Nevertheless, I think we should dive in and work on the quality. This is a great base to start off. We could keep the 1.x version on it's own `stable` branch and merge this PR to master.

.babelrc
```diff
@@ -0,0 +1,13 @@
...
+	],
+	"plugins": [
+    	"add-module-exports"
+  	],
+  	"env": {
+    	"test": {
+      		"plugins": [ "istanbul" ]
+    	}
+  	}
+}
```
katspaugh: Wrong indentation in this file.

example/equalizer/main.js
```diff
@@ -77,7 +77,7 @@ document.addEventListener('DOMContentLoaded', function () {
             });
             input.style.display = 'inline-block';
             input.setAttribute('orient', 'vertical');
-            wavesurfer.drawer.style(input, {
+            wavesurfer.util.style(input, {
```
katspaugh: 👍

example/microphone/index.html
```diff
@@ -53,60 +53,54 @@ <h1 itemprop="name"><a href="http://wavesurfer-js.org">wavesurfer.js</a><noindex
...
-                        <li>create a <code>WaveSurfer</code> instance</li>
-                        <li>create a <code>Microphone</code> instance</li>
+                        <li>add the Microphone plugin to the plugins property of the wavesurfer options</li>
+                        <li>create a new instance of wavesurfer by using the create function</li>
                         <li>control the Microphone using the <code>start</code>, <code>stopDevice</code>, <code>play</code>, <code>pause</code>, <code>stop</code> and <code>togglePlay</code> methods</li>
                       </ol>
                     </p>
                     <p>
-                        <a class="btn btn-large btn-success" href="../../plugin/wavesurfer.microphone.js" itemprop="downloadUrl">Download <strong>the plugin</strong> (3.2 KB)</a>
+                        <a class="btn btn-large btn-success" href="../../dist/plugin/wavesurfer.microphone.js" itemprop="downloadUrl">Download <strong>the plugin</strong></a>
```
katspaugh: Should it be `/plugin/wavesurfer.microphone.js`?

mspae: fixed. https://github.com/katspaugh/wavesurfer.js/pull/861/commits/cf925eea0b1b82b4a06335fcd51029da10959994

.babelrc
mspae: fixed.  https://github.com/katspaugh/wavesurfer.js/pull/861/commits/3c5c0e3481abda042efe4c094bfebda19411a57f

mspae:
I think maintaining the 1.x branch and exchanging fixes between v2 and v1 would be a good idea. I think that is probably a good compromise.

I wanted to add a PR which as a whole is functional, that's why it turned out to be so big. I could split this PR into smaller ones which each render the bundle non-functional (but in the end result in changes identical to this PR) but I think this would not really be less overwhelming or less work (to review) than working with this monster-PR.

I honestly think it's pretty stable (and I would rather invest the time which I would spend on splitting into multiple PRs into debugging and enhancing the code), so I agree with you @katspaugh – I would rather use this PR as a starting point for further development.

thijstriemstra:
Id say put it in a v2 branch first and have many release candidates before making it master

katspaugh:
@mspae OK, cool! I've created a copy of `master` called `next`. Can you fix the conflicts and create another PR to this new `next` branch?

mspae:
yes. will do!

PR #862, opened 2016-11-13
close the audiocontext in the destroy method of the webaudio backend

mspae:
There is a maximum amount (6) for active audioContexts, so this allows on the fly enabling/disabling without reaching that limit (if the number of currently active acs is always below 6)

https://developer.mozilla.org/en/docs/Web/API/AudioContext/close

### DIFF
--- a/src/webaudio.js
+++ b/src/webaudio.js
@@ -10,6 +10,15 @@ WaveSurfer.WebAudio = {
         return !!(window.AudioContext || window.webkitAudioContext);
     },
 
+    getAudioContext: function () {
+        if (!this.ac) {
+            this.ac = new (
+                window.AudioContext || window.webkitAudioContext
+            );
+        }
+        return this.ac;
+    },
+
     getOfflineAudioContext: function (sampleRate) {
         if (!WaveSurfer.WebAudio.offlineAudioContext) {
             WaveSurfer.WebAudio.offlineAudioContext = new (
@@ -21,9 +30,7 @@ WaveSurfer.WebAudio = {
 
     init: function (params) {
         this.params = params;
-        this.ac = params.audioContext || new (
-            window.AudioContext || window.webkitAudioContext
-        );
+        this.ac = params.audioContext || this.getAudioContext();
 
         this.lastPlay = this.ac.currentTime;
         this.startPosition = 0;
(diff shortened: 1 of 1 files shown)


src/webaudio.js
```diff
@@ -257,6 +257,7 @@ WaveSurfer.WebAudio = {
         this.gainNode.disconnect();
         this.scriptNode.disconnect();
         this.analyser.disconnect();
+        this.ac.close();
```
katspaugh: If you don't want to remove the `ac` parameter now, we could track when it's passed externally (vs created internally) and close or not close based on whether we created it.

mspae: But that wouldn't fix use case #1 when an outside script is accessing WaveSurfer.WebAudio.audioContext  – Although I think that is probably a very highly exotic use case: using the audioContext, even after destroying the wavesurfer instance. – I think we could probably get away with doing what you propose without a large fallout.

mspae: Something like this: https://github.com/katspaugh/wavesurfer.js/pull/862/commits/3c9a2bab729c75488814039a99084c1815bc9029

thijstriemstra:
Why are the tests failing? Seems like this started when the audiocontext code was changed.

src/webaudio.js
katspaugh: Just to clarify, what I mean is that we need to remove the caching.

katspaugh:
Yes, the tests started failing after this change of mine: 09f25635b3897641aab010ae9d3f37dd05d9959d

I'll take a look.

mspae:
ah, I see :)

katspaugh:
Sorry, pressed the wrong button.

src/webaudio.js
```diff
@@ -10,15 +10,6 @@ WaveSurfer.WebAudio = {
         return !!(window.AudioContext || window.webkitAudioContext);
     },
 
-    getAudioContext: function () {
```
katspaugh: Please don't remove the function itself, it's used here: https://github.com/katspaugh/wavesurfer.js/blob/33e01a6be465d328d0712a7836091fd6efc6d869/plugin/wavesurfer.microphone.js

katspaugh:
Merged, thanks! I'll make a release when I fix the build.

thijstriemstra:
Please update the changelog as well.

arianrhodsandlot:
I'm running into some promlems with v1.3.0, so I'm now using some workarounds in my project. However it seems this pull request already solved my problem... Will a new version with these changes be released recently? Thanks for your hard works!

PR #863, opened 2016-11-13
Es6

mspae:
So this PR is probably larger than PRs normally should be, but refactoring wavesurfer to work with a module system turned out to required some extra steps.

Since I realize I've been very opinionated please consider this as a base for discussion.

### TL;DR

* bundling with webpack/grunt was removed
  * es 6 transpiling with babel
  * linting with eslint
* plugin api
  * standardised plugin definition structure
  * new api for dynamic adding, initialising and destroying plugins
  * config api for automatic plugin initialisation on wavesurfer initialisation

### There are breaking changes :pensive:

* Due to a limitation with webpack the plugins on the global object are now lowercase (`window.WaveSurfer.elan`, not `window.WaveSurfer.ELAN`)
* All module systems now use the module name `WaveSurfer` (AMD as well)
* Using the old API to add plugins (e.g. `timeline.init()`) ~~doesn't work anymore~~ mostly works, but in some cases may not anymore
* I haven't managed to get the coverage reporting to work with webpack yet

### Updated build system

Previously all parts of the library (and its plugins) relied on the global WaveSurfer object to attach themselves to. Working this way with a module bundler means exposing this object which defeats the purpose of dependency injection and module bundling. The library needed to be refactored to use dependency injection so that there was no need for a global object.

Achieving this without some kind of build tool would have be a tedious task. So I opted for webpack, which provides convenient features to build multi-part libraries. Also it has an inbuilt dev server with livereloading etc. (Currently webpack 2, which is in beta, but apparently only until the docs are finished)

jscs and jshint, which were currently being used for linting have merged as eslint, so it only made sense to change to that linter (as a webpack loader).

I also updated karma to work with webpack.

So now there was really no reason to continue using grunt at all, so I removed it. Instead all the scripts are aliased from `package.json`.

###### Wavesurfer can be accessed like this:

* script-tags approach/global object:
```html
<script src="my/path/dist/wavesurfer.js"></script>
<script src="my/path/dist/plugin/wavesurfer.regions.js"></script>
<script>
    alert(window.WaveSurfer);
    alert(window.WaveSurfer.regions);
</script>
```
* commonjs/requirejs
```javascript
const WaveSurfer = require('wavesurfer.js');
const Regions = require('wavesurfer.js/dist/plugin/wavesurfer.regions.js');
```
* ES6 imports
```javascript
import Regions from 'wavesurfer.js/dist/plugin/wavesurfer.regions.js';
```

###### Development workflow changes

The new dev tasks are:

* `npm start` – Transpile the plugins and start the dev server at `localhost:8080` (reloading only works with the core lib currently)
* `npm run build`
* `npm run test`

### Plugin system

The plugins were written in a way which basically just added to the WaveSurfer global variable. To be able to implement this in a truly modular fashion there had to be some process of registering plugins, allowing them to augment the library without having to have that globally available. To achieve this I defined a standard plugin definition format. (See an example further below)

###### Adding, initialising and destroying plugins

There are three ways to add a plugin:

1. Adding and initialising plugin immediately (this will do in 99% of the cases):
```javascript
    container: '#waveform',
    waveColor: 'violet',
    plugins: [
        WaveSurfer.timeline({
            container: '#wave-timeline'
        })
    ]
});
```
2. Adding a plugin immediately and initialising it dynamically:
```javascript
    container: '#waveform',
    plugins: [
        WaveSurfer.timeline({
            container: '#wave-timeline',
            deferInit: true // stop the plugin from initialising immediately
        })
    ]
});
wavesurfer.initPlugin('timeline');
```
3. Adding and initialising a plugin dynamically:
```javascript
    container: '#waveform',
    waveColor: 'violet'
});
// adding and initialising a plugin after initialisation
wavesurfer.addPlugin(WaveSurfer.timeline({
    container: '#wave-timeline',
})).initPlugin('timeline');
```

Also a plugin can be destroyed like this:
```javascript
wavesurfer.destroyPlugin('timeline');
```

Apart from that the public plugin API has not changed.

###### Standard plugin definition format

The definition object is being returned in a function, this allows us to pass in parameters during wavesurfer initialisation. A custom built plugin could just as well be an object passed in during initialisation.

A plugin definition now looks like this:

```javascript
export default function(params) {
    return {
        // (required) plugin name, this is the namespace under which the instance of the
        // plugin instance is added to the wavesurfer instance
        name: 'myplugin',

        // defer the initialisation until later, can either be set
        // or depend on the params like this ...

        // static properties and methods are applied to the wavesurfer instance`
        static: {
            functionAttachedToWavesurfer() {
                // this refers to the wavesurfer instance
            },
        },

        // (required)
        // 1) plugin instance is instantiated and added to the wavesurfer
        //    instance under the "name" namespace
        //
        // 2) init is run:
        // - before the ready event: plugins now need to add a listener to the
        //   ready event if they need to do something after it has been triggered)
        // - after all the plugins have been instantiated, so init can for
        //   example check for the existence of this.regions
        instance: {
            // (required)
            init(wavesurfer) {
                // this refers to the instance of plugin at wavesurfer.myplugin
                this.params = params;
                this.wavesurfer = wavesurfer;

                this.methodOfThePlugin();
            },
            // (required)
            destroy() {
                // destroy this plugin
            },
            methodOfThePlugin() {
                console.log('Plugin haz been initialized!')
            }
        }
    };
}
```

cats rolling!

### Next steps

Depending on how this is received there would/could be a few next-steps
* Updating documentation
* Adding tests for the plugin system API methods
* Adding generic tests for the plugins (if they follow the format)
* Refactoring the code base to use es6 syntax (adding arrow functions, spread operator etc.)
* We could consider refactoring multicanvas drawer into a plugin (any maybe even the mediaelement backend) since they are not part of the immediate core of the library to reduce file size.


thijstriemstra:
replaced description with old one.

katspaugh:
Do we want to fix the build before merging this PR?

mspae:
I'll try and fix it beforehand.

mspae:
ha!

thijstriemstra:
@katspaugh why would we not fix the build before merging this PR..?

mspae:
Can we merge into next-branch?

katspaugh:
Merged!

Issue #864, opened 2016-11-14
WaveSurfer Hover Background


dsslimshaddy:
Can anyone add or explain from where to do the hover effect like in soundcloud ?
Im not talking bout hover cursor but the whole wavecolor from 0th position to hover position. 

I want to change the waveprgress color till the part where the mouse is hovered. Any Tricks ?

katspaugh:
There's no built-in option for this, but you can hack it to change the `progressColor` parameter on hover like this:

```
wavesurfer.container.addEventListener('mouseenter', (e) => {
  wavesurfer.params.progressColor = 'red';
  wavesurfer.drawBuffer();
});

wavesurfer.container.addEventListener('mouseleave', (e) => {
  wavesurfer.params.progressColor = 'green';
  wavesurfer.drawBuffer();
});
```

katspaugh:
Ah, now I see what you mean. From 0 to the position of the mouse, independent from the playback progress. No, we don't have this functionality.

dsslimshaddy:
How about creating **two progressWave** 
1st The original one, 

Other One will be a different color and its "width" will change as we hover(to mouseposition)
and on mouseleave(width = 0)

is this possible ?

katspaugh:
Yes, it's possible.

PR #866, opened 2016-11-17
Added Hover Effect

dsslimshaddy:
Its not Proper Plugged-In . But its a Must have feature.

See `/examples/mute`
I have edited it to work for the hover effect!!


katspaugh:
Sorry, I should've been clearer in my response in #864. While the hover effect is indeed a nice feature and having another canvas is a possible solution, I don't think this should be in the core code. It belongs more to [the cursor plugin](https://github.com/katspaugh/wavesurfer.js/blob/master/plugin/wavesurfer.cursor.js), I think.

PR #870, opened 2016-11-24
Enabled eslint indent rule

mspae:


mspae:
Okay for some reason it shows all commits from the original next-PR even though I merged in next, any idea why this is? I have a few other small PRs for next lined up but it would be nicer of course if they don't all show all these commits.

katspaugh:
It's showing the commits because I squashed and rebased when I accepted your previous PR. You need to do `git fetch --all; git rebase katspaugh/next`.  Doing `git add remote katspaugh git@github.com:katspaugh/wavesurfer.js.git` first, if you haven't already.

mspae:
okay, thanks! I will do that for the other PRs before opening them.

PR #872, opened 2016-11-26
change electro link to our new project site

jspear29:
Hey I finally setup a git hub project site for our project...  It would be great if you could link to that in you projects using.

I've used jekyll like once before and I think it's cool to edit those files directly?


katspaugh:
Yep! Thanks!

PR #875, opened 2016-12-05
Only call media.load() if it exists

pedantic-git:
When running code that uses Wavesurfer under PhantomJS (even if not explicitly testing the Wavesurfer code), the JS interpreter crashes because PhantomJS DOM `<audio>` elements don't have a `load()` method. This change skips that one line if the method doesn't exist.


katspaugh:
Cool, thanks!

katspaugh:
Released in v 1.2.8.

PR #889, opened 2016-12-09
Next: enabled eslint object-shorthand rule and fixed code

mspae:


src/wavesurfer.js
```diff
@@ -573,12 +573,12 @@ const WaveSurfer = util.extend({}, util.observer, { util }, {
         );
     },
 
-    getArrayBuffer: function (url, callback) {
+    getArrayBuffer(url, callback) {
         var my = this;
 
         var ajax = WaveSurfer.util.ajax({
-            url: url,
-            responseType: 'arraybuffer'
```
thijstriemstra: The url: url was more readable imo.

thijstriemstra:
This is already using a plugin API I'm unfamiliar with, I thought this was going to be introduced in steps?

src/wavesurfer.js
mspae: I changed the eslint rule to apply to methods, property shorthands are not enforced and I reverted this code to be like it was.

mspae:
I thought the consensus was to create the next branch with all new changes and iteratively improve on the quality from there.

PR #890, opened 2016-12-10
Next: Moved html-init script to example directory and added documentation

mspae:
I don't know if moving the init script out of the library code per se is something you would consider. IMHO it's not something which should be part of the core source. I think some example code would be good whatever we do.

This PR could be valid for the master version as well, but it would probably cause breakage, so maybe not.

If we do decide to keep the auto-initialisation code in core maybe we could add some code to dynamically load wavesurfer.js (and maybe plugin code) from CDN so the init script can be the only code which needs to be added. (To keep it really simple) – What do you think?

**Edit1:** Maybe it would be a good idea to use a hyphen in the custom html tag in preparation for custom elements (which require at least one hyphen in them)

**Edit2:** Also the auto-init-code could check for `window.wavesurferOptions` global variable and use those to extend the wavesurfer defaults. (So if there are 50 wavesurfer html tags on a page and they should all have a violet waveColor we don't have to set that as attribute in each).

I'm thinking of something like this:

```html
<script>
  window.wavesurferOptions = {
    waveColor: 'violet',
    plugins: {
      timeline: {
        fontFamily: 'Times New Roman'
      },
      // ... if we want other plugin options valid for the whole page put them here
    }
  };
</script>
```

Then using the custom html element like this:
* `data-pluginName-pluginOption` overrides/sets an option for a plugin
* `data-plugins` is a comma-separated list of plugins to initialise (could also just use the plugins property in `wavesurferOptions` to determine which to use)

```html
<wave-surfer
  data-progress-color="purple"
  data-url="../media/demo.wav"
  data-plugins="timeline,myOtherImaginaryPlugin"
  data-timeline-font-size="13px">
</wave-surfer>
```
wavesurfer.min.js and any plugin code we may need is being loaded automagically by this script
```html
<script src="://my-awesome-cdn.example/wavesurfer-auto-init.min.js"></script>
```


mspae:
Broken build should be fixed by https://github.com/katspaugh/wavesurfer.js/pull/891

katspaugh:
I like the idea of fetching plugins from a CDN, but I don't think auto-init should be a separate script that bootstraps everything else. It's not much code and is quite common. For example, masonry.js does that.

Regarding renaming the tag to `wave-surfer` – again, I don't see the point of breaking some existing code on the internet with no benefit.

mspae:
Okay. Sounds like a plan. I was thinking that we could use the same API 1:1 for when custom elements are the new standard. (Assuming that web components/custom elements will be the kind of frontend tech that gets traction and is widely adopted) – Except that one small detail (that they need a hyphen in them) – But maybe we can just change that API when and if the time comes.

PR #892, opened 2016-12-11
Next: refactor cursor plugin

mspae:
* Adding comments
* Checking naming consistency
* Checking to see if stuff exists before removing it in the destroy function
* Optimising minified build size


mspae:
Broken build should be fixed by https://github.com/katspaugh/wavesurfer.js/pull/891

PR #893, opened 2016-12-13
Next: Refactor timeline plugin

mspae:
### Description
* Timeline parameters are constructed in an object literal
* variables are locally scoped where possible to reduce build size when minifying
* plugin options now extend defaults in an object literal and are cached in `this.params`.
* Intervals for the timeline labels (`timeInterval`, `primaryLabelInterval`, `secondaryLabelInterval`) can now be defined as raw values (as before) – or as a callback which receive `pxPerSec` as parameter and return the interval. – The calculated defaults are the same as before.
* Improved plugin life cycle management, things get destroyed, unhooked etc.

### Motivation
* Smaller minified builds
* Make plugin work better with the (dynamic) plugin API
* Make code more readable

/

/

### Todos/Notes
https://github.com/katspaugh/wavesurfer.js/pull/893/commits/db392aa847f9f5b80f0e098180f0b8c71641e87c https://github.com/katspaugh/wavesurfer.js/pull/893/commits/b4f8020d7f99088e98ad36f17627fd94cce71c19 and https://github.com/katspaugh/wavesurfer.js/pull/893/commits/b5ac66c2f51f40f6b43b4b85e72e586a7f9f50a1 were added to fix broken build.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -114,7 +114,12 @@ module.exports = {
         "no-bitwise": "off",
         "no-caller": "error",
         "no-catch-shadow": "error",
-        "no-confusing-arrow": "error",
+        "no-confusing-arrow": [
+            "error",
+            {
+                "allowParens": true
+            }
+        ],
         "no-continue": "error",
         "no-div-regex": "error",
         "no-duplicate-imports": "error",
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,15 @@ sudo: required
 dist: trusty
 language: node_js
 node_js:
-  - "4.2"
+  - "4"
 before_install:
   - export CHROME_BIN=chromium-browser
   - export DISPLAY=:99.0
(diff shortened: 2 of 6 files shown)


thijstriemstra:
Looks good to me, but I'll leave final approval up to @katspaugh.

mspae:
I've noticed some issues I would like to improve on this PR before it's merged:

1. `this.opts` should be `this.params` for consistency
2. I've noticed some less than ideal performance related to canvas measureText, I haven't looked into it yet but maybe there is some cheap optimisation that can be achieved by caching

mspae:
Can't reproduce the perf issue. I would merge if that's okay with you.

thijstriemstra:
Before merging any other changes, I'd like to see a green travis build for a change ;) can you add a PR that does that, or incorporate these changes in this PR, or explain why it's not possible?

PR #896, opened 2016-12-14
Next: moved backend source files into backends directory

mspae:
No idea if this is welcome. I found it made sense but it's probably a matter of taste.


thijstriemstra:
I'm fine with the current situation and don't expect any other backend to be introduced soon.

katspaugh:
I agree with @thijstriemstra. I kinda liked the flat list.

mspae:
okay :)

PR #909, opened 2017-01-02
Improve rendering performance when pxPerSec is high

elijah-taylor:
* Adds an optional 'partialRender' parameter to enable
* Calculates and renders peaks only for current visible waveform
* Keeps track of currently calculated/rendered peaks to avoid
  duplicate calculation and only incremental scroll changes are rendered

Tested all combinations of Canvas/MultiCanvas and Wave/Bars rendering
at various zoom levels.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -25,6 +25,7 @@ module.exports = function (grunt) {
                     'src/mediaelement.js',
                     'src/drawer.js',
                     'src/drawer.*.js',
+                    'src/peakcache.js',
                     'src/html-init.js'
                 ],
                 dest: 'dist/wavesurfer.js'
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -6,7 +6,7 @@ module.exports = function (config) {
             'jasmine-matchers'
         ],
         hostname: 'localhost',
-        post: 9876,
+        port: 9876,
         singleRun: true,
         autoWatch: false,
         files: [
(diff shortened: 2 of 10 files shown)


katspaugh:
Thanks a lot! Sorry it took so long to merge!

swampthang:
Awesome work! Thanks for adding this. Is this gonna make its way into the npm version? Working in nodejs and have to work using PC and Mac so need to be able to update via npm.

katspaugh:
I'm waiting for feedback in the last open PR and will release afterwards.

swampthang:
ok, thanks.

swampthang:
Question on this. Should this resolve an issue I have where the waveform disappears when the pxsPerSec value is above 260? Using zoom(270)

elijah-taylor:
(wrote this before your edit) I think the waveform disappearing with zooming is a limitation of single canvas.  Probably the internal representation of your canvas is > 4000px or whatever the size limitation of canvas is on your particular browser's/gpu's implementation.  This PR will not fix that, because the canvas allocation will still fail.  You should try multi-canvas to address this, but now this PR will mean that you won't over-render to the whole multi-canvas when the option is set.

I've used zoom levels well above 270 in my test app, I think into the thousands.

IMO multi-canvas should be the default and you should consider removing single-canvas because it's strictly a subset of the functionality of multi-canvas.

katspaugh:
I've made MultiCanvas the default renderer and published version 1.3.0 on npm. Thanks everyone!

PR #916, opened 2017-01-10
Apply correct normalize absmax calculation to drawBars()

### LINKED ISSUE #907: barWidth breaks wavform display
Adding barWidth: 3 to the Wavesurfer options removes the waveform entirely. I have created a demo at http://codepen.io/RighteousIdealizedDung/pen/BQXeod. Without the barWidth option, all appears correctly. My code is based on code from http://stackoverflow.com/questions/38918909/how-can-i-load-only-waveform-and-wait-to-user-click-play-to-download-the-audio. I am using server-generated JSON peak data. I have experimented with various combinations of options with no success. It looks like a wavesurfer bug, any suggestions as to how this can be fixed? Thanks

chrisparton1991:
Fixed #907

### DIFF
--- a/src/drawer.canvas.js
+++ b/src/drawer.canvas.js
@@ -98,7 +98,9 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.Canvas, {
 
         var absmax = 1;
         if (this.params.normalize) {
-            absmax = WaveSurfer.util.max(peaks);
+            var max = WaveSurfer.util.max(peaks);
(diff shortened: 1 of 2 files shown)


chrisparton1991:
If and when this change gets applied to the `next` branch, I think we should move the `absmax` code out into a utility method, as the same code is now used in 4 places.

katspaugh:
Thanks! 👍

PR #918, opened 2017-01-12
fix draw wrong position while playing backward seeking

candyan:
while audio is playing, i seek to backward. the cursor is still in its original position. until playing to the original position. the cursor is return to normal.

### DIFF
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -183,7 +183,7 @@ var WaveSurfer = {
         this.fireEvent('interaction', this.seekTo.bind(this, progress));
 
         var paused = this.backend.isPaused();
-        // avoid draw wrong positon while playing backward seeking
+        // avoid draw wrong position while playing backward seeking
         if (!paused) {
             this.backend.pause();
         }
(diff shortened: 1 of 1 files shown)


src/wavesurfer.js
```diff
@@ -183,16 +183,17 @@ var WaveSurfer = {
         this.fireEvent('interaction', this.seekTo.bind(this, progress));
 
         var paused = this.backend.isPaused();
+        // avoid draw wrong positon while playing backward seeking
```
thijstriemstra: "positon" should be "position".

candyan: thx, my mistake

katspaugh:
I don't quite understand the issue that is being solved here. I don't see any problem with seeking backwards on the homepage.

thijstriemstra:
@candyan can you explain this PR in more detail?

qwertycopter:
I can confirm this issue exists and candyan's edits fix it.
When playing audio and clicking back to a previous part of the waveform sometimes the cursor does not move back (although the correct audio will play)
I'm using the MediaElement backend and very long ogg files which might contribute to the problem (the browser does not download the whole ogg file but makes several 206 partial content requests)

katspaugh:
Ah, so it's only for the MediaElement. Makes sense. I'll merge then.

PR #924, opened 2017-01-20
Next: Merge from master: Improve rendering performance when pxPerSec is high (#909)

mspae:
* Adds an optional 'partialRender' parameter to enable
* Calculates and renders peaks only for current visible waveform
* Keeps track of currently calculated/rendered peaks to avoid
  duplicate calculation and only incremental scroll changes are rendered

Tested all combinations of Canvas/MultiCanvas and Wave/Bars rendering
at various zoom levels.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,12 +2,15 @@ sudo: required
 dist: trusty
 language: node_js
 node_js:
-  - "4.2"
+  - "4"
 before_install:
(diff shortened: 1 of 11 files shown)


thijstriemstra:
@mspae is it possible to fix these travis failures here as well, e.g. https://travis-ci.org/katspaugh/wavesurfer.js/builds/193778276?

PR #926, opened 2017-01-21
Next: multicanvas by default

mspae:
### Description
* `MultiCanvas` as default renderer (like in the master branch: https://github.com/katspaugh/wavesurfer.js/commit/02d99ab7ec4c576132ad6a19b3e4d227b084bbeb)
* `Canvas` deleted
* `options.renderer` is now the renderer itself, not a string identifier
* Minimap plugin was updated to work with `MultiCanvas` renderer

### Motivation
* Reduce build size
* Allow injection of custom renderers in a module context

* `options.renderer` is a renderer object, not a string

* All references directly to the `Canvas` renderer object need to be updated to point to the renderer object or instance cached in the `wavesurfer` instance, respectively `this.drawer` and `this.Drawer`

### Todos/Notes
https://github.com/katspaugh/wavesurfer.js/pull/893/commits/db392aa847f9f5b80f0e098180f0b8c71641e87c https://github.com/katspaugh/wavesurfer.js/pull/893/commits/b4f8020d7f99088e98ad36f17627fd94cce71c19 and https://github.com/katspaugh/wavesurfer.js/pull/893/commits/b5ac66c2f51f40f6b43b4b85e72e586a7f9f50a1 were added to fix broken build.


thijstriemstra:
@mspae can you add a description to every pull request so we can see where you want to go with these PRs?

thijstriemstra:
p.s, this branch has conflicts atm.

thijstriemstra:
@chrisparton1991 can you also take a look at this?

chrisparton1991:
I've taken a look and I can't see any issues. It will be great to be using ES6 going forward :+1:

mspae:
@thijstriemstra you are right i will take more Care to test if builds work (e.g. `rm -rf node_modules && npm i && npm Run build`) and i will also describe the prs in a better way and outline what my general idea is/what should be discussed.

thijstriemstra:
Thanks for the feedback @mspae.

katspaugh:
@mspae when do you think would be the best time to make this branch the default?

thijstriemstra:
> @mspae when do you think would be the best time to make this branch the default?

The answer to that question for me personally; as soon as it's compatible with my videojs plugins (that are still non-ES6) that use wavesurfer.js

katspaugh:
@thijstriemstra how would we achieve this?

thijstriemstra:
> @thijstriemstra how would we achieve this?

Continue this way by merging into the next branch until it's fully ported. Then have some release candidates that others can test against, and finally merge and do final release.

mspae:
@katspaugh I think the first release candidate should be published when we can be reasonably sure that the public-facing API won't change anymore. So as soon as possible, as late as necessary.

Just to throw in a date we can discuss: what about we aim to release the first beta release on the 4th of March.

So here are the things which are going through my head and which I would love to hear your thoughts on. Also note I will be adding more information to the PRs I've opened regarding the exact extent of the breaking changes, architectural motivation etc. within the next few days.

### Things I would like to see fixed before doing the first beta release:

1. I think the most controversial remaining change I would propose is switching from the pattern `var xyz = Object.create(XYZ); xyz.init(params);` to the `class XYZ {} var xyz = new XYZ();` pattern. I don't know what the reasons were for picking the former – The reason why I propose to change is:
    * Under the hood the `class` syntax is actually very similiar to the `Object.create` pattern. (The only real difference being that there is a constructor function instead of an init function which is a prototype method) – However it is a lot more idiomatic and will be a standardised part of the language soon and would therefore probably make it easier for new developers to reason with the code and contribute.
    * Also as addition to the above point this means some tooling for ES6 is geared towards using classes (I'm specifically thinking of esdoc (documentation generator for es6 syntax using the jsdoc convention) which works really well with new ecmascript proposed syntax and has lots of awesome features like documentation coverage badges :tada: and integration of test code into documentation) – I think with the library this size documentation management is quite an important aspect to consider
    * We're probably not at a point where performance optimisation is paramount, but even so it's a good thing to keep in mind: V8's concept of hidden classes (which is a way to optimise repeatedly called code in an interpreted language like JS) only provides performance gains for classes if their properties are only set a) before they are instantiated (as prototype or static property) b) within the constructor function. So when the browser creates a new instance by using the `Object.create` syntax and we then immediately call the `init` prototype method and set all the properties there we are actually stopping the browser from using the hidden classes caching which could be used if we use the constructor function as init function. – Note that I have absolutely no numbers to back this up and I'm just cobbling together my half-knowledge from stackoverflow – here's a link http://stackoverflow.com/questions/17925726/clearing-up-the-hidden-classes-concept-of-v8
1. Add jsdoc style documentation wherever possible and embed some doc generator (see point above)
1. Minimap overviewRegion is currently non-functional. I would like to get that running properly
1. HTML initialisation code (see https://github.com/katspaugh/wavesurfer.js/pull/890 for info on the API I would propose)

### 'Nice to have's which wouldn't necessarily mean breaking the API:
1. write test for `getPeaks`
1. Visual regression testing for multicanvas
1. If we have well documented code through jsdoc style comments it would be trivial to add either typescript - typings as external library you can reference if you want to have intelligent autocomplete when coding in/with wavesurfer - or inline type declarations (= typescript) – I think that would definitely be a step worth thinking about because wavesurfer really is quite a large codebase and using typescript greatly increases coding speed and decreases the number of preventable bugs. ("Why is that float `NaN`?")

### 'Nice to have's which would probably cause breaking changes and which I'm unsure about:
1. Refactor wavesurfer to use setters and getters where possible

katspaugh:
Awesome plan! Thanks for the heads up!

mspae:
I've added a description to the currently open PRs and fixed the builds. They were broken because of out-of-sync dependencies and a wrong node version in `.travis.yml`

thijstriemstra:
@mspae this branch has conflicts, please fix. Thanks for adding these elaborate descriptions, helps a lot!

PR #927, opened 2017-01-22
Next: plugin dev server task + updated README.md with dev tasks

mspae:
### Description

* Previously there was only a dev server with reloading for the core lib, this adds a `npm run start:plugins` task which can be used when developing plugins for an improved development experience. (To achieve this it was necessary to change the `publicPath` property in `webpack` config. Essentially now the files generated by `webpack` in memory while running the dev server are in the correct location of their built counterparts in the file system)
* `webpack` version 2 is now used (not the v2-beta) with updated dependency to appropriate `karma-webpack` version
* Duplicate example data for the ELAN plugin in the `example/plugin-system` was removed, instead the data from `example/elan` is used.
* `README.md` was updated with correct dev tasks

### Motivation
* Ease of development
* Better documentation
* Smaller repository size/Deduplication of data

/

* Paths to built code in the example directory are now relative (and correct outside of in-memory webpack compilation) – The paths have been updated, so there is practically no breakage to be expected.

### Todos/Notes
* `.travis.yml` was changed to fix broken builds: node version is now `4`, `npm prune` and `npm update` is run to get more reliable dependencies.


example/zoom/index.html
```diff
@@ -14,7 +14,7 @@
         <link rel="screenshot" itemprop="screenshot" href="https://katspaugh.github.io/wavesurfer.js/example/screenshot.png" />
 
         <!-- wavesurfer.js -->
-        <script src="/wavesurfer.js"></script>
+        <script src="../../dist/wavesurfer.js"></script>
```
thijstriemstra: what are your thoughts on these `../../dist/` changes @katspaugh?

PR #936, opened 2017-01-27
Update webaudio.js

entonbiba:
add getPlaybackRate

PR #939, opened 2017-01-27
add wavesurfer.getPlaybackRate()

entonbiba:
create getPlaybackRate method and return playback rate.

### DIFF
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -242,7 +242,7 @@ var WaveSurfer = {
      * Get the playback rate.
      */
     getPlaybackRate: function () {
-        return this.backend.playbackRate;
+        return this.backend.getPlaybackRate();
     },
 
     /**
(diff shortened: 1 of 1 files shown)


thijstriemstra:
But you just added a `getPlaybackRate` method to the backend right? Why not use that in this PR, I'm confused.

entonbiba:
@thijstriemstra should I create a new pull request with both?

thijstriemstra:
@entonbiba you can make a new pull request or add to this one.

entonbiba:
@thijstriemstra just realized the webaudio.js change was already approved and merged. #936 
So this request is updating only the wavesurfer.js file.

thijstriemstra:
@entonbiba I understand but why is this not using `return this.backend.getPlaybackRate();`

entonbiba:
@thijstriemstra you're right, fixed 👍

PR #941, opened 2017-01-28
fix getDuration() return value when using MediaElement

entonbiba:
Change getDuration function in mediaelement.js to return the correct duration length same as WebAudio.

from
        var duration = this.media.duration; // incorrect duration value returned
        console.log preview:
image

to
        var duration = this.buffer.duration; // correct duration value returned same as WebAudio
        console.log preview:
image

### DIFF
--- a/src/mediaelement.js
+++ b/src/mediaelement.js
@@ -125,7 +125,7 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
     },
 
     getDuration: function () {
-        var duration = this.buffer.duration;
+        var duration = (this.buffer || this.media).duration;
         if (duration >= Infinity) { // streaming audio
             duration = this.media.seekable.end(0);
         }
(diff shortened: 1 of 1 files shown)


src/mediaelement.js
```diff
@@ -125,7 +125,7 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
     },
 
     getDuration: function () {
-        var duration = this.media.duration;
+        var duration = this.buffer.duration;
```
katspaugh: The problem is, in MediaElement the buffer is optional. It's either
 * using Web Audio to decode a file (in which case the buffer is there)
 * simply plays the file without a waveform (no buffer).

So, maybe `var duration = (this.buffer || this.media).duration;`?

entonbiba: @katspaugh yes I was thinking the same thing, updated.

katspaugh:
Thanks! :bow:

Issue #942, opened 2017-01-29
splitchannel renders flat waveform


jamiegau:
I have been trying to get splitchannel to work without success.  I simply get X number of channels as in the AAC file rendered as a flat line (No waveform).  However, it will still play sound etc as expected...
Under firefox it complains of "The buffer passed to decodeAudioData contains an unknown content type"

I am using AAC for the audio currently.
I am also trying to show 6 channels of audio..  but it appears to do this on stereo too.

Is there a limit in what source files will work with splitchannel?  ie only mp3 stereo? Only WAV file with 6 channels??

I would appreciate comment from others who have actually gotten multi channel to work and what limitations it may have.

dmongrel:
Was just about to report this error myself.  I had everything working with one of the January 2016 builds and just switched in the latest 1.3.0.  Seems peak data is being decoded (added some logging to my wavesurfer.js), but when splitchannel is true, it isn't passing the data properly.  Same problem seems to affect the minimap.  Everything seems to work otherwise (regions, playback, etc.), but with the same result  of a flat waveform.

In addition, it appears the minimap plugin doesn't get waveform data either (I only mention this as a help in troubleshooting, not a thread hijack).  New(er) options, like partialRender, don't appear to have an effect being on or off, and switching the canvas type also seems to have no effect.  Disabling the minimap plugin doesn't have an effect either.

@jamiegau In one of the January '16 builds I've used various mp3, mono, stereo, multi-channel, and they've all worked fine (each was about 15 min. long) with the splitChannels option.

jamiegau:
Cool, great the issue is real (And it's not ME).
Any idea how long to address?  It's probably a trivial issue (I hope).  And I don't have time to look at it myself.  I am under the pump.

I also have one suggestion, change I made to the Lib for my use model.
I'll post it in another note.  It is a simple change but a good use model.

Thanks @dmongrel

mspae:
I think I saw the same Bug while doing the "next" refactor. If I remember correctly the Split Channels issue was caused by an incorrect call to `drawWave`, `drawBars` (Start and End params missing) - the minimal issue is probably related to the Switch to multicanvas renderer (minimal extends the drawer) - probably the Special drawer props are Not being Set (`maxCanvasWidth` etc)

mspae:
Minimal = minimap :p

mspae:
If no one does it before me I will open a PR this evening.

jamiegau:
[...]

PR #945, opened 2017-01-30
Fixes broken splitChannel rendering (issue #942)

mspae:
fixes broken split channel rendering caused by out-of-date call to drawWave + drawBars without the start and end parameters introduced by peakCache


katspaugh:
Thanks!

PR #946, opened 2017-01-30
Next: Html Init module

mspae:
### Description

* Adds an updated HTML Init module, with an API as outlined  previously (https://github.com/katspaugh/wavesurfer.js/pull/890) (More info about the API below)
* Other minor changes:
    * disabled warnings about large builds in dev server
    * added `npm run start:htmlinit` task
    * added an html init example

### Motivation
/

* API works as before but has more features 

/

### Todos/Notes
* More documentation about how to hook the html API initialised wavesurfer elements to custom controls
* Add URIs to CDN

### HTML API

If the html-init[.min].js file is called within a context that has access to the global window object it automatically adds wavesurfer instances to the `<wavesurfer>` elements and dynamically loads the plugin specified in the elements' attributes. An example of such an element using the APIs:

```html
<wavesurfer
    data-url="../media/demo.wav"
    data-plugins="minimap,timeline"
    data-minimap-height="30"
    data-minimap-wave-color="#ddd"
    data-minimap-progress-color="#999"
    data-timeline-font-size="13px"
    data-timeline-container="#timeline"
>
</wavesurfer>
<div id="timeline"></div>
```

* The above happens unless the flag `window.WS_StopAutoInit` is set (to true)
* The plugin also looks for an optional `window.WS_InitOptions` to use as options (see below)

In any case the file exports the `Init` class, which can be called manually:

The Init class is constructed with two parameters `WaveSurfer` and `options`, where the first is the WaveSurfer library and the second is an object of options for the instantiation. Below you can see the defaults, taken from the code:

```js
{
    // wavesurfer parameter defaults so by default the audio player is
    // usable with native media element controls
    defaults: {
        backend: 'MediaElement',
        mediaControls: true
    },
    // containers to instantiate on, can be selector string or HTMLElement
    containers: 'wavesurfer',
    // @TODO insert plugin CDN URIs
    pluginCdnTemplate: '//localhost:8080/dist/plugin/wavesurfer.[name].js',
    // loadPlugin function can be overriden to inject plugin definition
    // objects, this default function uses load-script to load a plugin
    // and pass it to a callback
    loadPlugin(name, cb) {
        const src = _params.pluginCdnTemplate.replace('[name]', name);
        loadScript(src, { async: false }, (err, plugin) => {
            if (err) {
                return console.error(`WaveSurfer plugin ${name} not found at ${src}`);
            }
            cb(window.WaveSurfer[name]);
        });
    }
}
```


thijstriemstra:
I don't use this so I can't say anything useful about this change.

katspaugh:
AFAIK, @X-Raym is using it.

X-Raym:
WaveSurfer-WP and WaveSurfer-WP Premium loads player parameter from some div data attribute on the page ([this way](http://codepen.io/X-Raym/pen/xgWQJg)), and I think it is the case for all the others WaveSurfer based WordPress audio players, like WavePlayer, and WaveSurfer Plus.
What this API have to propose exactly ?

mspae:
Yes, they all use data attributes to some extent. But they're not using the `html-init.js` that is integrated into wavesurfer. This PR is meant as a replacement for that file.

thijstriemstra:
@mspae is it necessary to improve this feature or introduce new stuff, or can we just first port the old since this hasn't been used out there yet.

X-Raym:
A codepen demo of this usage can be helpful interesting :)

mspae:
@thijstriemstra absolutely. I just wanted to give `html-init` some of that sweet lovin' since we're at it. But it's absolutely not a must. We could also just use the old code with minor modifications.

@X-Raym You can see the API in action if you checkout the PR branch, `npm install`, `npm start` and then navigate to `localhost:8080/example/html-init` – I will try and get round to adding a codepen demo within the next week but here's a quick example (with screenshots):

the half-drawn waveforms are due to a minor bug introduced in a previous PR which will be fixed in an upcoming PR. – So just imagine totally normal waveforms :smile_cat: 

```html
<wavesurfer
  data-url="../media/demo.wav"
  data-plugins="regions"
  data-regions-regions='[{"start": 1,"end": 3,"color": "hsla(400, 100%, 30%, 0.5)"}, {"start": 5,"end": 7,"color": "hsla(200, 50%, 70%, 0.4)"}]'
>
</wavesurfer>
```


```html
<wavesurfer
  data-url="../media/demo.wav"
  data-plugins="minimap,timeline"
  data-minimap-height="30"
  data-minimap-wave-color="#ddd"
  data-minimap-progress-color="#999"
  data-timeline-font-size="13px"
  data-timeline-container="#timeline"
>
</wavesurfer>
<div id="timeline"></div>
```

Issue #954, opened 2017-02-02
minimap waveform only appears after resizing the window 


piem:
hi all, and thank you for the nice work.

I have an issue with minimap, where the waveform only appear after resizing the window. This issue existed in verions before 1.3.0, so I guess it's not related to #951. I still have to find a minimal test case to show it, since the examples work as expected.

cheers, piem

entonbiba:
@piem are you calling `minimap1.render();`? 

I made it work with the following:
````
var minimap1;
wavesurfer1.on('ready', function () {
         minimap1 = null;
         var minimap1 = wavesurfer1.initMinimap({
             container: document.querySelector('#miniwavecontainer'),
             height: 30,
             waveColor: '#ddd',
             progressColor: '#999',
             cursorColor: '#68A93D'
         });
         minimap1.render();
});
````

piem:
hi @entonbiba ,

No, this doesn't seem to be enough. I applied #955, but I'm still seeing this behaviour. It seems that the render function is being called, but too early.
 
Also, the container parameter has no effect, and the minimap is always created inside the main waveform container.

best, piem

entonbiba:
@ may you provide the code you are using? The sample below works fine on multiple browsers on my end.

http://codepen.io/entonbiba/pen/ggeGyG

entonbiba:
@piem I'm closing this since it's fixed in the latest version.

Here is a working example.
http://codepen.io/entonbiba/pen/ggeGyG

piem:
hi!

yes, i confirm this is fixed in 1.3.1. Thanks for your work!

best, piem

Issue #951, opened 2017-02-01
Minimap draws flat waveform due to old calls to getPeaks and drawPeaks


dmongrel:
The minimap plugin is making calls to old versions of getPeaks and drawPeaks (without start/end values).

I am not setup to have you do a pull request, but it appears the following lines fix the issue in the render() function (original lines commented):

<code>
 render: function () {

        var len = this.getWidth();

        // var peaks = this.wavesurfer.backend.getPeaks(len);

        // this.drawPeaks(peaks, len);

        var peaks = this.wavesurfer.backend.getPeaks(len, 0, len);

        this.drawPeaks(peaks, len, 0, len);
</code>

thijstriemstra:
@dmongrel can you make a pull request? If you have a github account you can make a PR (even using the browser editor).

dmongrel:
@thijstriemstra Thanks for letting me know about this quick editing method of providing fixes.  Still haven't learned everything about GitHub yet.  I created a pull request, hopefully I did it right.

dmongrel:
Fixed in PR #959

PR #955, opened 2017-02-02
Update wavesurfer.minimap.js

entonbiba:
renders the minimap without the need to resize the window. #954 #942 #951 

Preview:
http://codepen.io/entonbiba/pen/ggeGyG

Changed 

        this.wavesurfer.on('ready', this.render.bind(this));

to 

        this.wavesurfer.on('ready', my.render.bind(this));

### DIFF
--- a/plugin/wavesurfer.minimap.js
+++ b/plugin/wavesurfer.minimap.js
@@ -94,17 +94,8 @@ WaveSurfer.Minimap = WaveSurfer.util.extend({}, WaveSurfer.Drawer, WaveSurfer.Dr
 
     bindWaveSurferEvents: function () {
         var my = this;
-        // check if parameter renderOnLoad is definied and set to true to render minimap on load
-        var miniRenderOnLoad = (this.params.miniRenderOnLoad||false);
-        if (typeof this.params.miniRenderOnLoad !== 'undefined') {
-            if (miniRenderOnLoad) {
-                this.render();
-            } else {
-                this.wavesurfer.on('ready', this.render.bind(this));
-            }
-        } else {
-            this.wavesurfer.on('ready', this.render.bind(this));
-        }
+        // render on load
+        this.render();
         this.wavesurfer.on('audioprocess', function (currentTime) {
             my.progress(my.wavesurfer.backend.getPlayedPercents());
         });
--- a/plugin/wavesurfer.minimap.js
+++ b/plugin/wavesurfer.minimap.js
@@ -99,10 +99,10 @@ WaveSurfer.Minimap = WaveSurfer.util.extend({}, WaveSurfer.Drawer, WaveSurfer.Dr
         if (typeof this.params.miniRenderOnLoad !== 'undefined') {
             if (miniRenderOnLoad) {
                 this.render();
-            } else{
+            } else {
                 this.wavesurfer.on('ready', this.render.bind(this));
             }
-        } else{
+        } else {
             this.wavesurfer.on('ready', this.render.bind(this));
         }
         this.wavesurfer.on('audioprocess', function (currentTime) {
--- a/plugin/wavesurfer.minimap.js
+++ b/plugin/wavesurfer.minimap.js
@@ -94,7 +94,7 @@ WaveSurfer.Minimap = WaveSurfer.util.extend({}, WaveSurfer.Drawer, WaveSurfer.Dr
 
     bindWaveSurferEvents: function () {
         var my = this;
-        this.wavesurfer.on('ready', my.render());
+        this.wavesurfer.on('ready', this.render());
         this.wavesurfer.on('audioprocess', function (currentTime) {
             my.progress(my.wavesurfer.backend.getPlayedPercents());
         });
--- a/plugin/wavesurfer.minimap.js
+++ b/plugin/wavesurfer.minimap.js
@@ -94,7 +94,7 @@ WaveSurfer.Minimap = WaveSurfer.util.extend({}, WaveSurfer.Drawer, WaveSurfer.Dr
 
     bindWaveSurferEvents: function () {
         var my = this;
-        this.wavesurfer.on('ready', my.render.bind(this));
+        this.wavesurfer.on('ready', my.render());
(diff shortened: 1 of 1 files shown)


plugin/wavesurfer.minimap.js
```diff
@@ -94,7 +94,7 @@ WaveSurfer.Minimap = WaveSurfer.util.extend({}, WaveSurfer.Drawer, WaveSurfer.Dr
 
     bindWaveSurferEvents: function () {
         var my = this;
-        this.wavesurfer.on('ready', this.render.bind(this));
+        this.wavesurfer.on('ready', my.render());
```
piem: This is causing errors. I guess you meant `my.render.bind(this)`?

piem: the error i get with `my.render()` is:
```
Uncaught TypeError: Cannot read property 'apply' of undefined
    at [...]/wavesurfer.js/dist/wavesurfer.js:736:15
```

katspaugh: This change is not needed at all.

entonbiba: @piem what code are you using that it gives the error? 

@katspaugh without the change it wouldn't show the minimap on on ready event and the user is required to resize the window for it to work properly or call minimap.render(); right after.

````
    var minimap1 = wavesurfer.initMinimap({
        container: document.querySelector('#minimap'),
        height: 30,
        waveColor: '#ddd',
        progressColor: '#999',
        cursorColor: '#68A93D'
    });
    //without calling this it won't render on ready
    minimap1.render();
````


In the resize event it's calling my.render(); which makes it work when resizing the window. On the ready event it's not calling it so it won't render the minimap until you resize the window or call minimap.render() yourself after wavesurfer.initMinimap().

````
    var onResize = function () {
            if (prevWidth != my.wrapper.clientWidth) {
                prevWidth = my.wrapper.clientWidth;
                my.render();
                my.progress(my.wavesurfer.backend.getPlayedPercents());
            }
        };
````

Comment (thijstriemstra) [inline plugin/wavesurfer.minimap.js]: @piem @katspaugh happy with @entonbiba's feedback? thoughts?

Comment (katspaugh) [inline plugin/wavesurfer.minimap.js]: `this.wavesurfer.on('ready', my.render());` is certainly wrong, since it's calling the render function immediately.

You need to either subscribe like the original code:
```
this.wavesurfer.on('ready', this.render.bind(this))):
``` 

Or simply call `this.render()`.

Comment (entonbiba) [inline plugin/wavesurfer.minimap.js]: @katspaugh yes this.render(); will work. I had already updated it to that in the preview code but didn't in this pull. I'll do that now.

Comment (katspaugh) [inline plugin/wavesurfer.minimap.js]: @entonbiba you are missing the point.

`this.wavesurfer.on('ready', someCallback)` – that's a subscription. The first parameter is the name of an event, the second parameter is a function.

What you are doing is calling the callback immediately, instead of passing it as a parameter:
`this.wavesurfer.on('ready', someCallback())`

This is equivalent to
```
someCallback();
this.wavesurfer.on('ready', undefined)
```

So if you think the render method should be called immediately, just call it without the on-ready call.

**Comment by entonbiba** [comment]
@katspaugh what about adding a check to see if a parameter called "miniRenderOnLoad" is set to true? It would make the rending on load optional instead of default. This way this.wavesurfer.on('ready', this.render.bind(this))): remains as default.

Something like this ok to add?

**Comment by katspaugh** [comment]
@entonbiba I don't really get why it's needed. How can it render the minimap (which is just a small copy of the waveform) before the main waveform is ready?

File: plugin/wavesurfer.minimap.js [hunk 2]
```diff
@@ -94,7 +94,17 @@ WaveSurfer.Minimap = WaveSurfer.util.extend({}, WaveSurfer.Drawer, WaveSurfer.Dr
 
     bindWaveSurferEvents: function () {
         var my = this;
-        this.wavesurfer.on('ready', this.render());
+        // check if parameter renderOnLoad is definied and set to true to render minimap on load
```
thijstriemstra: typo 'definied'

plugin/wavesurfer.minimap.js
```diff
@@ -94,7 +94,17 @@ WaveSurfer.Minimap = WaveSurfer.util.extend({}, WaveSurfer.Drawer, WaveSurfer.Dr
...
+        var miniRenderOnLoad = (this.params.miniRenderOnLoad||false);
```
thijstriemstra: spaces around `||` missing

plugin/wavesurfer.minimap.js
```diff
@@ -94,7 +94,17 @@ WaveSurfer.Minimap = WaveSurfer.util.extend({}, WaveSurfer.Drawer, WaveSurfer.Dr
...
+        if (typeof this.params.miniRenderOnLoad !== "undefined") {
+            if (miniRenderOnLoad) {
+                this.render();
+            } else{
```
thijstriemstra: space after `else`

plugin/wavesurfer.minimap.js
```diff
@@ -94,7 +94,17 @@ WaveSurfer.Minimap = WaveSurfer.util.extend({}, WaveSurfer.Drawer, WaveSurfer.Dr
...
+                this.wavesurfer.on('ready', this.render.bind(this))):
+            }
+        } else{
```
thijstriemstra: space after `else`

entonbiba:
@katspaugh It will render after the `initMinimap` is called and if the parameter miniRenderOnLoad is set to true. Which will be inside the `wavesurfer ready event`. Therefore it will render after the main waveform is rendered, fixing the onload issue with the minimap.

katspaugh:
Let's just render it right away unconditionally, maybe?

entonbiba:
@katspaugh ok updated. I'll update the example and add it to the plugins section.

Once the debounce function is added to master, I'll update the resizing event for minimap. debounce pull request #964

katspaugh:
Merged, thank you!

PR #959, opened 2017-02-03
Updated render function for drawPeaks/getPeaks

dmongrel:
Updated the render function to call the new version of the getPeaks and drawPeaks calls which now require start and end parameters.  Fixes flat waveform drawing in minimap plugin.


katspaugh:
Thanks!

PR #961, opened 2017-02-03
avoid 'for in' array enumeration

dlandau:
I'm integrating wavesurfer.js into an Ember project and have found that in this context canvas enumeration in the timeline plugin is incorrectly iterating over array properties.


mspae:
Nice one! thanks!

Issue #949, opened 2017-01-31
wavesurfer destroy method errors on firefox 31.7


chadkoepf:
we recently upgraded to wavesurfer 1.3 through npm, and it seems now that the destroy method throws an error "this.ac.close is not a function" when using this ADMITTEDLY extremely old version of firefox.

mspae:
What version of firefox is it?

chadkoepf:
31.7

chadkoepf:
I tried just emptying the div element that the wav resides in.  this fixes the issue in firefox, but then in chrome I reach the max audio context (6) message after opening multiple audio files.  I'd rather not have to check for browser and version and do different things based on that, so I'm hoping someone has an idea of what's going on with the audiocontext, and why firefox doesn't wanna close it

entonbiba:
@chadkoepf you can do something like this `wavesurfer.backend.ac.close()`. It won't work in IE, only chrome and firefox.

It's the equivalent to AudioContext.close(). The reason you're getting that error is because you're creating more than the allowed AudioContext chrome can handle.

mspae:
https://github.com/katspaugh/wavesurfer.js/pull/862 related?

chadkoepf:
The problem is with firefox.  it works fine in chrome and chrome is the only one with the audiocontext max message.  when I do wavesurfer.destroy() everything works fine in chrome, it's the old version of firefox I have to support that causes the error.  My option is to just empty the div instead of destroying the ac which makes it work fine in firefox.  I don't want to check for browser versions and write different code. 

.backend.ac.close() is undefined according to firefox, which is the same issue as .ac.close()

entonbiba:
@mspae yes,

@chadkoepf ac.close() is not available in that version of FireFox. It was available in version 40.
https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/close 
![image](https://cloud.githubusercontent.com/assets/5193884/22555354/748cc02c-e931-11e6-928b-dee395d9e9a7.png)

chadkoepf:
it looks like #862 is closing the ac in the destroy() method, which might explain why it stopped working all of a sudden.  firefox says .ac.close() is undefined.  Perhaps it's understood by newer browsers, but needs do be instantiated differently to work with older browsers?

chadkoepf:
so is there  a workaround to manually close the ac in firefox that would also be understood by modern browsers?

entonbiba:
[...]

PR #962, opened 2017-02-04
check if browser supports AudioContext.close in webaudio.js

entonbiba:
add check for browser support of AudioContext.close(). This is only supported by Chrome 43+ and FireFox 40+ for now. Edge and other non supported browsers will give an undefined error. #949 

https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/close

// check if browser supports AudioContext.close()
            if(typeof this.ac.close() !== 'undefined'){
                this.ac.close();
            }

### DIFF
--- a/src/webaudio.js
+++ b/src/webaudio.js
@@ -287,7 +287,7 @@ WaveSurfer.WebAudio = {
         // not passed in as a parameter
         if (!this.params.audioContext) {
             // check if browser supports AudioContext.close()
-            if (typeof this.ac.close() !== 'undefined'){
+            if (typeof this.ac.close() !== 'undefined') {
                 this.ac.close();
             }
         }
--- a/src/webaudio.js
+++ b/src/webaudio.js
@@ -286,7 +286,10 @@ WaveSurfer.WebAudio = {
         // close the audioContext if it was created by wavesurfer
         // not passed in as a parameter
         if (!this.params.audioContext) {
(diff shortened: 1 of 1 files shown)


src/webaudio.js
```diff
@@ -286,7 +286,10 @@ WaveSurfer.WebAudio = {
         // close the audioContext if it was created by wavesurfer
         // not passed in as a parameter
         if (!this.params.audioContext) {
-            this.ac.close();
+            // check if browser supports AudioContext.close()
+            if(typeof this.ac.close() !== 'undefined'){
```
thijstriemstra: can you add spaces around the parentheses like line 288? Weird our style checker didn't notice it, I thought it did before.

entonbiba: @thijstriemstra done

src/webaudio.js
```diff
@@ -286,7 +286,10 @@ WaveSurfer.WebAudio = {
         // close the audioContext if it was created by wavesurfer
         // not passed in as a parameter
         if (!this.params.audioContext) {
-            this.ac.close();
+            // check if browser supports AudioContext.close()
+            if (typeof this.ac.close() !== 'undefined'){
```
thijstriemstra: missed one

thijstriemstra:
Also can you not use 'Update filename.js' as title for your PRs but give it a few word title that describes the change better.

PR #963, opened 2017-02-04
Next: Fixes a few minor bugs and adds responsive option

mspae:
### Description

* Fixes the bugs that came up because of out-of-date calls to `getPeaks` and `drawPeaks` (minimap plugin and when rendering split channels) (on master https://github.com/katspaugh/wavesurfer.js/pull/945 and https://github.com/katspaugh/wavesurfer.js/pull/959)
* Fixes the incorrect `for ... in this.canvases` loops in timeline plugin and multicanvas renderer, refactored to use `forEach` where possible (on master https://github.com/katspaugh/wavesurfer.js/pull/961)
* Turn unnecessary `let` into `const`
* Fixes "only half the waveform is being drawn" bug
* Added `responsive` option (set to `true` to enable automatic rescaling when the window resizes)

### Motivation
* I created on PR for all these small changes because I was lazy, hope that's ok.
* Fix bugs …
* Responsive option was added because I think it's a very common request and the pattern used seems to appear quite often. (It is also being used in the minimap plugin)

/

/

### Todos/Notes
/

### DIFF
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -33,6 +33,7 @@ const WaveSurfer = util.extend({}, util.observer, { util }, {
         mediaType     : 'audio',
         autoCenter    : true,
         partialRender : false,
+        responsive    : false,
         plugins       : []
     },
 
@@ -232,6 +233,18 @@ const WaveSurfer = util.extend({}, util.observer, { util }, {
         this.drawer = Object.create(this.Drawer);
         this.drawer.init(this.container, this.params);
         this.fireEvent('drawer-created', this.drawer);
+        let prevWidth = 0;
+        this._onResize = () => {
+            if (prevWidth != this.drawer.wrapper.clientWidth) {
+                prevWidth = this.drawer.wrapper.clientWidth;
+                this.empty();
+                this.drawBuffer();
+            }
+        };
+
+        if (this.params.responsive) {
+            window.addEventListener('resize', this._onResize, true);
+        }
 
         this.drawer.on('redraw', () => {
             this.drawBuffer();
@@ -699,6 +712,7 @@ const WaveSurfer = util.extend({}, util.observer, { util }, {
         this.cancelAjax();
         this.clearTmpEvents();
         this.unAll();
+        window.removeEventListener('resize', this._onResize, true);
         this.backend.destroy();
         this.drawer.destroy();
         this.isDestroyed = true;
--- a/src/drawer.multicanvas.js
+++ b/src/drawer.multicanvas.js
@@ -261,7 +261,7 @@ export default util.extend({}, drawer, {
         for (j = canvasEnd - 1; j >= canvasStart; j--) {
             const peak = peaks[2 * j + 1] || 0;
             const h = Math.round(peak / absmax * halfH);
-            ctx.lineTo((i - first) * scale + this.halfPixel, halfH - h + offsetY);
+            ctx.lineTo((j - first) * scale + this.halfPixel, halfH - h + offsetY);
         }
 
         ctx.closePath();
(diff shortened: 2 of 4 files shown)


thijstriemstra:
Does it make sense to enable responsive by default? Or even better, not add an option but make it always responsive? Am I missing something?

src/wavesurfer.js
```diff
@@ -232,6 +233,18 @@ const WaveSurfer = util.extend({}, util.observer, { util }, {
...
+            if (prevWidth != this.drawer.wrapper.clientWidth) {
+                prevWidth = this.drawer.wrapper.clientWidth;
+                this.empty();
+                this.drawBuffer();
+            }
+        };
+
+        if (this.params.responsive) {
+            window.addEventListener('resize', this._onResize, true);
+        }
```
entonbiba: @mspae I would do the following instead to avoid running drawBuffer() continuously on resize event. It will hang the browser or crash it depending on how big the waveform is. 100 - 200ms should be fine, I set the timeout to 150ms.

````
let prevWidth = 0;
var resp;

this._onResize = () => {
   clearTimeout(resp)
   resp = setTimeout(function(){
      if (prevWidth != this.drawer.wrapper.clientWidth) {
                prevWidth = this.drawer.wrapper.clientWidth;
                this.empty();
                this.drawBuffer();
      }
   }, 150);
};

if (this.params.responsive) {
    window.addEventListener('resize', this._onResize, true);
}
````

thijstriemstra:
ps. i you wouldn't have added a that Responsive option in this PR it would've landed by now. I prefer if you do not add new features in PRs that port or bugfix something.

src/wavesurfer.js
thijstriemstra: What about mobile devices, they won't be able to maintain such a speed I expect. This doesn't sound like an all-round improvement across all devices, and I guess either a option or even plugin makes more sense.

entonbiba: @thijstriemstra yea, maybe we can add a `orientationchange` check and event for mobile devices? On mobile you can zoom and change the orientation vertical/horizontal.

mspae:
@thijstriemstra point taken. I will do that in future.

src/wavesurfer.js
mspae: @entonbiba Wouldn't be something like the debounce function be more appropriate (like here https://davidwalsh.name/javascript-debounce-function)

mspae:
I will close this PR and split it into
1. Minor bugfixes and ports of bugfixes
2. Responsive option. (Obviously we've not agreed on the specifics, maybe we could come to an agreement (here?) and then either I will open another PR or someone else)

src/wavesurfer.js
mspae: Maybe add `util.debounce` and use that

thijstriemstra:
awesome, thx

mspae:
So to summarise what I think is probably the common ground regarding the responsive feature:

* Expose [debounce](https://www.npmjs.com/package/debounce) at `util.debounce`(I propose to simply expose code by other people because we don't have to maintain it and I think this module is probably pretty stable :sunglasses: and will be for some time)
* Add a `responsive` option to the wavesurfer option object, the default value is false
* The value is dynamically set to true (let's call it a conditional default, even though that probably doesn't make that much syntactic sense) if `scrollParent` is not true, `pixelsPerSec` is not set and `responsive` is not explicitly set to false. (= When the waveform is 100% width and won't be scrollable)
* A `responsive` which is `true` makes wavesurfer add an event listener to the `resize` and `orientationchange` events of the `window`, the event listener is a debounced function which checks if the `clientWidth` of the drawer wrapper has changed and if so calls `this.empty` and `this.drawBuffer`

What do you think @thijstriemstra @entonbiba ?

entonbiba:
@mspae @thijstriemstra  the debounce feature has already been implemented with current latest version 1.3.1, #964 
`wavesurfer.util.debounce(FUNCTION,TIMEINMS);`

The rest of the points I agree with, the resize/orientationchange debounce feature should be set to 100ms or 150ms to make it smooth on mobile as well. I'll update the minimap plugin to include the orientationchange event as well.

PR #964, opened 2017-02-04
create debounce method in util.js

entonbiba:
Can be used with a resize event to prevent drawBuffer() from running concurrently when resizing the browser window. @mspae @thijstriemstra #963 

**preview:**
http://codepen.io/entonbiba/pen/jyKrEz

var responsiveWave = WaveSurfer.util.debounce(function() {
  wavesurfer.empty();
  wavesurfer.drawBuffer();
}, 150);

window.addEventListener(‘resize’, responsiveWave);

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -9,6 +9,29 @@ WaveSurfer.util = {
         });
         return dest;
     },
+    
+    debounce: function (func, wait, immediate) {
+        var args, context, timeout;
+        var later = function() {
+            timeout = null;
+            if (!immediate) {
+                func.apply(context, args);   
+            }
+        };
+        return function() {
+            context = this;
+            args = arguments;
+            var callNow = immediate && !timeout;
+            clearTimeout(timeout);
+            timeout = setTimeout(later, wait);
+            if(!timeout){
+                timout = setTimeout(later, wait);   
+            }
+            if (callNow) {
+                func.apply(context, args);   
+            }
+        }
+    },
 
     min: function(values) {
         var min = +Infinity;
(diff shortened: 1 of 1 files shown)


mspae:
Hey awesome! Maybe you might want to checkout the next branch (thats where my previous PR was directed) - as far as I'm concerned we can do the same for master for this feature though.

katspaugh:
I'm a bit reluctant to add code that is not used inside wavesurfer.js itself, or inside any plugins. Why can't the app authors define or import the debounce function themselves?

mspae:
The original idea came up in the context of this PR https://github.com/katspaugh/wavesurfer.js/pull/963 (responsive option)

I would like to add debounce to the next version to enable debouncing of event listeners which are called constantly (mostly `resize` and `zoom`, but maybe also `scroll` and `audioprocess` in some cases) – Off the top of my head I can think of two use cases where I would like to use debounce: the new responsive feature and the resizing of the minimap plugin. Not adding debounce would mean we can't implement the responsive feature event because it would crash or cause immense lags for large files and waveforms.

I have also though about how this could be kept seperate from the wavesurfer lib so we don't have to maintain it. But apart from providing a debounce parameter where the function can be specified I couldn't think of a good way of doing it. – and IMHO that would be even more complicated than just exposing the well-tested debounce package. (documenting, answering questions, maintaining extra code for parameter checking etc.)

Note that this is my opinion regarding the next branch/version.

katspaugh:
OK, if it's going to be used in the code, let's add it. Thanks, @entonbiba!

katspaugh:
I've published version 1.3.1 on npm.

entonbiba:
@katspaugh thanks! Yes, I'm going to update a couple of the plugins to include the debounce feature on the windows resize event. This feature also comes in handy when having more than one wavesurfer initialized on the page and in case the user keeps resizing the window.

PR #968, opened 2017-02-06
create minimap.html plugin page

entonbiba:
create minimap.html plugin page and link to it from the plugins index.html page


katspaugh:
Thanks! :thumbsup:

Issue #973, opened 2017-02-08
Bug: Browsers don't respect setting preload to "none" in wavesurfer.load()


SeanBannister:
Demo at http://codepen.io/anon/pen/QdVxbe

Preload is set to none:
`wavesurfer.load('https://wavesurfer-js.org/example/split-channels/stereo.mp3', peaks, 'none');`
But if you check in "developer tools" the file is clearly loading on page load before the file is played. The DOM shows `preload="none"` so it is adding the attrbitue. 

I understand that `preload="none"` is only a suggestion to the browser but if I manually add the <audio> element to the page it respects it:
`<audio controls preload="none" src="https://wavesurfer-js.org/example/split-channels/stereo.mp3">`

Tested under Chrome 55.0.2883.87 and Firefox 51.0.1.

If you can give me an idea what might be causing this I'll try to fix it.

@X-Raym

X-Raym:
@SeanBannister I use `preload="metadata"` and it seems to work as expected, so why not `preload="none"`... I don't have any idea.

SeanBannister:
@X-Raym I just edited my example to use `preload="metadata"` and I'm definitely getting preloading, have a look: http://codepen.io/SeanBannister/pen/GrXBEp

X-Raym:
@SeanBannister I did a some experimentation with my old codepens but I still don't have any hint about the situation. :(

entonbiba:
@SeanBannister The problem is the `media.load();` that is being called in `_load: function (media, peaks) {` section.

With preload set to none it works if we check for 'none' in the _load function, but peaks will not load and audio file will not download which is the expected result.
![image](https://cloud.githubusercontent.com/assets/5193884/22764185/0ce1f77a-ee37-11e6-8da1-e2a3d2710dad.png)



You can make it work by loading an empty 0-1 sec audio file. After the user clicks the play button; it would re-load the .mp3 file with peaks again. It is still loading the audio file to render the peaks but it's 1 second vs the full audio file.

Preview:
http://codepen.io/entonbiba/pen/xgyGdr

![image](https://cloud.githubusercontent.com/assets/5193884/22764223/463ee780-ee37-11e6-980f-b1a35acf9ee5.png)

mspae:
What's the status on this? @entonbiba @X-Raym @SeanBannister

entonbiba:
[...]

PR #974, opened 2017-02-08
Adding missing parameters to .load()

SeanBannister:
There were 2 missing parameters for .load() in the documentation: peaks and preload.

I also removed the line "Returns XHR object" as this is not the case.

This is potentially blocked by #973


katspaugh:
Thanks! :thumbsup:

PR #976, opened 2017-02-08
Documentation for the waveform-ready event.

SeanBannister:
There was no documentation for the waveform-ready event.


entonbiba:
@SeanBannister do you have an example we can look at? Not sure why we need to add 'wavesurfer-ready' as an event when it's not. By default it's WebAudio, if anything a note in the ready event would be sufficient for the MediaElement part. 

````
note: When using MediaElement, the event will fire before the waveform loads.
````

katspaugh:
That's a correct change. Thanks, @SeanBannister!

PR #978, opened 2017-02-10
getVolume method

entonbiba:
Returns the current gainNode volume.
You can now do wavesurfer.getVolume() to return the current playback volume.


katspaugh:
Useful, thanks!

PR #982, opened 2017-02-10
add getFilters method

entonbiba:
The method will return an array of all the set filters.

For example when doing something like this:
````
// setFilters
var lowpass = wavesurfer.backend.ac.createBiquadFilter();
wavesurfer.backend.setFilter(lowpass);

// getFilters
wavesurfer.getFilters();
````

### DIFF
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -295,11 +295,7 @@ var WaveSurfer = {
      * Filters must be set with setFilters method first
      */
     getFilters: function() {
-        if (typeof this.backend.filters !== 'undefined') {
-            return this.backend.filters;
-        } else{
-            return [];
-        }
+        return this.backend.filters || [];
     },
 
     toggleScroll: function () {
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -289,6 +289,15 @@ var WaveSurfer = {
         return this.isMuted;
     },
 
+    /**
+     * Get the list of current set filters as an array.
+     */
+    getFilters: function() {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/wavesurfer.js [hunk 1]
```diff
@@ -289,6 +289,17 @@ var WaveSurfer = {
         return this.isMuted;
     },
 
+    /**
+     * Get the list of current set filters as an array.
+     *
+     * Filters must be set with setFilters method first
+     */
+    getFilters: function() {
+        if (typeof this.backend.filters !== 'undefined') {
```
katspaugh: You can simply return without this check.

entonbiba: @katspaugh it will return undefined without the check since filters is not set in the wavesurfer.backend by default to empty.

image

Once it's set via `wavesurfer.backend.setFilter(lowpass)`, it contains the filters in .backend.

image

katspaugh: It will return `undefined` in your current code as well. There's no `else`-clause, so the function returns nothing, i.e. `undefined`.

entonbiba: @katspaugh I added the else statement to return an empty array.

src/wavesurfer.js
```diff
@@ -289,6 +289,19 @@ var WaveSurfer = {
...
+            return this.backend.filters;
+        } else{
+            return [];
+        }
```
katspaugh: `return this.backend.filters || [];`

entonbiba:
@katspaugh I've made it one line as suggested.
`return this.backend.filters || [];`

Issue #810, opened 2016-08-26
Spectrogram vertical frequency label


JulieMarie:
This is actually an enhancement request for the Spectrogram plugin to display a vertical axis frequency label.  I'm using the timeline plugin for the horizontal axis which works well and now I'm just missing the frequency labels.

I'm trying to figure it out on my own, but not there yet!

entonbiba:
@JulieMarie I created a sample that displays the labels for you. When I get a chance I'll add it to the official plugin.
http://codepen.io/entonbiba/pen/Bpmqrq

JulieMarie:
Thanks @entonbiba, I'll check it out!

entonbiba:
@JulieMarie I've updated the spectrogram with khz labels, the labels with custom font/color etc will be available in the next version,

JulieMarie:
Fantastic... Thank you so much!

entonbiba:
The spectrogram frequency labels parameter is now available in the latest version. #984

PR #984, opened 2017-02-11
create new labels feature for spectrogram plugin

entonbiba:
feature request #810

- create labels parameter (true|false)
- create labels element and canvas
- create labels frequency method loadLabels();
- update createWrapper() method and apply labels feature parameter.

preview:
http://codepen.io/entonbiba/pen/Bpmqrq

### DIFF
--- a/plugin/wavesurfer.spectrogram.js
+++ b/plugin/wavesurfer.spectrogram.js
@@ -97,7 +97,7 @@ WaveSurfer.Spectrogram = {
         if (this.params.labels) {
             this.drawer.style(specView, {
                 left: 0,
-                position: 'relative' 
+                position: 'relative'
             });
         }
         this.wrapper = this.container.appendChild(
--- a/plugin/wavesurfer.spectrogram.js
+++ b/plugin/wavesurfer.spectrogram.js
@@ -77,7 +77,7 @@ WaveSurfer.Spectrogram = {
         // if labels are active
         if (specLabels) {
             var specLabelsdiv = document.createElement('div');
-            specLabelsdiv.setAttribute("id", "specLabels");
+            specLabelsdiv.setAttribute('id', 'specLabels');
             specLabelsdiv.style.cssText = 'position:absolute;left:0;z-index: 9;';
             specLabelsdiv.innerHTML = '<canvas></canvas>';
             this.wrapper = this.container.appendChild(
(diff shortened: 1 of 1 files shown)


plugin/wavesurfer.spectrogram.js
```diff
@@ -73,10 +73,29 @@ WaveSurfer.Spectrogram = {
             this.container.removeChild(prevSpectrogram);
         }
 
+        var specLabels = (this.params.labels||false);
```
katspaugh: You don't need to do this kind of conversion explicitly. Just check directly `if (this.params.labels)`.

plugin/wavesurfer.spectrogram.js
```diff
@@ -73,10 +73,29 @@ WaveSurfer.Spectrogram = {
...
+        // if labels are active
+        if (specLabels) {
+            var specLabelsdiv = document.createElement('div');
+            specLabelsdiv.setAttribute('id', 'specLabels');
+            specLabelsdiv.style.cssText = 'position:absolute;left:0;z-index: 9;';
```
katspaugh: You could use this method for a prettier CSS object: https://github.com/katspaugh/wavesurfer.js/blob/master/src/drawer.js#L98-L105

plugin/wavesurfer.spectrogram.js
```diff
@@ -204,6 +223,69 @@ WaveSurfer.Spectrogram = {
         return ajax;
     },
 
+    freqType: function (freq) {
+        return (freq >= 1000 ? (freq/1000).toFixed(1) : Math.round(freq));
```
katspaugh: `freq/1000` -> `freq / 1000`

plugin/wavesurfer.spectrogram.js
```diff
@@ -204,6 +223,69 @@ WaveSurfer.Spectrogram = {
...
+        var bgFill = bgFill || 'rgba(68,68,68,0.5)';
+        var fontSizeFreq = fontSizeFreq || '12px';
+        var fontSizeUnit = fontSizeUnit || '10px';
+        var fontType = fontType || 'Helvetica';
+        var textColorFreq = textColorFreq || '#fff';
+        var textColorUnit = textColorUnit || '#fff';
+        var textAlign = textAlign || 'center';
+        var container = container || '#specLabels';
+        var getMaxY = frequenciesHeight || 512;
+        var labelIndex = 5 * (getMaxY/256);
```
katspaugh: `getMaxY / 256`

entonbiba:
@katspaugh ok updated

katspaugh:
Awesome stuff, thanks!

PR #988, opened 2017-02-12
Next: Class syntax

mspae:
### Description

* Using es6 class syntax as described here
https://github.com/katspaugh/wavesurfer.js/pull/926#issuecomment-274932307 – See some code examples below.
– In effect the main difference in the inheritance model is the
introduction of constructor functions to construct the objects.
Basically what was previously the `init` function is now refactored into:
  * the constructor function (constructing the object, assigning any
computed values to the object instance)
  * and the now-reduced init
function (actually starting whatever the object should do)
* Plugins also follow this pattern, this means that the extends property
in the plugin definition object now represents the parent class of the plugin class (Not a list of objects from which to build
a single prototype object). The instance property of the plugin
definition is now actually a factory function (which returns the plugin
class) which is called by the single superclass constructed from the
extends property.
* As part of the change to the es6 class syntax this pattern
`this.prototype.myParentMethod.call(this)` has been changed to
`super.myParentMethod()`
* Other minor changes include:
    * Minimap plugin was refactored a little more than other plugins due to the dependency on the drawer which needed a few more changes to work properly.
    * Removing `wavesurfer.PeakCache` caching of PeakCache uninitialised
object – This is probably never going to be used
    * The state objects in the webaudio class are now accessed through a
simple map (not an indexed array) and the state is set using the keys of
that map – This was done to simplify code (no performance benefit to
using indexed arrays over a map in this scenario) – A check was added to graciously fail
should code try and set an invalid state.
  * build config was changed to allow for class syntax usage
  * tests of the plugin api where updated accordingly

### Motivation
* Idiomatic use of the ES6 syntax for better tooling and shallower
learning curve for new developers
* Better performance through the use of constructor functions which
initialise properties of the class instead of dynamically adding them at
runtime

* `var wavesurfer = Object.create(WaveSurfer);` pattern is now `var
wavesurfer = new WaveSurfer(options)` – `var wavesurfer =
WaveSurfer.create(options);` has not changed.
* Plugin API has changed in regard to the extend and instance property.
  * The extend property is now a string that represents the name of the parent class (the class with which to extend the plugin). This means
that `['drawer', 'observer']` was replaced by `'drawer'` because the
drawer class already extends the observer class, no additional extending
is required to make the resulting plugin class into an observer (for example in the
case of the mimimap plugin)
  * The instance property is now a plugin class factory – it is a function which is called with either `null` or the parent class defined in `extends` property. – It returns a class which may extend the parent class.

* Initialising objects now follows the constructor function pattern. and only actually start doing something when the `init` function is called. Except in the case of the `PeakCache` class where there is no use-case for that. – This has been changed everywhere within the library, so no breakage should occur

### Todos/Notes
* ESdoc comments can now be added

### Plugin code example

More info on the plugin definition format here: https://github.com/katspaugh/wavesurfer.js/pull/863
```javascript
function myPlugin(params) {
        return {
            name: 'myPluginName',
            static: {
                myStaticMethod name: function() { /* ... */ }
            },
            extends: 'observer',
            instance: Observer => class MyPlugin extends Observer {
                constructor(wavesurfer) {
                    super();
                    this.wavesurfer = wavesurfer;
                    // do other instantiation things
                }
                init() {
                    // start doing something
                }
                destroy() {
                    // stop doing something
                }
            }
        };
    }
```


thijstriemstra:
Awesome, thanks @mspae. Could you also create a wiki page or add an UPGRADE.md file that describes these breaking changes, it would be a shame to discard all the instructions you wrote up in the description of this, and other, PRs.

mspae:
@thijstriemstra will do. I will add a section to the readme with information about upgrading I think.

PR #998, opened 2017-02-17
Fixed AudioContext.close()

meefik:
Fixed wrong check close function in AudioContext instance.


katspaugh:
I've published version 1.3.3 with this fix. Thanks a lot!

PR #1000, opened 2017-02-19
Next: Documentation

mspae:
### Description
* Added complete and consistent (hopefully) jsdoc comments to all functions (in the root library)
* Added typedefs for often used object formats (like parameter objects) and documented them
* Null initialise variables in the constructor function and documented them with jsdoc comments.
  * I have tested this and it doesn't cause any problems.
  * This is in general a good idea because the javascript engine does not need to break out of the hidden classes cache when properties are added dynamically (within a method) ([Stackoverflow](http://stackoverflow.com/questions/17925726/clearing-up-the-hidden-classes-concept-of-v8))
  * Esdoc currently does not support the `@inheritDoc` or automatically inheriting documentation of inherited properties, for this reason I null-initialised them in the child (again) – This is not very DRY, but keeps the documentation nice. (https://github.com/esdoc/esdoc/issues/148)
* Other minor changes:
  * Added additional styling for the esdoc generated documentation to make it a bit friendlier
  * Fixed an extra parameter in the call to `createDrawer` which crept in

### Motivation
/

/

/

### Todos/Notes
* Events are currently not properly supported by esdoc, We could use `@typedef` to document them better. (https://github.com/esdoc/esdoc/issues/81)
* Looking at the docs currently means running `npm run doc && npm run start` and going to `http://localhost:8080/doc` – I can think of two ways to fix this:
  * Somehow transfer the built `doc` directory in the `gh-pages` branch to deploy docs to the website
  * Using the [esdoc hosting service](https://doc.esdoc.org/) – I don't know if they support custom styles.

### DIFF
--- a/.esdoc.json
+++ b/.esdoc.json
@@ -1,9 +1,20 @@
 {
   "source": "./src",
   "destination": "./doc",
+  "title": "wavesurfer.js documentation",
+  "styles": ["example/css/doc.css"],
+  "lint": true,
+  "builtinExternal": true,
   "experimentalProposal": {
     "classProperties": true,
     "objectRestSpread": true,
     "exportExtensions": true
+  },
+  "test": {
+    "type": "mocha",
+    "source": "spec"
+  },
+  "manual": {
+    "changelog": ["CHANGES.md"]
   }
 }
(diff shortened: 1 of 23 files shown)


mspae:
Incidentally … :tada: 
no

PR #1008, opened 2017-02-24
add link to minimap plugin into navigation sidebar

tabascoeye:
I noticed the "Minimap" Plugin is present on the "Plugins" page but not in the Menu on the sidebar, so I added it.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -43,6 +43,8 @@ navigation:
         url: /plugins/timeline.html
       - title: Microphone
         url: /plugins/microphone.html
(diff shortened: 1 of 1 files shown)


katspaugh:
Thanks, good catch!

@entonbiba the embedded codepen on https://wavesurfer-js.org/plugins/minimap.html doesn't seem to display a minimap. When I open it on codepen, it works. Has to do with the external scripts?

entonbiba:
@katspaugh fixed,

Issue #1011, opened 2017-02-26
export image broken since MultiCanvas default


thijstriemstra:
Error:

```
TypeError: this.drawer.getImage is not a function
```

mspae:
Fixed by #1027

PR #1012, opened 2017-02-26
add getImage method for MultiCanvas

entonbiba:
@thijstriemstra I added the method which returns the first canvas in mutli-canvas mode. I say we can get each canvas data and put it in an array, then return. The array size might end up being too long for larger audio files. #1011

What do you think about the array solution?

### DIFF
--- a/src/drawer.multicanvas.js
+++ b/src/drawer.multicanvas.js
@@ -336,7 +336,7 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
         for (var i in this.canvases) {
             var getEntry = this.canvases[i].wave.getContext('2d');
             availableCanvas += getEntry.canvas.toDataURL(type, quality);
-        };
+        }
         return availableCanvas;
     }
 });
--- a/src/drawer.multicanvas.js
+++ b/src/drawer.multicanvas.js
@@ -331,8 +331,12 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
     },
 
     getImage: function(type, quality) {
-        // this.canvases
-        var getEntry = this.canvases[0].wave.getContext('2d');
-        return getEntry.canvas.toDataURL(type, quality);
+        // combine all available canvasses together
+        var availableCanvas = "";
+        for (var i in this.canvases) {
+            var getEntry = this.canvases[i].wave.getContext('2d');
+            availableCanvas += getEntry.canvas.toDataURL(type, quality);            
+        };
+        return availableCanvas;
     }
 });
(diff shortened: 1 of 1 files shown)


thijstriemstra:
Ha, I was about to look into this. Let me check your PR.

thijstriemstra:
I guess it makes sense to only show the first. What about stitching the canvasses together, or at least their image data, if there are more than 1 canvasses @entonbiba?

entonbiba:
@thijstriemstra yea if I add them all together in a loop it can work, we need to test the impact it has on browser with audio files 30 minutes long.

entonbiba:
@thijstriemstra I'll add the loop solution first, if it causes browser issues; we can see if utilizing web worker can work better.

thijstriemstra:
@entonbiba users can always complain about it taking too long, we can only point out that we can't control that.

thijstriemstra:
we could also return it async (some day).

entonbiba:
@thijstriemstra ok I added the loop to combine the data and return it.

mspae:
@entonbiba  @thijstriemstra  This results in a data URI like this: `data:image/png;base64,[encodedCanvasData]data:image/png;base64,[encodedCanvasData]` with waveforms with more than one canvas.

The solution is not so simple I think:
* base64 encoded data can only be joined if it is decoded before.
* Images can only be combined with a canvas element (see: http://stackoverflow.com/questions/10929067/convert-multiple-canvases-to-dataurl-in-html5) – This means we still have the canvas size limitation to deal with.

entonbiba:
@mspae yes at the moment it just returns the base64 data combined. The best we can do at the moment is return it the way it is combined or return it into as an array and add a note to the docs.

I'm was looking into using the offscreencanvas api to fix this, but it's not fully supported and only firefox does by enabling it in the settings.
https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas

thijstriemstra:
Might be worth investigating canvas.toBlob (https://developers.google.com/web/updates/2016/03/canvas-toblob-in-chrome-50) and createImageBitmap (https://developers.google.com/web/updates/2016/03/createimagebitmap-in-chrome-50).

mspae:
@entonbiba Maybe it would be a better idea to return an array of data URIs –  having the strings concatenated means they can't be decoded at all.

OffscreenCanvas allows canvas drawing in  a web worker which means it can happen asynchroneously. I haven't managed to find any information about size restrictions, but since there is nothing mentioned in the docs I would assume the size limitations of the ordinary canvas apply.

@thijstriemstra If we use OffscreenCanvas we could supply an extra `output` parameter to define what (data) format to return (blob, image, dataURI)

entonbiba:
@mspae I updated it. @thijstriemstra The canvas.toBlob looks interesting.

thijstriemstra:
Opened #1600 for `OffscreenCanvas` support.

Issue #1005, opened 2017-02-24
Provide better error messages if the maximum number of audiocontexts are reached/will be reached


CYDP0127:
Hi there, I have a question.

I am getting this error below when I am creating multiple waveforms

Uncaught DOMException: Failed to construct 'AudioContext': The number of hardware contexts provided (6) is greater than or equal to the maximum bound (6).

here's my code

```
for(var i =0; i < 10; i++){
    var wavesurfer = WaveSurfer.create({
        container: '#waveform-0',
        waveColor: 'red',
        progressColor: 'purple',
        hideScrollbar: true
    });
    wavesurfer.load('audio.mp3');
}
```

And My guess was it's because I am creating multiple instance but I found that it occurs error when I use the newest version (1.3.2) and it works without error when I use an old one(1.0.52)

If it's not a bug, can someone let me know how to create multiple waveforms in a right way ??

mspae:
Hello @CYDP0127 – The issue you're experiencing is probably due to https://github.com/katspaugh/wavesurfer.js/pull/862 (Don't reuse the audiocontext in `WaveSurfer.WebAudio.audioContext` automatically, instead use the `audioContext` provided in the options (`params.audioContext`) or create a new one (and close it when wavesurfer is destroyed)).

Maybe one of these would solve your issues:
* Cache the `audioContext` and pass as parameter to the new wavesurfer instances
* Decode the audio (on the server or wherever), cache the peak data and use the media element backend
* Use one wavesurfer instance to successively load all the audio, decode them, save the peak data and then create the other wavesurfer instances with media element and explicitly set peak data. (That way you only have one audiocontext)

entonbiba:
@CYDP0127 chrome only supports up to 6 audiocontexts, this is fine since it protects the browser from eating up all the CPU. You can try the media element solution as suggested by @mspae with pre-rendered peaks.

thijstriemstra:
Perhaps a meaningful error should be thrown when this maximum is (about to be) reached (if possible?).

katspaugh:
@mspae why did we remove the singleton audio-context? Why creating it every time?

mspae:
@CYDP0127 The limit should not be reached anymore with the new changes merged into master (see above) – This should be fixed with the next release.

Issue #994, opened 2017-02-13
Minimap plugin does not render complete waveform with large files


mspae:
In the annotation example you can see that the minimap plugin does not completely render. Apparently it is limited to the currect viewport of the main wave. (Maybe related to peak cache …)

![annot](https://cloud.githubusercontent.com/assets/1044829/22899760/41e77e1a-f22c-11e6-9d14-9b09101e545d.png)

entonbiba:
@mspae can you attach the file here?

mspae:
The online example uses 1.2.7 – try it with the current version locally.

mspae:
Maybe fixed by #1203 or #1198 – needs to be checked once #1203 is merged.

Issue #990, opened 2017-02-13
v1.3.2 Waveform display problem with timeline


janjon:
## demo
http://codepen.io/janjon/pen/jyEyMe?editors=1111

katspaugh:
I don't see any problems now. I think the issue was that cdnjs is not immediately putting up package updates.

katspaugh:
Ah, wait. It's borked indeed. The waveform is too small.

janjon:
@katspaugh Thank you for your quick response, are there any new changes to the plot?

janjon:
1.2.8  normal, Problem in 1.3.0

katspaugh:
The change in 1.3.0 was we made MultiCanvas the default. If you set `renderer: 'Canvas'`, it should work.

janjon:
@katspaugh There are problems with timeline, after switching to different files because of the use of `peak ` timeline is always back and forth after the switch can not be displayed

http://codepen.io/janjon/pen/jyEyMe

katspaugh:
Gotcha, thanks @janjon

janjon:
Thank you for your contribution

janjon:
@katspaugh `renderer: 'Canvas'` no effect

janjon:
@katspaugh How to quickly repair，I am more anxious

Thank you

thijstriemstra:
Can someone post a before/after screenshot?

janjon:
@katspaugh how to solve the problem， please

janjon:
@thijstriemstra 
http://codepen.io/janjon/pen/jyEyMe?editors=1111

After upgrading to 1.3.2 waveform can not be displayed properly, 1.2.3 can.
Waveform shows only the front part, incomplete

thijstriemstra:
@janjon what about v2.0.1?

thijstriemstra:
Please try 2.x and report create a new issue if it still exists, any older version is deprecated.

Issue #967, opened 2017-02-06
Bad zooming performance when using timeline plugin


GianlucaCal:
Hi,

I'm experiencing some problems with huge audio files (I'm talking about a 24 hours mp3s, 256kbit/s with a sample rate of 48000 Hz ).
I'm able to visualize the whole waveform with prerendered peaks generated server-side but there are several other major problems:

-the timeline is messed up and impossible to read

-as soon as I zoom-in the browser collapse and the process of zoom-in is insanely slow

-chrome can at least visualize the waveform, firefox cant and crashes

The initial situation on chrome is like this:

![cattura](https://cloud.githubusercontent.com/assets/24889920/22644526/489891b4-ec63-11e6-981f-0a07635ce48f.PNG)

My wavesurfer options are these:
```
wavesurfer.init({
        container: document.querySelector('#waveform'),
        height: 120,
        waveColor: '#A8DBA8',
        progressColor: '#3B8686',
        backend: 'MediaElement',
        normalize: true,
        renderer: 'MultiCanvas',
        dragSelection: true,
        minPxPerSec: 10,
        partialRender : true,
        pixelRatio:  1

    });
```

So I was wonder if wavesurfer can handle such big files.

Also I know I have some problem with the zoom options; even with shorter files (30 min) the zoom is particularly slow and doesnt go back to the initial waveform.
I wouldnt know where to start in order to fix these problems.

Hope someone could help me out

katspaugh:
Have you tried with the new `partialRender` parameter from #909?

GianlucaCal:
yes it's in my wavesurfer options

katspaugh:
Ah, true, sorry 😅

mspae:
@katspaugh maybe we can debounce the zoom callback for some quick perf optim

katspaugh:
Debouncing should probably be on the app side, wrapping the slider UI event.

GianlucaCal:
how can I accomplish this?

katspaugh:
Libraries like underscore.js or lodash.js have a debounce function.

Assuming your code looks like this:
```
    slider.addEventListener('input', function () {
        wavesurfer.zoom(Number(this.value));
    });
```

You need to turn it into
```
    slider.addEventListener('input', _.debounce(function () {
        wavesurfer.zoom(Number(this.value));
    }, 30)); // run not more often than once in 30 ms
```

GianlucaCal:
[...]

PR #1015, opened 2017-02-27
check if hardware context limit reached

entonbiba:
return false if audiocontext limit has been reached and console outputs the following message "This browser supports up to 6 hardware contexts".

Might need to add check for `console.log` or default to window.alert to return the message.

#1005

here is the chrome test check,
https://chromium.googlesource.com/experimental/chromium/blink/+/master/LayoutTests/webaudio/audiocontext-max-contexts-expected.txt

### DIFF
--- a/src/webaudio.js
+++ b/src/webaudio.js
@@ -14,7 +14,12 @@ WaveSurfer.WebAudio = {
         if (!this.ac) {
             // check if audiocontext limit has been reached
             if(this.ac.length >= 6){
-                console.log('This browser supports up to 6 hardware contexts');
+                var hwcText = 'This browser supports up to 6 hardware contexts';
+                if (typeof console != "undefined") {
+                    console.log(hwcText);
+                } else{
+                    window.alert(hwcText);
+                }
                 return false;
             } else{
                 this.ac = new (
--- a/src/webaudio.js
+++ b/src/webaudio.js
@@ -13,9 +13,8 @@ WaveSurfer.WebAudio = {
     getAudioContext: function () {
         if (!this.ac) {
             // check if audiocontext limit has been reached
-            if(this.ac.length >= 5){
-                var contextlimitText = 'This browser supports up to 6 hardware contexts';
-                throw new Error(contextlimitText);
+            if (this.ac.length >= 5) {
+                throw new Error('This browser supports up to 6 hardware contexts at a time');
             } else {
                 this.ac = new (
                     window.AudioContext || window.webkitAudioContext
--- a/src/webaudio.js
+++ b/src/webaudio.js
@@ -13,7 +13,7 @@ WaveSurfer.WebAudio = {
     getAudioContext: function () {
         if (!this.ac) {
             // check if audiocontext limit has been reached
-            if(this.ac.length >= 6){
+            if(this.ac.length >= 5){
                 var contextlimitText = 'This browser supports up to 6 hardware contexts';
                 throw new Error(contextlimitText);
             } else {
--- a/src/webaudio.js
+++ b/src/webaudio.js
@@ -16,7 +16,6 @@ WaveSurfer.WebAudio = {
             if(this.ac.length >= 6){
                 var contextlimitText = 'This browser supports up to 6 hardware contexts';
                 throw new Error(contextlimitText);
-                return false;
             } else {
                 this.ac = new (
                     window.AudioContext || window.webkitAudioContext
--- a/src/webaudio.js
+++ b/src/webaudio.js
@@ -15,11 +15,7 @@ WaveSurfer.WebAudio = {
             // check if audiocontext limit has been reached
             if(this.ac.length >= 6){
                 var contextlimitText = 'This browser supports up to 6 hardware contexts';
-                if (typeof console != 'undefined') {
-                    console.log(contextlimitText);
-                } else {
-                    window.alert(contextlimitText);
-                }
+                throw new Error(contextlimitText);
                 return false;
             } else {
                 this.ac = new (
--- a/src/webaudio.js
+++ b/src/webaudio.js
@@ -12,9 +12,15 @@ WaveSurfer.WebAudio = {
 
     getAudioContext: function () {
         if (!this.ac) {
-            this.ac = new (
-                window.AudioContext || window.webkitAudioContext
-            );
+            // check if audiocontext limit has been reached
+            if(this.ac.length >= 6){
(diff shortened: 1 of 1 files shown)


katspaugh:
I think the request in #1005 was that the error message should be more meaningful, not how the exception is delivered. So just catch the error and throw your own exception with an instructive message.

thijstriemstra:
Agreed.

entonbiba:
@katspaugh @thijstriemstra what do you think of that solution?

entonbiba:
@katspaugh @thijstriemstra what do you think of that solution to throw custom error message if it reaches 5 audio contexts instead of catching the error on the `this.ac = new().catch();`?

src/webaudio.js
```diff
@@ -12,9 +12,14 @@ WaveSurfer.WebAudio = {
 
     getAudioContext: function () {
         if (!this.ac) {
-            this.ac = new (
-                window.AudioContext || window.webkitAudioContext
-            );
+            // check if audiocontext limit has been reached
+            if (this.ac.length >= 5) {
```
katspaugh: I don't think we should hardcode any specific number. It probably depends on the browser. Just catch the error the browser is throwing when instantiating AudioContext.

katspaugh:
> what do you think of that solution to throw custom error message

I've asked @mspae why we allow this error to happen in the first place. If it's really justified, then the message should be something like "You've created too many Web Audio contexts. Pass a single instance in the ac parameter". Sounds like bad UX anyway.

mspae:
@katspaugh I wanted to remove the dependence on a global variable (and I might have run into other problems with `react-wavesurfer`, but I'm not sure what that was exactly)

I agree that there is probably no way round a global variable AFAIK in this case to get the proper singleton behaviour that the audiocontext API should be used with. (= there should only ever be one audiocontext ideally)

I will research some more on different aspects that need to be taken into account here (usability, "hot-rendering" of wavesurfer in for example react, server side rendering (= dependence on window variable), performance, extendability etc.) and get back to you with an update this evening (also concerning the "state of the union" for the v2 release since we're nearing the release date I envisioned in https://github.com/katspaugh/wavesurfer.js/pull/926#issuecomment-274932307) – Sorry for the delay but I was busy the last few days and want to give this the proper attention it deserves.

katspaugh:
Awesome, thanks @mspae!

mspae:
So I think we should
* reintroduce the audiocontext singleton as a global variable
* document how / add a new method / add a parameter to `wavesurfer.destroy` – to destroy the singleton for those who want that
* until we figure out how to port wavesurfer to the server (nodejs) I would recommend users disable loading of the wavesurfer library on the server. I think when projects like [web-audio-api](https://www.npmjs.com/package/web-audio-api) become more mature we could try and do that.

The v2 update of mine will follow friday. I didn't have enough time today as it turns out. :(

mspae:
So I've also put together an example to test memory profiling by using lots and lots of wavesurfer instances with different options: https://github.com/mspae/wavesurfer.js/tree/audiocontext-singleton

… I won't open a PR obviously since this is for development only. If you want to give it a try start the normal dev task and go to `http://localhost:8080/example/audiocontext/?level=0&elements=100`
* `level` parameter reflects the state that should automatically be invoked `(['create html elements', 'create instances', 'load audio', 'destroy', 'close audio context'])`
* `elements` determines the number of instances
* `interval` is the wait interval between the different levels
* wavesurfer options can also be set as url parameters

This might be adjusted for memory profiling for other aspects of the library/plugins.

mspae:
And here is my general v2 update:

### Things that have been done:
* HTML init script
* Added (more) complete jsdoc comments for the main library and customised esdoc for our needs. I would propose we use this to replace the documentation on the website (and provide it there) because it is probably considerably easier to maintain than a seperate documentation.
* ES6 class syntax has been implemented

### One PR to go ...
There is one architectural PR I would like to merge before we release the first beta IMHO (I'm currently working on it) – The issues the PR addresses are all to do with the plugin API:
* The plugin definition format includes two (!) properties that are actually reserved words in the JS language :rofl: (`extends` and `static`) – although this is not a problem within the current spec, old browsers might have problems with that ([stackoverflow](http://stackoverflow.com/questions/7022397/using-reserved-words-as-property-names-revisited))
* It would be nice to use esdoc generated documentation in plugins, esdoc only supports documentation for classes which are actually exported, the current plugin factory approach doesn't work.
* I feel the plugin inheritance mechanism is the unstable-est part of the lib right now and want to fix that. – Also it's basically used to make things into and observer AND in the minimap plugin it is used to inherit from the `Drawer`

… So the upcoming PR will introduce the following changes:
* All plugins automatically inherit from the observer
* Plugin files export the plugin definition factory and the plugin class. This has the `@inherits {Observer}` jsdoc tag so the documentation generator gets things right.

I think this is a good middle ground between a simple API which provides for 99% of the use cases and still allows pretty advanced features.

## 'Nice to haves' which can happen after first beta release

... in the order of their approximate order of priority IMO

* ... Merging all the small changes from master of course
* More documentation/examples/advanced use cases and linking this to the website
* Responsive option
* Doing more memory profiling and doing perf optimisation
* To mark the v2 change I would like to update and harmonise, refresh the general look and feel of the documentation + website + examples and provide more consistent branding. :ribbon:
* Adding visual regression testing (we can use media element backend and hard coded peak data to have this available even on travis/headless browser AFAIK) – but I haven't looked into this.
* Maybe switching to typescript – My IDE (visual studio code) does provide autocompletion and typechecking from the jsdoc comments, which I didn't anticipate before but which is pretty nice. – So this is a low-priority for me.

## Bugs which apply for master and next AFAIK

... I'm just including this here for a more complete roadmap

* Minimap plugin does not render complete waveform with large files https://github.com/katspaugh/wavesurfer.js/issues/994
* export image broken since MultiCanvas default https://github.com/katspaugh/wavesurfer.js/issues/1011
* Waveform display problem with timeline https://github.com/katspaugh/wavesurfer.js/issues/990
* Bad zooming performance when using timeline plugin https://github.com/katspaugh/wavesurfer.js/issues/967
* empty + redrawBuffer breaks player with pre-rendered peaks https://github.com/katspaugh/wavesurfer.js/issues/943
* partialRender with zoom = broken https://github.com/katspaugh/wavesurfer.js/issues/932

mspae:
@entonbiba I think this can be closed – The limit will not be reached when https://github.com/katspaugh/wavesurfer.js/issues/1021 is merged.

PR #1017, opened 2017-02-27
create playlist parser for m3u, pls, smil and json playlists

entonbiba:
#1016


thijstriemstra:
I don't really need this functionality to be part of 'core' and rather see it in a plugin. what you @katspaugh?

katspaugh:
Definitely.

entonbiba:
ok I'll create it as a plugin

PR #1018, opened 2017-02-28
create playlist plugin

entonbiba:
### create playlist plugin
  - user util.ajax to get file content - **ok**
  - support m3u - **ok**
  - check if mp3 or wav file urls set in playlist - **ok**
  - return empty array if playlist type not supported or set - **ok**
  - create plugin example and docs - **ok** #1019 

### Parameters:
**playlistFile** - link to playlist file

**playlistType**
'm3u' or 'audio/mpegurl' will default to m3u

Preview:
http://codepen.io/entonbiba/pen/OpMQjR

### DIFF
--- a/plugin/wavesurfer.playlistparser.js
+++ b/plugin/wavesurfer.playlistparser.js
@@ -24,7 +24,7 @@ WaveSurfer.Playlist = {
 
             var _this = this;
             ajaxData.on('success', function (data, e) {
-                _this.parse(data);
+                _this.loadPlaylist(data);
                 _this.wavesurfer.fireEvent('playlist-ready');
             });
             ajaxData.on('error', function (e) {
@@ -36,18 +36,18 @@ WaveSurfer.Playlist = {
         }
     },
 
-    parsedList: function() {
+    getPlaylist: function() {
         return this.playlistData;
     },
 
-    parse: function(playlistFile) {
+    loadPlaylist: function(playlistFile) {
         // check if playlist type is given
         var playlist = [];
 
         if (this.playlistType == 'm3u' || this.playlistType == 'audio/mpegurl') {
             playlist = playlistFile.replace(/^.*#.*$|#EXTM3U|#EXTINF:/mg, '').split('\n');
         } else {
-            throw new Error('No valid playlist file provided, valid formats are m3u pls smil json or their valid mime types');
+            throw new Error('No valid playlist file provided');
         }
 
         // playlist type is set return the playlist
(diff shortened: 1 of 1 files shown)


entonbiba:
@thijstriemstra yes don't merge until I complete the remaining tasks, I added the label 'needs-update'.

thijstriemstra:
@entonbiba i'm not sure if thats the way to go. maybe you should not make a PR before that checklist is completed?

mspae:
I don't have any strong opinions about this. I think we all agree the PR should only be merged when it's ready. But whether PRs should only be opened when they are truly ready (or the author considers them to be ready at least) I don't know – It does add clutter to the PR list and may be confusing if everybody starts doing it.

katspaugh:
I think it's OK to open a PR to discuss the approach before it's too late. Also, is `m3u` the most popular playlist format? If so, this plugin would already be useful to people even without other formats.

entonbiba:
@mspae @thijstriemstra @katspaugh yea I opened a pr so I can share the code updates directly here instead of linking to them.

I'll remove the other formats for now and create the example for m3u.

entonbiba:
@thijstriemstra @katspaugh the playlist parser is now ready 👍

entonbiba:
@thijstriemstra sure, all agree on this @katspaugh @mspae ? `PlayList`

mspae:
@entonbiba I like that name.

plugin/wavesurfer.playlistparser.js
```diff
@@ -0,0 +1,68 @@
...
+        } else {
+            throw new Error('No playlist file provided');
+        }
+    },
+
+    parsedList: function() {
+        return this.playlistData;
+    },
+
+    parse: function(playlistFile) {
```
thijstriemstra: parse sounds too techy, I'd personally call it load. And I also prefer `playlistData` over `parsedList`, can you come up with a better name that avoids `parse`?

plugin/wavesurfer.playlistparser.js
```diff
@@ -0,0 +1,68 @@
...
+        // check if playlist type is given
+        var playlist = [];
+
+        if (this.playlistType == 'm3u' || this.playlistType == 'audio/mpegurl') {
+            playlist = playlistFile.replace(/^.*#.*$|#EXTM3U|#EXTINF:/mg, '').split('\n');
+        } else {
+            throw new Error('No valid playlist file provided, valid formats are m3u pls smil json or their valid mime types');
```
thijstriemstra: I would remove the listing of the supported formats in the error and just keep it short: No valid playlist file provided.

entonbiba:
@thijstriemstra updated, I changed it to getPlaylist and loadPlaylist

Issue #944, opened 2017-01-29
Suggested Improvement, Waveform magnification value.


jamiegau:
I made this change myself (It was trivial, but suggest it also be added.)
You can turn on the normalize flag so the waveform-height fills the window.
However, this is based on the amplitude of the audio you are pointing at.

I suggest you also add the ability to set a magnification constant as well.
Ie.  I am showing a few different waveforms on the screen at once.  But I want them to be relative to each other.  However, they are quite low and do not produce a very detailed waveform.
So I added the ability to set the magnification (0 to 1) so I could for example.
Double all waveform height by setting to 0.5
This could be a percentage or whatever based on how its surfaced...

The ability to be able to set just how much the waveform is magnified, in general, is a good idea.

I did this myself by looking at the normalize parameter and simply added my own.

mspae:
Can you post your Code? It's a bit easier to talk about. Thanks!

jamiegau:
Ok,

firstly I added the property
waveHeightScale, default 1...

```
var WaveSurfer = {
    defaultParams: {
        height        : 128,
        waveColor     : '#999',
        progressColor : '#555',
        cursorColor   : '#333',
        cursorWidth   : 1,
        skipLength    : 2,
        minPxPerSec   : 20,
        pixelRatio    : window.devicePixelRatio || screen.deviceXDPI / screen.logicalXDPI,
        fillParent    : true,
        scrollParent  : false,
        hideScrollbar : false,
        normalize     : false,
        waveHeightScale : 1,
        audioContext  : null,
        container     : null,
        dragSelection : true,
        loopSelection : true,
        audioRate     : 1,
        interact      : true,
        splitChannels : false,
        mediaContainer: null,
        mediaControls : false,
        renderer      : 'MultiCanvas',
        backend       : 'WebAudio',
        mediaType     : 'audio',
        autoCenter    : true,
        partialRender : false
    },
```
Then in the code base where it deals with the normalize parameter and inits the absmax variable default as 1, I change it to init it to the waveHeightScale parameter,  about 3-4 spots.. I have replaced the following line....
```
        // var absmax = 1;
        var absmax = this.params.waveHeightScale;
```

About 5 lines of changed code..  But I hope that makes it clear now..
As I can set the parameter waveHeightScale to set a specific SCALE/Magnification.

Thanks,
James

Issue #915, opened 2017-01-10
Waveform microphone plugin is not mobile responsive


ToJen:
Looks perfect on desktop but the wave becomes too thin.

I'm only using the microphone so `waveform.zoom` may not be a solution.

This is what I think might be help but it remains the same:

`
wavesurfer.on('ready', function() {
    	wavesurfer.zoom(100);
    	console.log("zoomed");
    });
`

Here's a [screenshot](https://drive.google.com/open?id=0B584NUUwKdOHN1hwTjE4YTRVSXc) of what it looks like right now.

You can also run the demo plugin [here](https://wavesurfer-js.org/plugins/microphone.html) on mobile to see what I mean.

chrisparton1991:
Hi @ToJen, can you please provide a screenshot? There isn't enough information in your question for me to know what might be the issue.

Try going to https://wavesurfer-js.org/ on your mobile device, does the waveform look okay for you there? If so, the issue is likely with your HTML file not having the correct viewport settings (see https://developer.mozilla.org/en/docs/Mozilla/Mobile/Viewport_meta_tag).

ToJen:
@chrisparton1991 I've just updated the issue with a link to the image.

ToJen:
@katspaugh is there a way I can use `zoom`  by default?

thijstriemstra:
@ToJen, can you provide a better description? waveform is not mobile responsive can mean many things. Cheers.

ToJen:
@thijstriemstra hi, i meant to say that while I using the `microphone.js` plugin on a mobile device, the waves generated are really thin and appear to be just a line.

I attached a screenshot to the original question above ^^^

thijstriemstra:
@ToJen what browser? what platform? the waveform looks like room ambient noise, what kind of audio  are you providing in that example?

ToJen:
@thijstriemstra on firefox, chrome, safari for android and iOS. I've tested with loud music right in front of the microphone and it only comes up as tiny spikes.

ToJen:
But the thing is, if you launch mobile view on chrome(desktop), it'll come out looking fine. Try [this](https://wavesurfer-js.org/plugins/microphone.html) and see what I mean.

thijstriemstra:
@ToJen what device?

ToJen:
@thijstriemstra do you mean the device i'm testing it on? iPhone 5 and 6, LG l90, samsung galaxy, google nexus, different types

[...]

PR #1020, opened 2017-03-02
add bar height parameter

entonbiba:
add barHeight paramter, I'll update the docs once this gets merged.
#915 #944

example: http://codepen.io/entonbiba/pen/VpedWZ


mspae:
Awesome!

Currently a higher value produces a larger waveform and a lower value a smaller one. Maybe we could change the name of the parameter, otherwise it will be confusing. (Something like `shrinkage` :grinning:)

entonbiba:
ok I made it accept high numbers :P for example `100` makes it huge, while 0.5 smaller.

src/drawer.multicanvas.js
```diff
@@ -220,7 +220,7 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
         var offsetY = height * channelIndex || 0;
         var halfH = height / 2;
 
-        var absmax = 1;
+        var absmax = 1/this.params.barHeight;
```
thijstriemstra: spaces around operator

entonbiba: @thijstriemstra updated

Issue #374, opened 2015-03-01
Unstable after attempting to load consecitve long mp3s.


tomByrer:
When loading a large mp3s (40 minute lecture, 1 hour DJ mix), http://www.wavesurfer.fm/?fill works fine the first time.  Load another afterwards, wavesurfer crashes or won't load.  Chrome & Fx Win7.  Reloading the webpage resets it, so it will work for the first long MP3.

Memory/array issue?

katspaugh:
I guess it's a duplicate of #320. Web Audio issue, no known workaround.

tomByrer:
> Web Audio issue

I don't think so; I can open same files consecutively in same browsers using native players with perfect playback.
Reopen please, or should I see first if other players have same issue first?  (Though other issue may help, I'm very unsure.)

katspaugh:
You might assume I'm indignant in closing useful issues, but unless you do some memory profiling and prove that the memory is hold up in JavaScript I'm not inclined to change my view. I'm reopening and hope you'll find something interesting out. Please, make sure you tell Web Audio from "HTML5" audio.

tomByrer:
Sorry, I didn't mean to offend.

[paullewis/music-dna](https://github.com/paullewis/music-dna) will play all my test files in Chrome.  The hour-long ones will hang in Fx, but will play the 40 minute MP3s.

katspaugh:
No offence taken :smile: 

Thanks for the link! It seems they use Web Audio's `decodeAudioData` (the same as wavesurfer.js). I'll investigate what can we learn from them.

tomByrer:
I should note in the other program, Chrome didn't display any waveform, & fFx barely did.
A few thoughts:
* check how HTML5 game writers handle large arrays & files; they tend to push limits
* perhaps better just to stop & warn when taking too long?
* would one of those fancy new webworkers help?

entonbiba:
@tomByrer may you try with the latest version? Many improvements have been made.

mspae:
@entonbiba I'm not sure if the changes that were merged recently have fixed this. I think the issue is probably caused by some memory not being collected when loading new audio files, not by the debounce/zoom type of performance issues. I think there is probably no way to fix this without doing timeline analysis to try and find the issue(s)

entonbiba:
@mspae yes but we did add the `audiocontext.close()` on the destroy method in the latest version which should fix the memory issues for chrome and firefox as these are the only browsers currently supported.

[...]

PR #1021, opened 2017-03-02
Readded global audio singleton, added closeAudioContext parameter

mspae:

* audio context is now reused and not destroyed everytime. This allows more than 6 waves to exist.
* audio context is destroyed in the `destroy` method if `params.closeAudioContext` is set to true

* None, unless usage relied on the closing of the wavesurfer AudioContext, this needs to be done explicitly (see above)

/

### Todos/Notes:
* For v2/next I would use `window.WaveSurferAudioContext` since the library object isn't necessarily a global variable.

* https://github.com/katspaugh/wavesurfer.js/issues/1005
* https://github.com/katspaugh/wavesurfer.js/pull/1015
* https://github.com/katspaugh/wavesurfer.js/pull/998
* https://github.com/katspaugh/wavesurfer.js/pull/962
* https://github.com/katspaugh/wavesurfer.js/issues/374

### DIFF
--- a/src/webaudio.js
+++ b/src/webaudio.js
@@ -11,12 +11,12 @@ WaveSurfer.WebAudio = {
     },
 
     getAudioContext: function () {
-        if (!window.WaveSurfer.WebAudio.audioContext) {
-            window.WaveSurfer.WebAudio.audioContext = new (
+        if (!WaveSurfer.WebAudio.audioContext) {
+            WaveSurfer.WebAudio.audioContext = new (
                 window.AudioContext || window.webkitAudioContext
             );
         }
-        return window.WaveSurfer.WebAudio.audioContext;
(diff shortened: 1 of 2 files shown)


mspae:
Just noticed the context destruction doesn't work for `params.audioContext` + `params.closeAudioContext`.

src/webaudio.js
```diff
@@ -11,12 +11,12 @@ WaveSurfer.WebAudio = {
     },
 
     getAudioContext: function () {
-        if (!this.ac) {
-            this.ac = new (
+        if (!window.WaveSurfer.WebAudio.audioContext) {
```
katspaugh: You don't need `window` here, since `WaveSurfer.WebAudio` is defined in the same file. It should not necessarily be a global object.

Issue #736, opened 2016-06-02
ready event fires before waveform appears


tctc91:
In the docs, it says that the 'ready' event should fire when "audio is loaded, decoded and the waveform drawn.". However, for me it is firing as soon as the audio starts playing (BEFORE the waveform is drawn).

Edit-

I see this is a duplicate of #429. Could we have a separate event fired for the waveform? From what I understand, the solution in #429 won't work for me as I need the track to start playing before the waveform is loaded.

katspaugh:
Do you have a proposal for the new event name?

tctc91:
Nothing nice and simple like the other event names! 

waveformReady / waveformready ?

belackriv:
this seems to work in the meantime:
```js
wavesurfer.once('redraw', ()=>{
      debugger;
    });
```

KainLaVey:
I really need an Event that is fired when the waveform is drawn, too. :(

tctc91:
Any update on this issue :) ?

KainLaVey:
You can use the loading event and wait until it has load the track 100%. 

`wavesurfer.on('loading', function(e){ if(e == 100){ //doSomething; } });`

Its nearly the same as to wait the waveform drawn, because it will be drawn immediately after the track is loaded 100%.

X-Raym:
The mentioned didn't work for me :/ too bad; it would really help me build my peak cache system.

X-Raym:
I found that having an event which fires when waveform is decoded from audio loaded with media element backend is very useful to create peaks file. No other native events works as nicely, so I think it could be good to have it.
https://github.com/X-Raym/wavesurfer.js/commit/250cde1410622b68d219f23ec2a3571cf1e30e26

I can make a pull request if you are interested :)

X-Raym:
@katspaugh Hi,
Are you ok for a pull request with the event mentionned above ?
I use it for WaveSurfer-WP-Premium, and it is especially useful for peaks file creation, as this event is only trigger if wavesurfer is loaded with MediaElement and without peaks.

katspaugh:
@X-Raym I'm all for it, but I don't get why you named it "noPeaks". In the`decodeArrayBuffer`, callback peaks are, on the opposite – present.

X-Raym:
@katspaugh 'noPeaks' because it is trigger only if no peaks have be passed as parameter of the `load` function. This event is called when waveform is drawn from audio analysis, and not from other sources (like a peaks file).

[...]

PR #1023, opened 2017-03-03
Next: Updated plugin definition format and class exports

mspae:
* For the reasons described here: https://github.com/katspaugh/wavesurfer.js/pull/1015#issuecomment-283609182 the plugin definition format was updated.
  * All plugins automatically inherit from the observer
  * Plugin files export the plugin definition factory and the plugin class. This has the `@inherits {Observer}` jsdoc tag so the documentation generator gets things right.
* Other minor changes:
  * Fixed wrong source map config and disabled hot module reloading
  * Plugin files now export the factory function (the function which returns the plugin definition) as `default`, the plugin classes are exported as `PluginNameClass` – This results in the following code
  * Fixed an issue related to https://github.com/katspaugh/wavesurfer.js/issues/736 (minimap does not render on initial load with backend `MediaElement`) by checking which backend is used and using the correct event to hook into. (`ready` or `waveform-ready`)
  * Updated jsdoc comments and added a correct `PluginClass` interface class (is removed in minified build by build system)

* The plugin factory function is now always the `default` export, this means code needs to be adjusted (see Code example)

* See above, but should be fixed everywhere

### Todos/Notes:
* Esdoc now parses the plugin classes, this has caused the documentation coverage to decrease to `52.94% (171/323)`
* Upgrade guide needs to be updated
* Example HTML needs to be updated

### Code example

```javascript
// common js
const pluginFactory = require('wavesurfer.js/dist/plugin/cursor.wavesurfer.js').default;

//  es6
import pluginFactory from 'wavesurfer.js/dist/plugin/cursor.wavesurfer.js';

// oldschool, (WaveSurfer is a global variable)
const pluginFactory = WaveSurfer.cursor.default;

  plugins: [
    pluginFactory(pluginParams);
  ]
});
```

### DIFF
--- a/example/annotation/app.js
+++ b/example/annotation/app.js
@@ -17,14 +17,14 @@ document.addEventListener('DOMContentLoaded', function () {
         minimap: true,
         backend: 'MediaElement',
         plugins: [
-            window.WaveSurfer.regions(),
-            window.WaveSurfer.minimap({
+            window.WaveSurfer.regions.default(),
+            window.WaveSurfer.minimap.default({
                 height: 30,
                 waveColor: '#ddd',
                 progressColor: '#999',
                 cursorColor: '#999'
             }),
-            window.WaveSurfer.timeline({
+            window.WaveSurfer.timeline.default({
                 container: "#wave-timeline"
             })
         ]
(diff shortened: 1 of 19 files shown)


mspae:
Actually I want to try if I can produce a cleaner API.

mspae:
Reopened as https://github.com/katspaugh/wavesurfer.js/pull/1025.

PR #1024, opened 2017-03-03
Parameter to render the sum of absolute negative and positive value of peak in drawBars

ale-grosselle:
As i written in title, loop present into drawBars sometimes use negative value.
Is not correct.
And then each bar is calculated adding negative value + positive value


mspae:
Just a quick questions, because I couldn't easily see what the difference visuallly would be: This causes the height of the bars to be more regular, right? (I've notices drawBars sometimes rendering very short bars)

ale-grosselle:
@entonbiba I have update my code

ale-grosselle:
@mspae Yes.
As you can see in the pic below, first wave is new waveStyle, the second is the default.

image

mspae:
Ah yes, nice one! Thanks for the screenshot!

So since I think there will probably not be more wave styles could we maybe just add a boolean parameter `normalizeBars` or similiar which is set to `false` as default, @entonbiba.

Maybe we should wait with feedback from the others (@katspaugh and @thijstriemstra) before changing the param again – otherwise you might need to change it again @ale-grosselle.

entonbiba:
@mspae yea, I thought it actually made it look like soundcloud wave with the bottom half being shorter like below. If we don't add it as a new wavestyle parameter, we might not need to merge it since barHeight parameter will be available.

image

mspae:
@entonbiba IMHO that kind of custom rendering should be in a custom renderer. It's not something that should be supported by default since it involves a lof of custom code just to get a wave/bar form that looks like that:
1. bottom half of form has a different `fillStyle` – This means you would need to draw two distinct shapes with different fill style for both progress and the normal waveform
2. The `barHeight` or "`normalizeBars`" parameter used in the calculation of the bottom half is different from the top half.

What this PR we're currently discussing does is normalize the height of the bars so they look more like their wave counterparts. This is something which should be part of the core library.

I think in future we could introduce rendering functions which can be overridden in the parameters (Like I have done in the timeline plugin here: https://github.com/katspaugh/wavesurfer.js/pull/893) so people can customise their waveform without having to inherit/reimplement the whole renderer object. I think we need to seriously think about the specifics of the API before opening any PRs on that issue. Maybe it would also be wise to add that kind of feature to the v2 release branch, otherwise the branches will diverge a lot more and merging will become a pain in the ****.

katspaugh:
I think you can make this the default without a new parameter, it looks better.

entonbiba:
@mspae @katspaugh sounds good, I'm thinking of adding the following option to the master via a parameter, it will essentially be a transparent div. That way we don't have to re-render all the time. This would essentially be able to create the "soundcloud" wave effect.

Something like this, what you think?
image

image

mspae:
@entonbiba please read my comment above https://github.com/katspaugh/wavesurfer.js/pull/1024#issuecomment-284233026

entonbiba:
@mspae sure but this will only be a transparent div in around 5-10 lines of code and will only show if the parameter is active. It won't re-render anything it will be just inside the <wave> elements.

thijstriemstra:
I think he's saying, no more new parameters till v2-ish.

mspae:
@entonbiba @thijstriemstra I was kind of saying both: "It's not going to be simple so we shouldn't necessarily have it in core" and "maybe this totally new feature would be a good point to start working more on v2"

I think at first glance it could be done in 5-10 lines of code – BUT consider all use cases (even "just" the soundcloud style wave is far more complex than is immediately apparent)

I think there are two ways to go about this:
1. **Quick:** Write a custom renderer for a speciific use case (essentially customise multicanvas with custom rendering logic) and tell people to do the same if they want to have "soundcloud" style (https://github.com/katspaugh/wavesurfer.js/pull/1024#issuecomment-284227312), or "half-transparent waveform" (https://github.com/katspaugh/wavesurfer.js/pull/1024#issuecomment-284508903) style or "bottom half is red and bars, top half is blue and waveform" style or whatever :lollipop: :stuck_out_tongue:
2. **Correct:** Write a kickass API which is well documented, doesn't cause performance regressions and is well-architected in the sense that it is flexible, intuitive and extensible. – This is would be okay with if we have it in core IF it meets those requirements. (and I would also propose we build it on the next branch)

So just as an idea here is what an API that meets the requirements of 2. would look like for me: (Note that this is just the parameters aspect, not the implementation)

```javascript
const wsParams = {
  // ... other wavesurfer options

  // ... not to be confused with splitChannel, default is false
  splitWaveform: {
    // options like barWidth and progressColor can be defined
    // here to apply to the bottom half of the waveform
  }
};
```

As you can see defining `splitWaveform` causes wavesurfer to render a complete set of `progressWave` and `wave` forms at the bottom which (can) have completely different wave style properties – This would certainly be very powerful when it comes to creating custom styles for waves. It would also be pretty complex to implement because of the way the renderer works currently. But I would rather do that than implement one parameter which does one thing and then have people opening issues and feature requests for all eternity "how can I make the waveform look like X"

src/drawer.multicanvas.js
```diff
@@ -180,11 +180,48 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
...
         var scale = length / width;
-
-        for (var i = (start / scale); i < (end / scale); i += step) {
-            var peak = peaks[Math.floor(i * scale * peakIndexScale)] || 0;
-            var h = Math.round(peak / absmax * halfH);
-            this.fillRect(i + this.halfPixel, halfH - h + offsetY, bar + this.halfPixel, h * 2);
+        var maxPeak = -Infinity;
+        var peaksToDraw = [];
+
+        if(this.params.waveStyle == 'soundWave'){
```
mspae: Can you check for the parameter `this.params.normalizeBars` here

src/wavesurfer.js
```diff
@@ -37,6 +37,7 @@ var WaveSurfer = {
         skipLength    : 2,
         splitChannels : false,
         waveColor     : '#999',
+        waveStyle: 'default'
```
mspae: Can you replace this with `normalizeBars: true`

src/drawer.multicanvas.js
```diff
@@ -180,11 +180,48 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
...
+                var pk = Math.abs(peaks[idx]) + Math.abs(peaks[idx + 1])  || 0;
+                //Set max peak:
+                if(maxPeak < pk){
+                    maxPeak = pk;
+                }
+                //Save peak to draw:
+                var item = {};
+                item.index = j;
+                item.fixedIndex = idx;
+                peaksToDraw.push(item);
```
mspae: Can you use an object literal here:

```javascript
peaksToDraw.push({
  index: j,
  fixedIndex: idx
});
```

PR #1025, opened 2017-03-04
Next: Updated plugin definition format and class exports

mspae:
* For the reasons described here: https://github.com/katspaugh/wavesurfer.js/pull/1015#issuecomment-283609182 the plugin definition format was updated.
    * All plugin classes are turned into observers when they are passed through `addPlugin`. (The plugin classes have the `@inherits {Observer}` jsdoc tag so the documentation generator gets things right.)
* Plugin classes are exported from plugin files and have a static `create` function, which is the factory function for the plugin definition object (which includes the plugin class as the `instance` property).
* Other minor changes:
  * Fixed wrong source map config and disabled hot module reloading
  * Fixed an issue related to https://github.com/katspaugh/wavesurfer.js/issues/736 (minimap does not render on initial load with backend `MediaElement`) by checking which backend is used and using the correct event to hook into. (`ready` or `waveform-ready`)
  * Updated jsdoc comments and added a correct `PluginClass` interface class (is removed in minified build by build system)
  * Updated upgrade guide
  * Updated example html and js

* The plugin definition factory function is now the static `create` method of the plugin class, which is the default export of the plugin files. This means code needs to be adjusted (see Code example)

* The above changes have been applied within the code base, so there should be no internal breakage.

### Todos/Notes:
* Esdoc now parses the plugin classes (the ones that are exported, not the single `Region` class), this has caused the documentation coverage to decrease to `57.96% (171/295)`

### Code example

```javascript
// EITHER - accessing modules with <script> tags
var WaveSurfer = window.WaveSurfer;
var TimelinePlugin = window.WaveSurfer.timeline;
var MinimapPlugin = window.WaveSurfer.minimap;

// OR - importing as es6 module
import TimelinePlugin from 'wavesurfer.js/dist/plugin/wavesurfer.timeline.min.js';
import MinimapPlugin from 'wavesurfer.js/dist/plugin/wavesurfer.minimap.min.js';

// OR - importing as require.js/commonjs modules
var WaveSurfer = require('wavesurfer.js');
var TimelinePlugin = require('wavesurfer.js/dist/plugin/wavesurfer.timeline.min.js');
var MinimapPlugin = require('wavesurfer.js/dist/plugin/wavesurfer.minimap.min.js');

// ... initialising waveform with plugins
    container: '#waveform',
    waveColor: 'violet',
    plugins: [
        TimelinePlugin.create({
            container: '#wave-timeline'
        }),
        MinimapPlugin.create()
    ]
});
```

### DIFF
--- a/example/cursor/index.html
+++ b/example/cursor/index.html
@@ -61,7 +61,7 @@
 <pre><code>var wavesurfer = WaveSurfer.create({
     container: document.querySelector('#waveform'),
     plugins: [
-        WaveSurfer.cursor()
+        WaveSurfer.cursor.create()
     ]
 });
 </code></pre>
--- a/example/spectrogram/app.js
+++ b/example/spectrogram/app.js
@@ -12,7 +12,7 @@ document.addEventListener('DOMContentLoaded', function () {
         loaderColor   : 'purple',
         cursorColor   : 'navy',
         plugins: [
-            window.WaveSurfer.spectrogram({
+            window.WaveSurfer.spectrogram.default({
                 container: '#wave-spectrogram'
             })
         ]
--- a/example/spectrogram/app.js
+++ b/example/spectrogram/app.js
@@ -12,7 +12,7 @@ document.addEventListener('DOMContentLoaded', function () {
         loaderColor   : 'purple',
         cursorColor   : 'navy',
         plugins: [
-            window.WaveSurfer.spectrogram.default({
+            window.WaveSurfer.spectrogram.create({
                 container: '#wave-spectrogram'
             })
         ]
--- a/example/spectrogram/app.js
+++ b/example/spectrogram/app.js
@@ -5,14 +5,14 @@ var wavesurfer;
 // Init & load
 document.addEventListener('DOMContentLoaded', function () {
     // Create an instance
-    wavesurfer = window.WaveSurfer.create({
+    wavesurfer = WaveSurfer.create({
         container     : '#waveform',
         waveColor     : 'violet',
         progressColor : 'purple',
         loaderColor   : 'purple',
         cursorColor   : 'navy',
         plugins: [
-            window.WaveSurfer.spectrogram.create({
+            WaveSurfer.spectrogram.create({
                 container: '#wave-spectrogram'
             })
         ]
--- a/example/timeline/app.js
+++ b/example/timeline/app.js
@@ -11,7 +11,7 @@ document.addEventListener('DOMContentLoaded', function () {
         loaderColor   : 'purple',
         cursorColor   : 'navy',
         plugins: [
-            window.WaveSurfer.timeline({
+            window.WaveSurfer.timeline.default({
                 container: '#wave-timeline'
             })
         ]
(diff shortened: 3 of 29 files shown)


example/cursor/index.html
```diff
@@ -61,7 +61,7 @@ <h1 itemprop="name">Cursor Plugin Example</h1>
 <pre><code>var wavesurfer = WaveSurfer.create({
     container: document.querySelector('#waveform'),
     plugins: [
-        WaveSurfer.cursor()
+        WaveSurfer.cursor.create()
```
thijstriemstra: ill have to re-read your motivation but i'd def prefer cursor() over create(). This wouldn't work either would it:

```
var cursorPlugin = WaveSurfer.cursor.create();
...
plugins: [cursorPlugin]
...
```

example/spectrogram/app.js
```diff
@@ -12,7 +12,7 @@ document.addEventListener('DOMContentLoaded', function () {
         loaderColor   : 'purple',
         cursorColor   : 'navy',
         plugins: [
-            window.WaveSurfer.spectrogram({
+            window.WaveSurfer.spectrogram.create({
```
thijstriemstra: in the regions example you removed `window.`, why not here?

example/timeline/app.js
```diff
@@ -11,7 +11,7 @@ document.addEventListener('DOMContentLoaded', function () {
         loaderColor   : 'purple',
         cursorColor   : 'navy',
         plugins: [
-            window.WaveSurfer.timeline({
```
thijstriemstra: See previous comment.

src/plugin/microphone.js
```diff
@@ -13,309 +13,338 @@
...
- * Microphone plugin definition factory
+ * Visualise microphone input in a wavesurfer instance.
  *
- * @param  {MicrophonePluginParams} params parameters use to initialise the plugin
- * @return {PluginDefinition} an object representing the plugin
+ * @implements {PluginClass}
+ * @extends {Observer}
+ * @example
+ * // es6
+ * import MicrophonePlugin from 'wavesurfer.js/dist/plugin/wavesurfer.microphone.js';
```
thijstriemstra: while it's technically correct, I would leave out the whole path and just import from 'wavesurfer.microphone.js'.

src/plugin/microphone.js
```diff
@@ -13,309 +13,338 @@
...
+        this.wavesurfer.un('backend-created', this._onBackendCreated);
+        this.stop();
+    }
+
+    /**
+    * Allow user to select audio input device, eg. microphone, and
+    * start the visualization.
+    */
+    start() {
+        navigator.mediaDevices.getUserMedia(this.constraints)
```
thijstriemstra: this doesn't work in all browsers. Why remove the previous logic?

example/cursor/index.html
mspae: See comment https://github.com/katspaugh/wavesurfer.js/pull/1025#pullrequestreview-25144096

PR #1027, opened 2017-03-04
return array when using getImage()

entonbiba:
#1012 #1011


thijstriemstra:
I would much rather have a data uri with a single canvas and an array with multiple canvasses. That way existing code doesn't break.

mspae:
@entonbiba He means only export an array if there are multiple canvases. Code which relied on the function returning a string and which worked before will still work. (Assuming there weren't multiple canvases)

thijstriemstra:
Come on, it returned a string before, and now it returns an array.

entonbiba:
@thijstriemstra ok lol, updated :)

src/drawer.multicanvas.js
```diff
@@ -337,6 +337,10 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
             var getEntry = this.canvases[i].wave.getContext('2d');
             availableCanvas.push(getEntry.canvas.toDataURL(type, quality));
         }
-        return availableCanvas;
+        if(this.canvases.length == 1){
```
thijstriemstra: spaces around parentheses.

src/drawer.multicanvas.js
```diff
@@ -332,11 +332,15 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
 
     getImage: function(type, quality) {
         // combine all available canvasses together
-        var availableCanvas = '';
+        var availableCanvas = [];
         for (var i in this.canvases) {
             var getEntry = this.canvases[i].wave.getContext('2d');
-            availableCanvas += getEntry.canvas.toDataURL(type, quality);
+            availableCanvas.push(getEntry.canvas.toDataURL(type, quality));
```
mspae: 1. The `entry` object already includes a cached version of the 2d context.
2. The `canvas` property of the 2d context is just a reference to the underlying canvas HTMLElement. So at the moment this code is unnecessarily complex. Please do instead: `availableCanvas.push(entry.wave.toDataURL(type, quality));` – No need for `getEntry`

src/drawer.multicanvas.js
```diff
@@ -332,11 +332,15 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
 
     getImage: function(type, quality) {
         // combine all available canvasses together
-        var availableCanvas = '';
+        var availableCanvas = [];
         for (var i in this.canvases) {
```
mspae: `this.canvases` is an array, please use `this.canvases.forEach(function (entry) {`– This is a pattern which we shouldn't use in this context. see https://github.com/katspaugh/wavesurfer.js/pull/961

src/drawer.multicanvas.js
```diff
@@ -332,11 +332,15 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
...
             var getEntry = this.canvases[i].wave.getContext('2d');
-            availableCanvas += getEntry.canvas.toDataURL(type, quality);
+            availableCanvas.push(getEntry.canvas.toDataURL(type, quality));
+        }
+        if (this.canvases.length == 1) {
+            return availableCanvas[0];
+        } else {
+            return availableCanvas;
```
mspae: It's a matter of taste, but I'd prefer if you did `return availableCanvas.length > 1 ? availableCanvas : availableCanvas[0];`

src/drawer.multicanvas.js
```diff
@@ -332,11 +332,15 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
 
     getImage: function(type, quality) {
```
mspae: Can you add Jsdoc comments?

entonbiba:
@mspae updated

mspae:
Thanks!

PR #1032, opened 2017-03-05
one more change for playlist plugin sample

entonbiba:
- remove extra parameters text
- add `audio/mpegurl` to playlistType option
- increase sample height for playlist plugin

### DIFF
--- a/plugins/playlist.html
+++ b/plugins/playlist.html
@@ -6,20 +6,9 @@ layout: default
 
   <p>Adds a playlist capability to wavesurfer.</p>
 
-  <h2>Parameters</h2>
-  <p>
-    The `playlistFile` parameter requires a valid http link to a supported playlist file type.<br/><br/>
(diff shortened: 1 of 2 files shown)


plugins/playlist.html
```diff
@@ -6,20 +6,9 @@
 
   <p>Adds a playlist capability to wavesurfer.</p>
 
-  <h2>Parameters</h2>
-  <p>
-    The `playlistFile` parameter requires a valid http link to a supported playlist file type.<br/><br/>
-    The `playlistType` parameter will default to `m3u` unless specified.<br/>
```
mspae: Can we use `<p>` tags instead of manual carriage returns?

entonbiba: @mspae yea I removed that section, I already added the tables.

btw I added the version badge to the footer file and linked to github project page, what do you think?

PR #1033, opened 2017-03-05
Check AC State

Stoyvo:
Should not close audiocontext when it's already closed. Chrome throws a warning if done so.


mspae:
When https://github.com/katspaugh/wavesurfer.js/pull/1021/ is merged it will reintroduce the global audio context singleton. Which means destroying it will be done in the `destroy` function if the parameter `closeAudioContext` is set to true. So this will be a manual process. We could add the check there and complain if it's already closed.

Issue #943, opened 2017-01-29
empty + redrawBuffer breaks player with pre-rendered peaks


omerts:
Hey,

We have a player we pre-rendered peaks. 
We have bounded to the window's resize method, and onresize call:
```javascript
wavesurfer.empty()
wavesurfer.redrawBuffer()
```
This works fine after play has been clicked (aka wavesurfer.play()).
The problem is before play has been clicked, if the window has been resized, it breaks the player, and it doesn't play once play is clicked.

Here is a reproduction example:
http://codepen.io/omerts/pen/EZbBZw

Click break -> play, will not work
Click play -> break, will work

Omer

entonbiba:
@omerts here you go, a working example.
http://codepen.io/entonbiba/pen/VPqvME

mspae:
Still an issue?

entonbiba:
don't think, works fine in the example

mspae:
Ok. I'll close this for now then.

Issue #932, opened 2017-01-27
partialRender with zoom = broken


Stoyvo:
Hey Folks,

Pulled the latest, started using partialRender and WOW big performance boost... but also not working with MultiCanvas.

![image](https://cloud.githubusercontent.com/assets/6148099/22360332/bad40294-e41c-11e6-8e64-9f57fb0bc190.png)

Notice the waveform getting "fuzzy". It's either fuzzy or a solid blank line depending on zoom level.

Config is as follows:

    var WaveSurfer = {
        defaultParams: {
            height        : 60,
            waveColor     : '#999',
            progressColor : '#555',
            cursorColor   : '#333',
            cursorWidth   : 1,
            skipLength    : 2,
            minPxPerSec   : 30,
            pixelRatio    : window.devicePixelRatio || screen.deviceXDPI / screen.logicalXDPI,
            fillParent    : true,
            scrollParent  : false,
            hideScrollbar : true,
            normalize     : true,
            audioContext  : null,
            container     : null,
            dragSelection : false,
            loopSelection : false,
            audioRate     : 1,
            interact      : true,
            splitChannels : false,
            mediaContainer: null,
            mediaControls : false,
            renderer      : 'MultiCanvas',
            backend       : 'WebAudio',
            mediaType     : 'audio',
            autoCenter    : true,
            partialRender : true,
            quickCenter   : true
        },

Stoyvo:
Looks like "normalize" option was causing the issue.

thijstriemstra:
@chrisparton1991 thoughts?

Stoyvo:
Any sugguestions @thijstriemstra @chrisparton1991 ?

thijstriemstra:
can you try to reproduce with less options? what example audio clip are you using? kinda hard to give meaningful feedback on this bug report.

Stoyvo:
@thijstriemstra These are all the options plus some overrides. Provided an exact list to help make it easier to replicate. It happens on all files, tested with MP3. Only happens when using partial rendering. Noticed it mostly with normalize on.

entonbiba:
@Stoyvo why exactly are you using partialRender when using MultiCanvas backend?

Stoyvo:
@entonbiba Specifically for the performance increase while zooming and because also because there's nothing saying I shouldn't.

Either:
1) We fix the issue with the rendering
2) We add handling to restrict this as an option and fall back to the suggested/required setting

[...]

PR #1035, opened 2017-03-06
Next: Ports from master and bugfixes

mspae:

* This ports hopefully all the (relevant) recent commits on the master branch to the next branch.
* nullified `this.arraybuffer` in `wavesurfer.destroy` to fix memory leak

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,26 @@
+# Hey, thank you for testing and contributing to wavesurfer.js!
+
+## Please make sure you can check all of these points below before opening an issue:
+
+(You don't have to post this section)
+
+- [ ] I have checked [the FAQ](https://wavesurfer-js.org/faq/) and it doesn't solve my problem.
+- [ ] I have checked [the documentation](https://wavesurfer-js.org/docs/) and it doesn't solve my problem
+- [ ] I have searched for [already open issues](https://github.com/katspaugh/wavesurfer.js/issues) which desribe my problem.
+- [ ] The issue I'm having is related to and caused by wavesurfer.js, not by other software (which maybe packages and uses wavesurfer incorrectly) – In that case you should open the issue on the respective project pages.
+
+## Please make sure you provide the following information (if applicable):
+
+### Wavesurfer.js version(s):
+
+
+### Browser version(s):
+
+
+### Code needed to reproduce the issue:
+
+(Please reduce your code as much as possible and only post the minimum code needed to reproduce the issue. [A Code pen](http://codepen.io/) is an excellent way to share such code)
+
+
+### Use behaviour needed to reproduce the issue:
+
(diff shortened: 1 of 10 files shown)


mspae:
@katspaugh @thijstriemstra @entonbiba AFAIK we could release `2.0.0-beta01` what do you say?

entonbiba:
@mspae sounds great! haven't checked but are all the plugins available in version 2?

thijstriemstra:
Sounds good. A branch with a release is fine, ill help test.

entonbiba:
@mspae to celebrate st. patricks day I want to give away a few free download codes for my DJ Mixer app, maybe even throw in a couple when version 2 is released. What do you guys think is that alright? Should I create an issue?

mspae:
@entonbiba I don't mind if you give away free download codes, but I think we should make it clear that it's a different project, otherwise people might open mixer app related issues here and vice versa.

@thijstriemstra Thanks for the support!

So there definitely are a few issues which are still open, but the library as a whole is functional. Here's an overview of the current status:

* All but the newest plugins are ported. (I think porting them will be uncomplicated)
  * The **timeline** plugin zooming performance issues of the v1 version are still there
  * The **minimap** plugin works *mostly* – There are still some issues because of `ready`/`waveform-ready` event inconcistencies (WebAudio and MediaElement backends with peaks and without) – I'm currently working on that and will open a PR to make the events more consistent (and document them) – then the minimap plugin should work properly. (overviewRegion and rendering of regions works)
  * The **spectrogram** labels PR I've ported
  * **regions**, **microphone** and **elan** plugin wasn't changed so much, they should work as before.
* Open bugs AFAIK the same as on master
  * Minimap plugin does not render complete waveform with large files https://github.com/katspaugh/wavesurfer.js/issues/994
  * Waveform display problem with timeline (see above) https://github.com/katspaugh/wavesurfer.js/issues/990
  * Bad zooming performance when using timeline plugin (see above) https://github.com/katspaugh/wavesurfer.js/issues/967
  * empty + redrawBuffer breaks player with pre-rendered peaks https://github.com/katspaugh/wavesurfer.js/issues/943
  * partialRender with zoom = broken https://github.com/katspaugh/wavesurfer.js/issues/932
* Also updating the jsdoc comments on the plugins would be a super way of providing complete documentation for all the plugin APIs. (The initialisation parameters are already documented that way but I think especially the regions plugin would benefit greatly from a complete documentation)
* Other things:
  * Responsive parameter was added
  * Minimap plugin doesn't have a mandatory `container` init param, if none is passed it is just appended, I would like to do the same for the timeline plugin.
  * A memory leak (nullifying `this.arraybuffer` on `wavesurfer.destroy()`) was fixed, this could be backported. I'm sure there are some other memory improvements that could and should be done, but for me fixing bugs are currently the 1. priority.
  * I would recommend we create a `v2` directory in the `gh-pages` branch to contain the website for v2, then we could link to it from the current v1 page (like the webpack project website: https://webpack.github.io/) – Then we could cleanly seperate marketing content and examples from the actual (generated) technical documentation. The latter will be always up-to-date since it is part of the release workflow to generate new docs and push the `gh-pages` branch.

Issue #1041, opened 2017-03-16
Documentation site doesn't list setMute method


tomasvilda:
On documentation website https://wavesurfer-js.org/docs/methods.html

Method setMute is not documented

entonbiba:
@tomasvilda thanks, added

PR #1044, opened 2017-03-16
add setMute() method to docs

entonbiba:
add setMute() method to docs #1041


tomasvilda:
setMute() has boolean parameter, so it can mute on true and unmute on false.

thijstriemstra:
please see prev comment

entonbiba:
@thijstriemstra updated

thijstriemstra:
is it really called `boolean`?

entonbiba:
@thijstriemstra updated, set it as boolean since it had true/false values ><

Issue #1040, opened 2017-03-16
Wavesurfer doesn't reset mute state on next loaded source


tomasvilda:
When I do

wavesurfer.load(file);
wavesurfer.play();

file is played with sound

wavesurfer.muteToggle();

now sound is muted

wavesurfer.load(nextfile);
wavesurfer.play();

file is played with sound (this is expected as I have loaded new file)

wavesurfer.muteToggle();

file is still played with sound (Wrong! as I see in source isMuted is not reset on load() method)

wavesurfer.muteToggle();

after second method call sound is muted.

entonbiba:
@tomasvilda fixed in the pr

PR #1045, opened 2017-03-16
set isMuted to false on load method

entonbiba:
set isMuted to false on new load

#1040


entonbiba:
@katspaugh should I add this as an option to allow muting to continue even after loading new audio?

tomasvilda:
That would logical too. If I mute wavesurfer object maybe I want it to be muted on new tracks too, until I unmute.

katspaugh:
Sure, let's add it. To me, it seems more logical to actually keep the next tracks muted by default since it's the same instance of the wavesurfer. But it's subjective.

katspaugh:
Merged, thanks!

PR #1048, opened 2017-03-17
add split-wave-plot drawer for drawing points above wave

jrweth:
**Title: Added an extension of Wavesurfer.Canvas.Drawer which renders a time aligned point graph above the waveform** 

For some academic purposes at the University of Pennsylvania I needed to incorporate a pitch graph to accompany the sound wave.  I generalized the problem by creating a drawer which extends the canvas drawer to do this (it spoofs being a split channel to do this).  I didn't touch any of the existing files and simply added the **drawer.split-wave-point-plot.js** file as well as the example files in the **examples/split-wave-point-plot** folder.  

I did run the grunt process so all of the dist files got updated, but it was really just the one source file that I added.

So feel free to pull or not.  Others may find it useful.  Thanks so much for heading up this effort.

Shouldn't have changed anything except if someone wants to both split channels and use this renderer.  It won't split the channels appropriately, but only display one and display the point plot instead of the second channel.

As noted above, splitting channels.


entonbiba:
@jrweth sounds interesting, can you post a demo on codepen using this change?

jrweth:
I don't have a codepen pro account so I would rather not upload.  I did commit a demo to the repo in the example/split-wave-point-plot folder.  I also have something online to look at (although it also incorporates 2 other plugins extensions to the Elan plug-in so much more is going on there).  That demo can be seen at http://jrweth.com/penn_sound/i_know_a_man.html

jrweth:
I threw up the example on my website for a look at a simpler version.   http://www.jrweth.com/wavesurfer2/example/split-wave-point-plot/index.html

piem:
thank you for this @jrweth, this will be very useful for us too to display annotations extracted with [aubio](/aubio/aubio).

best, Paul

katspaugh:
It looks amazing! Thanks a lot! :thumbsup:

jrweth:
Thanks for the catch with the thickening.  I have fixed the issue (along with another one I found).

katspaugh:
Merged, thank you!

PR #1054, opened 2017-04-04
Prevent waveform vertical scrolling in chrome

dbunker:
The waveform is scrollable vertically in Chrome, however should only be scrollable horizontally, as is the case in Firefox. This can be observed here: http://codepen.io/katspaugh/pen/GozaaG, when the screen width is sufficiently reduced to result in horizontal scrolling.


katspaugh:
Thanks! 👍

PR #1061, opened 2017-04-06
Fixed styling bug in the spectrogram canvas height

zlapper:
There was a bug in the spectrogram canvas height displayed. Somehow this was more evident in retina displays, but affected all nonetheless.


plugin/wavesurfer.spectrogram.js
```diff
@@ -153,9 +153,11 @@ WaveSurfer.Spectrogram = {
 
     updateCanvasStyle: function () {
         var width = Math.round(this.width / this.pixelRatio) + 'px';
+        var height = this.height + 'px';
         this.canvas.width = this.width;
         this.canvas.height = this.height;
         this.canvas.style.width = width;
+        this.canvas.style.height = height;
```
katspaugh: Shouldn't the height be a subject to pixel ratio conversion, like the width?

entonbiba:
@zlapper can you provide a screenshot of the height issue?

zlapper:
@entonbiba sure thing!

**Before fix:**
Before fix
Notice how, altough the canvas element has the `height="256"` property, it is being rendered as 128px. Notice also that there is a (css) style for the width, but not for the height.

----

**After fix:**
After fix
Here you can see the element being rendered correctly, just by adding the height value to the element styles.

Regarding @katspaugh's comment, I have the same impression, but it is not necessary according to my tests (i.e. the height style value ends up being 128px, not 256px, as the element height property). I might be confused, tho :)

entonbiba:
@zlapper  I noticed the sample you used is old and using an outdated version of the library. can you confirm the same issue is happening with this sample?
http://codepen.io/entonbiba/pen/Bpmqrq

thijstriemstra:
any feedback @zlapper?

thijstriemstra:
I suggest we close this if there is no more feedback.

entonbiba:
@zlapper please refer to the following sample which uses an updated version for testing. If issue persists you can reopen this pr.
http://codepen.io/entonbiba/pen/Bpmqrq

closing for now.

Issue #1067, opened 2017-04-18
Load peaks from file post v1.2.8


X-Raym:
Hi !

With v1.2.8,
To get peaks and create peaks file from it,
I use to use this:

`peaks = object.backend.getPeaks(960);`

This seems to be broken post v1.2.8 (like v1.4.0).

After quite some testing, I found that the following code produces the same json file than in v1.2.8

`peaks = object.backend.getPeaks(960, 0, 959);`

Note: I think there should have been something do about this to assure **backward compatibility**, as I spent hours on it, thinking the problem came from elsewhere.

Well I thought that it was good, but despite the files are now the same as before,
it seems that something have also change in the way they are render !

Here is my code:
[wavesurfer.load(file_url, peaks, preload);](url)

and here is what I got with v1.2.8

![V1.2.8](https://monosnap.com/file/NtrEwtLZMqXTvWfF0cEmKPulsvJwZF.png)

And here is the same peaks files, but loaded with v1.4.0 (and v1.3.x)

![v1.4.0](https://monosnap.com/file/vRaT3wznOnle4ogL0wCS9CTD4jixOY.png)

Why is there such difference ?
What is the new way to load peaks ?
Could backward compatibility be assured right within wavesurfer-js ?

Thanks for your assistance !

X-Raym:
@katspaugh I did a bit more careful testing, in fact the problem appears from v1.3.0. v1.2.8 works great, but not after that.

entonbiba:
@X-Raym yea the getPeaks method should not break when using just `getPeaks(960)`, I created a pull request to fix the back-end compatibility for this method.

Not sure why the rendering it doing that for you. Can you provide the peaks and audio you're using? Also what are you setting as the preload?

X-Raym:
Thanks for the `getPeak` fix !

For my case: 

- [Audio File](https://www.extremraym.com/wp-content/uploads/2016/12/SalmanKhan_2015P.mp3)
- [Peaks File](https://www.extremraym.com/wp-content/uploads/wavesurfer-wp-peaks/1fdb6302422ea3b5b368136c3fba8c7e-mixed.json).

(of course, I only use the 'peaks' section of my peak file in the wavesurfer load method).

`Preload` is set to `'metadata'`.

Thanks for your support :)

X-Raym:
https://github.com/katspaugh/wavesurfer.js/issues/1055
this issue seems to be related to my problem.

entonbiba:
[...]

PR #1072, opened 2017-04-22
add getPeaks background compatibility

entonbiba:
Fix to allow old style getPeaks(length) usage with new getPeaks(length,first,last).
If no `first` and `last` parameter is set it will set the default `0` and `length-1`.

#1067

### DIFF
--- a/src/webaudio.js
+++ b/src/webaudio.js
@@ -217,7 +217,7 @@ WaveSurfer.WebAudio = {
      */
     getPeaks: function (length, first, last) {
         if (this.peaks) { return this.peaks; }
-        
+
         var first = first || 0;
         var last = last || length-1;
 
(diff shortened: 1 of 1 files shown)


src/webaudio.js
```diff
@@ -218,6 +218,9 @@ WaveSurfer.WebAudio = {
     getPeaks: function (length, first, last) {
         if (this.peaks) { return this.peaks; }
 
+        var first = first || 0;
```
katspaugh: No need for `var`, since it's already declared as an argument.

src/webaudio.js
```diff
@@ -218,6 +218,9 @@ WaveSurfer.WebAudio = {
...
+        var last = last || length-1;
```
katspaugh: `last = last || length - 1;` (no `var` and spacing around the minus)

entonbiba:
@katspaugh updated :)

katspaugh:
Thanks! 👍

PR #1073, opened 2017-04-23
fix spectrogram plugin frequency labels position

entonbiba:
fix spectrogram plugin frequency labels position `absolute`


thijstriemstra:
we should have some unit tests one day that verify these changes, e.g. a png image with the expected output of the waveform.

thijstriemstra:
Anyway, can you add a comment on the line above explaining why it should be `absolute`, and then I'll approve it.

entonbiba:
@thijstriemstra, currently:
image


with the absolute fix
image

thijstriemstra:
Thanks!

Issue #1062, opened 2017-04-10
Dragging Regions after initialisation to false


jawrainey:
Hi all,

I would like to add regions that are by default not resizable or draggable. When a user requests to edit a region, then it will be made resizable/draggable programmatically. However, I could not make these regions resizable or draggable once they are initialised to false. Although there are ways around this (such as doing another pass after adding and setting these properties to false), this feels like the wrong approach.

I thought I would raise this issue before jumping into the codebase and proposing changes as **(1)** what I am trying to achieve may be expected behaviour, or **(2)** this is a known issue that I have not been able to find more details on or **(3)** I am using the API in an alternative way. 

I'm happy to contribute a PR after discussing this further if this is an issue 👍 

<hr>

### Source &amp; Browser versions

- **Wavesurfer.js:** 1.3.7
- **Wavesurfer.regions.js:** 1.3.7
- **Google Chrome:** Version 55.0.2883.95 (64-bit)
- **Firefox:** 52.0.2 (64-bit)

### Code needed to reproduce the issue:

- http://codepen.io/anon/pen/oZKdZd

### Use behaviour needed to reproduce the issue:

- Try dragging the first (green) region.

entonbiba:
@jawrainey try this one

http://codepen.io/entonbiba/pen/mmOxmm 
```
wavesurfer.regions.list[2].drag = true;
wavesurfer.regions.list[2].resize = true;
```

jawrainey:
@entonbiba, thanks for the response, though I think I did not describe the issue sufficiently! The issue is in relation to moving the first (green) region.

Changing these to true will only work because the object properties are not set (_drag_ & _resize_):

    wavesurfer.addRegion({id: 2, start: 28, end: 36, color: 'hsla(400, 100%, 30%, 0.1)'});

> @jawrainey try this one, it might be a bug with the .update() part of the plugin.

The update method works when the properties are not set when creating the object (as above):

    wavesurfer.regions.list[2].update({drag: false, resize: false});

The issue is that if these properties are not set, then I cannot update the region object to move the region. What's interesting is that the update method, or updating the properties manually as you have suggested does work, but does not have any affect!

http://codepen.io/anon/pen/oZKdZd

Note that the hover icon changes when trying to move the green region, and the properties are set the true

entonbiba:
@jawrainey you should use `wavesurfer.region` instead of `wavesurfer.regions`

http://codepen.io/entonbiba/pen/zwZyOg
```
[...]

PR #1076, opened 2017-05-01
fix regions update method

entonbiba:
**to do:**
- fix cursor on on drag
- create method for 
```
if(!currentDrag || !currentResize){
    this.bindEvents(regionEl);
}
```

enables resize and drag after user has created regions with params set to false. #1062

```
  wavesurfer.addRegion({
    id: 1,
    start: 2, // time in seconds
    end: 5, // time in seconds
    color: "hsla(100, 100%, 30%, 0.1)",
    drag: false, <---
    resize: false <---
  });
```

before fix:
http://codepen.io/entonbiba/pen/ZKKadq
```
  wavesurfer.regions.list[1].update({ resize: "true", drag: "true" });
// won't work since .update doesn't do anything since already set to false in addRegion. 
```

after fix:
http://codepen.io/entonbiba/pen/zwZyOg
```
  wavesurfer.regions.list[1].update({ resize: "true", drag: "true" });
// works as expected
```


plugin/wavesurfer.regions.js
```diff
@@ -182,6 +184,36 @@ WaveSurfer.Region = {
...
+            this.style(handleRight, css);
+            this.style(handleRight, {
+                left: '100%'
+            });
+            this.element = this.wrapper.appendChild(regionEl);
+        }
+
+        if(!currentDrag || !currentResize){
+            this.bindEvents(regionEl);
+        }
```
katspaugh: I think it's better to extract this part of the [`render` function](https://github.com/katspaugh/wavesurfer.js/blob/master/plugin/wavesurfer.regions.js#L213-L257) into a separate function and call it from both places passing either a freshly created region element or an existing element.

entonbiba: @katspaugh yes that sounds good, I'll do that

agamemnus:
Fundamentally, this needed some more changes: https://github.com/katspaugh/wavesurfer.js/pull/1103.

agamemnus:
There are really two ways to do this.

1) Adding/removing the elements themselves.
2) Updating whether they are displayed via the DOM.
3) Updating the logic to say whether mouse events on them should have any effect.

They all have pros and cons...

thijstriemstra:
Please open a new pull request against the latest master branch, this one diverged too much.

Issue #1014, opened 2017-02-27
Ember.js Uncaught TypeError: Cannot set property 'fillStyle' of undefined


davidchalifoux:
Hey, so I'm trying to use WaveSurfer within my Ember.js project with no prevail(I doubt Ember is causing these issues).

I imported the latest version of wavesurfer.js through Bower, and then used this exact code in one of my templates:
```
    <div id="waveform"></div>
    <script>
        var wavesurfer = WaveSurfer.create({
            container: '#waveform',
            waveColor: 'red',
            progressColor: 'purple'
        });
        wavesurfer.load('assets/sound4.mp3');
    </script>
```

Now when I try to see if it works, my console errors saying `Uncaught TypeError: Cannot set property 'fillStyle' of undefined`.

Here's what line in wavesurfer.js it's erroring at:

```
    setFillStyles: function (entry) {
        entry.waveCtx.fillStyle = this.params.waveColor; <--- HERE
        if (this.hasProgressCanvas) {
            entry.progressCtx.fillStyle = this.params.progressColor;
        }
    },
```

thijstriemstra:
can you try to reproduce it without ember?

davidchalifoux:
@thijstriemstra Actually I tested it out about 30min ago and got it to work. I wonder why it doesn't work within Ember?

thijstriemstra:
I'm not familiar with ember.js but if you figure it out let us know here.

davidchalifoux:
@thijstriemstra a user on StackOverflow sent me [this](https://github.com/katspaugh/wavesurfer.js/issues/950) discussion and I seem to be in the same boat somehow. I rolled back from 1.3.4 to 1.2.8 and it's working now. Any suggestions as to how I could get it to where I can be using the current versions?

agamemnus:
@thijstriemstra I have the same problem, but I am not using Ember. Investigating...

mspae:
Actually I think there was an issue a few months back somebody had while using ember. I think the cause was incorrect `for (entry in this.canvases)` (`this.canvases` being an array and not an object) https://github.com/katspaugh/wavesurfer.js/pull/961  and https://github.com/katspaugh/wavesurfer.js/issues/947 – Maybe this is related somehow.

agamemnus:
Thank you. This was exactly the problem!

davidchalifoux:
Hey, I just tried out the 2.0 beta and it's working great! Good work!

PR #1081, opened 2017-05-05
"for... in" should not be used with arrays. This causes bugs!

agamemnus:
See: https://github.com/katspaugh/wavesurfer.js/issues/1014.

I also changed "entry" to "canvas" since they are canvases. And, uh, just a few styling ideas.

### DIFF
--- a/src/drawer.multicanvas.js
+++ b/src/drawer.multicanvas.js
@@ -67,8 +67,8 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
     },
 
     addCanvas: function () {
-        var canvas = {};
-        var leftOffset = this.maxCanvasElementWidth * this.canvases.length;
+        var canvas = {},
+            leftOffset = this.maxCanvasElementWidth * this.canvases.length;
 
         canvas.wave = this.wrapper.appendChild(
             this.style(document.createElement('canvas'), {
@@ -107,8 +107,8 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
     },
 
     updateDimensions: function (entry, width, height) {
-        var elementWidth = Math.round(width / this.params.pixelRatio);
-        var totalWidth   = Math.round(this.width / this.params.pixelRatio);
+        var elementWidth = Math.round(width / this.params.pixelRatio),
+            totalWidth = Math.round(this.width / this.params.pixelRatio);
 
         // Where the canvas starts and ends in the waveform, represented as a decimal between 0 and 1.
         entry.start = (entry.waveCtx.canvas.offsetLeft / totalWidth) || 0;
@@ -283,7 +283,8 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
 
     fillRect: function (x, y, width, height) {
         var startCanvas = Math.floor(x / this.maxCanvasWidth);
-        var endCanvas   = Math.min(Math.ceil((x + width) / this.maxCanvasWidth) + 1, this.canvases.length);
+        var endCanvas   = Math.min(Math.ceil((x + width) / this.maxCanvasWidth) + 1,
+            this.canvases.length);
 
         for (var i = startCanvas; i < endCanvas; i++) {
             var canvas = this.canvases[i];
--- a/src/drawer.multicanvas.js
+++ b/src/drawer.multicanvas.js
@@ -99,10 +99,10 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
     },
 
     removeCanvas: function () {
-        var lastCanvas = this.canvases.pop();
-        lastCanvas.wave.parentElement.removeChild(lastCanvas.wave);
+        var lastEntry = this.canvases.pop();
+        lastEntry.wave.parentElement.removeChild(lastEntry.wave);
         if (this.hasProgressCanvas) {
-            lastCanvas.progress.parentElement.removeChild(lastCanvas.progress);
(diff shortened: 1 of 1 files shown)


katspaugh:
Could you just replace the `for-in` with a `forEach` without the other stylistic changes? Btw, `forEach` takes a third argument which is the context of the callback, so you could pass `this` and not introduce another `var my = this` to keep the context.

agamemnus:
I already ended up removing many of those changes due to your draconian style checker! The rest is as follows --

There are only four differences now, other than the forEach fixes, and corresponding "my" variables -- the "my" variables isn't something I came up with as it was present in an existing function.
1) Some brackets had an errant starting space: e.g.: `{ width: elementWidth + 'px'}`. and the `getImage` function didn't have a space between it and the parameters.
2) `entry` -> `canvas` and `lastEntry` -> `lastCanvas`. IMO, if you have `this.canvases`, then one of them should be a `canvas` and not an `entry`.
3) Some breaking up of a multi-line var:

````
var elementWidth = Math.round(width / this.params.pixelRatio),
totalWidth = Math.round(this.width / this.params.pixelRatio);
````
to
````
var elementWidth = Math.round(width / this.params.pixelRatio);
var totalWidth   = Math.round(this.width / this.params.pixelRatio);
````
and
````
var entry = {},
leftOffset = this.maxCanvasElementWidth * this.canvases.length;
````
to
````
var canvas = {};
var leftOffset = this.maxCanvasElementWidth * this.canvases.length;
````
4)
````
var endCanvas = Math.min(Math.ceil((x + width) / this.maxCanvasWidth) + 1,
this.canvases.length);
````
to
````
var endCanvas = Math.min(Math.ceil((x + width) / this.maxCanvasWidth) + 1, this.canvases.length);
````

**Comment by katspaugh** [comment]
While some of your changes make sense (I'll mark them with a thumbs-up in the changeset), some are arbitrary, and I would rather not create more maintenance work for the `next` branch.

File: src/drawer.multicanvas.js [hunk 1]
```diff
@@ -52,24 +53,24 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
             this.removeCanvas();
         }
 
-        for (var i in this.canvases) {
+        my.canvases.forEach (function (canvas, i) {
```
katspaugh: 👍  except for the extra space after `forEach` – it's not needed.

src/drawer.multicanvas.js
```diff
@@ -91,51 +92,52 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
...
     },
 
     clearWave: function () {
-        for (var i in this.canvases) {
-            this.clearWaveForEntry(this.canvases[i]);
-        }
+        var my = this;
+        my.canvases.forEach (function (canvas) {
+            my.clearWaveForEntry(canvas);
+        });
```
katspaugh: 👍 

```
this.canvases.forEach(function (canvas) {
    this.clearWaveForEntry(canvas);
}, this);
```

src/drawer.multicanvas.js
```diff
@@ -223,31 +223,29 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
...
-        for (var index in this.canvases) {
-            var entry = this.canvases[index];
-
-            this.setFillStyles(entry);
-
-            this.drawLineToContext(entry, entry.waveCtx, peaks, absmax, halfH, offsetY, start, end);
-            this.drawLineToContext(entry, entry.progressCtx, peaks, absmax, halfH, offsetY, start, end);
-        }
+        var my = this;
+        my.canvases.forEach (function (canvas) {
```
katspaugh: The same comment as above.

src/drawer.multicanvas.js
```diff
@@ -333,16 +330,16 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
...
     /**
-     * Combine all available canvasses together.
+     * Combine all available canvases together.
      *
      * @param {String} type - an optional value of a format type. Default is image/png.
      * @param {Number} quality - an optional value between 0 and 1. Default is 0.92.
      *
      */
-    getImage: function(type, quality) {
+    getImage: function (type, quality) {
```
katspaugh: :thumbsup:

agamemnus:
OK, give me a few minutes to make the appropriate changes.

agamemnus:
Ok, I removed almost all of the other changes.

I also implemented your suggestion about inserting a `this` variable in forEach.

katspaugh:
I still see a lot of other changes in the diff.

agamemnus:
I fixed some errors.

The only other change now is removing space between opening/closing brackets and text. You don't really even have that consistent...

agamemnus:
Ok, a few more I didn't notice.

agamemnus:
I really hate those extraneous spaces, but I added them back. I also added some extra extraneous spaces in style definition calls for it to be consistent.

src/drawer.multicanvas.js
```diff
@@ -191,23 +190,22 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
...
+                }, this);
                 return;
             } else {
                 peaks = channels[0];
             }
         }
 
         // Support arrays without negative peaks
-        var hasMinValues = [].some.call(peaks, function (val) { return val < 0; });
+        var hasMinValues = [].some.call(peaks, function (val) {return val < 0;});
```
katspaugh: Could you revert this change, too?

agamemnus:
Ok.

katspaugh:
Merged, thanks!

PR #1096, opened 2017-05-16
fixes safari browser missing buffer when using webaudio

### LINKED ISSUE #952: How can I play one song and have 2 waveforms in sync? & safari render issue
In an app i'm making I need to show the waveform twice in separate pieces of ui. These would need to be in sync, is this possible? I see you allow split channels which is close to what i want but in my case it would be a straight duplicate.

Thanks,

entonbiba:
fixes #952 safari buffer undefined

```
var newBuffer = this.createBuffer(1, 4096, this.sampleRate);
```
- it creates the missing buffer based on 1 channel, 4096 samples and the sampleRate from the current webaudio context.

### DIFF
--- a/src/webaudio.js
+++ b/src/webaudio.js
@@ -223,6 +223,12 @@ WaveSurfer.WebAudio = {
 
         this.setLength(length);
 
+       /**
+        * The following snippet fixes a buffering data issue on the Safari browser which returned undefined
+        * It creates the missing buffer based on 1 channel, 4096 samples and the sampleRate from the current webaudio context
+        * 4096 samples seemed to be the best fit for rendering
+        * will review this code once a stable version of Safari TP is out
+        */
         if (!this.buffer.length) {
             var newBuffer = this.createBuffer(1, 4096, this.sampleRate);
             this.buffer = newBuffer.buffer;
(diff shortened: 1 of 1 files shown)


thijstriemstra:
why are these values (1 channel, 4095 samples) hard coded? I guess Safari doesn't supply these values somewhere?

entonbiba:
@thijstriemstra yea that's the problem it doesn't. Safari TP should return the correct buffer, but it's a long way until that will be available. For now that's the best we can do to fix it in safari.

src/webaudio.js
```diff
@@ -223,6 +223,11 @@ WaveSurfer.WebAudio = {
 
         this.setLength(length);
 
+        if (!this.buffer.length) {
```
katspaugh: Could you add a short comment regarding the Safari quirk?

thijstriemstra: ping @entonbiba

entonbiba: @katspaugh sure

entonbiba: @thijstriemstra @katspaugh done

katspaugh:
Thank you @entonbiba!

thijstriemstra:
> yea that's the problem it doesn't. Safari TP should return the correct buffer, but it's a long way until that will be available. 

@entonbiba can this workaround be removed yet?

PR #1097, opened 2017-05-17
add responsive parameter and set to true

entonbiba:
Responsive waveform on resize and device orientation change set to true by default

parameter `responsive` default `true`, can be set to `false` to disable responsive waveform.

**todo**
- create onResize method
- add event handler 
`this._onResize = this.onResize.bind(this)`

PR #1098, opened 2017-05-17
add responsive parameter to docs

entonbiba:
add `responsive` parameter to options

should be merged after pr #1097


katspaugh:
Thanks! 👍

PR #1100, opened 2017-05-18
add xhr with credentials option

rom1504:
When using an API that requires authentification with cookies, it is required to use the [withCredentials](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials) option.

This pull request add an option xhrWithCredentials to make retrieving media files from such APIs possible.


thijstriemstra:
@katspaugh do we have any params that accept js objects? We could introduce a `xhr` setting that accepts an object to avoid bloating the params..

katspaugh:
@thijstriemstra yeah, this sounds much better than passing the params individually. In the next version, we should use `fetch`.

thijstriemstra:
Please re-open a new pull request against current master branch.

Issue #1099, opened 2017-05-17
ES6 beta - problems Safari


rkuhl:
Great library!  
Using the ES6 beta ver `2.0.0-beta01`  
Noticed some problems on Safari v 10.0.2  
```javascript  
import WaveSurfer from "wavesurfer.js"
let wavesurfer = WaveSurfer.create( {container: "#waveform"} )
wavesurfer.on('ready', function () {
    wavesurfer.play();
});
wavesurfer.load("sample.mp3");
```
On Safari I get an error:  
_IndexSizeError (DOM Exception 1): The index is not in the allowed range._   
Everything works just fine on Chrome / Firefox.  
Thanks for your assistance!

entonbiba:
#1102

mspae:
Should be fixed once #1102 is released.

PR #1102, opened 2017-05-19
Fixed references to static property audioBufferSize

jraoult:
The way the static property `audioBufferSize` is referenced (through `this`) is not valid ES2015 code and generates an error in Safari iOS at least ("IndexSizeError (DOM Exception 1): The index is not in the allowed range." since undefined is not a valid buffer size)

My fix is using the class name to reference the property but we can also use `this.constructor`

N/A

N/A

### DIFF
--- a/src/webaudio.js
+++ b/src/webaudio.js
@@ -219,9 +219,9 @@ export default class WebAudio extends util.Observer {
     /** @private */
     createScriptNode() {
         if (this.ac.createScriptProcessor) {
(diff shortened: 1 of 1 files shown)


entonbiba:
#1099

mspae:
Thanks, @jraoult !

maxwelladapoe1:
Hello please has this issue been fixed? I still came across this issue in safari. It comes only in safari i am also using the latest release of wavesurfer

PR #1103, opened 2017-05-26
Fixes/enhancements to regions plugin.

agamemnus:
I'm modifying this for my own use and thought it could use a few additions and enhancements. More to come.

1] Fix for cursor fighting when dragging.
Previously, default cursor sometimes fought with dragging cursor, resulting in flickering cursors. The main issue here is that there is a cursor assigned to each region and there's no overall action state. This is actually a fairly complex problem. More work will be required to handle extreme edge cases. (although, it is already better than current code in that respect)

2] New events. `region-resized-start` and `region-resized-end`:
These will trigger when the region start or end is resized.

3a] New options. `resizeStart` and `resizeEnd`:
These determine whether the resize start and resize end control is created. (Sometimes, you might not want both)

3b] More new options: `cssResizeStyle`, `cssResizeStartClass`, `cssResizeEndClass`.
Optional parameters to change the default style of the resize controls. `cssResizeStyle` will overwrite the default style added to each resize control. `cssResizeStartClass` and `cssResizeEndClass` will add classes to the start/end controls.

4] Still more new options: `showTitle`.
Optional parameter `showTitle`. Default is true. If false, title will not be shown.


agamemnus:
This is a real hassle. Is there a way to just have it auto-format?

katspaugh:
Those are all good changes and useful options. Could you maybe add all of the changes to the `next` branch? We would like to make it the new master it in June.

Also, the code style is off in some places: inline objects need spacing inside, the `void` in unnecessary. I thought we had linting for this 🤔

agamemnus:
Is there a tool I can use to convert the code using those options? Using
this code style is a huge hassle.

agamemnus:
Well, it's not even particular to this code style. Just in general. Idk. I
know there's prettydiff but that doesn't do all the things in the styling
requirements that you need...

katspaugh:
`eslint --fix` should be able to auto-fix the code style.

agamemnus:
It does not work without a void, or I am guessing an extra pair of parentheses.

I made a few more changes:
* Instead of duplicating the fireEvent calls, I created a fireEventPropagate function. But then there are issues: what about `remove`+`region-removed`, and `update`+`region-updated`? It's not really consistent as the click event firings (dubious usefulness..) don't have that kind of structure -- it's the same for both the wavesurfer and region object.

* Optimized `onDrag` and `onResize` code.
* The update function will now only call the updateRender function if something actually changed.

agamemnus:
P.S.: Once I am done working on this... pretty close now... I'll make the changes in the other branch.

The only question now is the events. They are not backward-compatible since I changed a few of them from past-tense to present in order to standardize everything.

katspaugh:
I think we can change the names in the next branch, since the plugin init code has changed anyway.

agamemnus:
Great. Thanks. I think the edits are pretty much done now. I just added a mechanism to prevent the seek cursor from moving if the mouse is released. I also split up drag into dragstart and drag (after the cursor has moved) for future-proofing...

katspaugh:
Thanks! I'll need to check and test it, will get back to you.

agamemnus:
Sorry -- one more change.

agamemnus:
Actually, a few errors with the latest change, but I am falling asleep... tomorrow.

agamemnus:
@katspaugh: Another idea on the events:
A slider function I made has the parameter that states whether the update is final (i.e.: user does mouseup, etc.) as the last parameter.

agamemnus:
Duration of a clip? Timestamp?

???

@katspaugh : ok. So I dunno what is up with all these extra files. But anyway, I have a new file right here:

https://github.com/agamemnus/wavesurfer.js/blob/master/plugin/wavesurfer.regions.js

(There are quite a lot of changes. It's much improved... one thing I did recently was replace the multiple document mouseup/mousedown event listeners in favor of a single unified listener.)

agamemnus:
I have once more updated my code at:

https://github.com/agamemnus/wavesurfer.js/blob/master/plugin/wavesurfer.regions.js

Bueller!

agamemnus:
Closing as too many things have changed.

Moving this to https://github.com/katspaugh/wavesurfer.js/pull/1211.

PR #1104, opened 2017-05-27
Listen to and relay play and pause events on media element

nlicitra:
Added automatic listening to the `"play"` and `"pause"` events of the provided media element. In the current implementation, I can only control audio playback from within the context of `wavesurfer.js`. In the application I'm building, I want to have the media element be the single source of audio playback control, allowing the `wavesurfer.js` instance to be a single part of a group of visualizations rather than the main controller.

None that I am aware of

None that I am aware of

### Todos/Notes: 
I don't know if this has any unexpected consequences but it worked perfectly for me with this simple change when running `wavesurfer.js` with the `{backend: "MediaElement"}` configuration.

N/A

### DIFF
--- a/src/mediaelement.js
+++ b/src/mediaelement.js
@@ -119,7 +119,7 @@ WaveSurfer.util.extend(WaveSurfer.MediaElement, {
             my.fireEvent('play');
         });
 
-        media.addEventListener('ended', function () {
+        media.addEventListener('pause', function () {
             my.fireEvent('pause');
         });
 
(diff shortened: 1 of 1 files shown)


katspaugh:
This is a good idea!

However, won't this double fire the play and pause events when launched through wavesurfer.js (and not via the media element)? Maybe we need to remove the other `fireEvent` calls in the play and pause methods?

nlicitra:
Good thinking! I've removed the emit statements from the play and pause as the calls to `this.media.play()` and `this.media.pause()` should now cause the event to be fired automatically.

nlicitra:
@katspaugh just want to follow up on my changes. Thanks!

katspaugh:
Sorry, I was on vacation. Thank you, merged!

Issue #1084, opened 2017-05-10
Canvas draw and animation creating Memory Leaks


gitdisrupt:
### Wavesurfer.js version(s):
1.4.0

### Browser version(s):
Safari and Chrome

### Use behaviour needed to reproduce the issue:

All use cases ...

I was seeing a never ending increase in memory usage when instantiating a wavesurfer instance and playing back an audio file (tried both, as `MediaElement` and `WebAudio`) that is **NOT** released after calling `myWave.destroy()` to destroy the instance.

At first i suspected the memory leaks had something to do with WebAudio or even the Media element, but then to test this theory, i modified the wavesurfer code and disabled the render of the canvas.

As soon as i disabled the render of the canvas the memory leak was gone.

It seems there are references to the canvas elements, event listeners and animation that are not properly removed on destroy.

gitdisrupt:
OK EVERYONE!! I can't believe i didn't think to try this earlier ... but i believe i have made a **HUGE improvement to performance and memory usage!!**

VERY SIMPLY ...
For both the `drawBars` and `drawWave` methods, i wrapped their contents with a `requestAnimationFrame()` call, and this **DRAMATICALLY** reduced memory usage!!!!

**ALSO**, at the end of those calls, i cancel the request and `NULL` the property.

Here is what i did ... 

```
drawBars: function (peaks, channelIndex, start, end) {
        let requestAnimationFrame = window.requestAnimationFrame || window.webkitRequestAnimationFrame;
        requestAnimationFrame(() => {
            // contents of the drawBars method here
        });
        cancelRequestAnimFrame(requestAnimationFrame);
        requestAnimationFrame = null;
}
```
And then of course i did the same for the `drawWave` method as well.

I am betting there are other improvements that can be done as it relates to memory leaks, like nulling properties when they are no longer needed, but i feel like this alone makes for a great improvement.

Hey @entonbiba, can this update be applied to the latest version?

Cheers!!

entonbiba:
@gitdisrupt looks fine to me, can you create a pull request for it?

entonbiba:
@gitdisrupt I say let's create a util method for `requestAnimationFrame = window.requestAnimationFrame || window.webkitRequestAnimationFrame;` instead and just reference via `this.util.requestAF(functionhere(this));` method. This way it will automatically set it to null in the util method.

katspaugh:
This looks cool, @gitdisrupt, thank you very much!

[...]

PR #1106, opened 2017-05-27
Wrap drawing in requestAnimationFrame (multicanvas).

agamemnus:
Wrap drawing in requestAnimationFrame (multicanvas) per https://github.com/katspaugh/wavesurfer.js/issues/1084.

Blocked by https://github.com/katspaugh/wavesurfer.js/pull/1107.

### DIFF
--- a/src/drawer.multicanvas.js
+++ b/src/drawer.multicanvas.js
@@ -140,97 +140,105 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
     },
 
     drawBars: function (peaks, channelIndex, start, end) {
-        // Split channels
-        if (peaks[0] instanceof Array) {
-            var channels = peaks;
-            if (this.params.splitChannels) {
-                this.setHeight(channels.length * this.params.height * this.params.pixelRatio);
-                channels.forEach(function(channelPeaks, i) {
-                    this.drawBars(channelPeaks, i, start, end);
-                }, this);
-                return;
-            } else {
-                peaks = channels[0];
+        var requestAnimationFrame = window.requestAnimationFrame || window.webkitRequestAnimationFrame;
+        var my = this;
+        requestAnimationFrame(function() {
+            // Split channels
+            if (peaks[0] instanceof Array) {
+                var channels = peaks;
+                if (my.params.splitChannels) {
+                    my.setHeight(channels.length * my.params.height * my.params.pixelRatio);
+                    channels.forEach(function(channelPeaks, i) {
+                        my.drawBars(channelPeaks, i, start, end);
+                    });
+                    return;
+                } else {
+                    peaks = channels[0];
+                }
             }
-        }
 
-        // Bar wave draws the bottom only as a reflection of the top,
-        // so we don't need negative values
-        var hasMinVals = [].some.call(peaks, function (val) {return val < 0;});
-        // Skip every other value if there are negatives.
-        var peakIndexScale = 1;
-        if (hasMinVals) {
-            peakIndexScale = 2;
-        }
+            // Bar wave draws the bottom only as a reflection of the top,
+            // so we don't need negative values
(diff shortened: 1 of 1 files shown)


src/drawer.multicanvas.js
```diff
@@ -140,97 +140,105 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
...
-            var channels = peaks;
-            if (this.params.splitChannels) {
-                this.setHeight(channels.length * this.params.height * this.params.pixelRatio);
-                channels.forEach(function(channelPeaks, i) {
-                    this.drawBars(channelPeaks, i, start, end);
-                }, this);
-                return;
-            } else {
-                peaks = channels[0];
+        var requestAnimationFrame = window.requestAnimationFrame || window.webkitRequestAnimationFrame;
```
mspae: Can you add support for opera and IE? … Maybe this would be something to put in utils, since it is also used in MultiCanvas and potentially other places.

mspae: Maybe something like this:  https://gist.github.com/mrdoob/838785

agamemnus:
K. I'll change it in a bit.

agamemnus:
I don't know what's going on here.

src/drawer.multicanvas.js
```diff
@@ -231,15 +231,15 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
 
         // Always draw a median line
         this.fillRect(0, halfH + offsetY - this.halfPixel, this.width, this.halfPixel);
-    },
+    }),
 
-    drawLine: function (peaks, absmax, halfH, offsetY, start, end) {
+    drawLine: utils.frame(function(peaks, absmax, halfH, offsetY, start, end) {
```
katspaugh: `utils` is undefined. It has to be `WaveSurfer.util`. Please also restore the spaces after the function keyword.

agamemnus:
It's a function call though. Tough call..

agamemnus:
Also I wrapped drawLine by accident instead of the one above it (or so). My bad.

thijstriemstra:
anyone an idea what this error is about? https://travis-ci.org/katspaugh/wavesurfer.js/builds/237172952#L352

agamemnus:
Yeah... it's because I don't have all the requestAnimationFrame PRs in one place, so it doesn't know about util.frame...

thijstriemstra:
You'll probably have to bring in more code in this PR so the travis build doesn't fail, or find a way to make travis build happy.

agamemnus:
Once the util PR is accepted, this one will work.

agamemnus:
The one you pinged me about 20 min ago...! :P

https://github.com/katspaugh/wavesurfer.js/pull/1107

thijstriemstra:
cool, just for reference so others don't have to dig around for the PR nr. You should probably add to the PR description something like "blocked by PR #"

thijstriemstra:
Tests fail with:

```
TypeError: Illegal invocation

	    at Object.drawWave (dist/wavesurfer.min.js:4:9038)

	    at Object.t.Drawer.drawPeaks (dist/wavesurfer.min.js:4:21459)

	    at Object.t.empty (dist/wavesurfer.min.js:4:8462)

	    at Object.t.load (dist/wavesurfer.min.js:4:5975)

	    at Object.<anonymous> (spec/wavesurfer.spec.js:23:20)
```

so this typoe of method definition isn't allowed it seems..

agamemnus:
It's completely unclear what is going on here. The same changes worked in the Canvas PR.

katspaugh:
Weird. Can you reproduce this error locally?

agamemnus:
Found the issue. See: https://github.com/katspaugh/wavesurfer.js/pull/1120.

katspaugh:
The tests pass now, thanks!

PR #1107, opened 2017-05-27
requestAnimationFrame addition to utils.

agamemnus:
Sorry -- idk how to add to a PR.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -1,5 +1,13 @@
 /* Common utilities */
 WaveSurfer.util = {
+    requestAnimationFrame:
+        window.requestAnimationFrame ||
+        window.webkitRequestAnimationFrame ||
+        window.mozRequestAnimationFrame ||
+        window.oRequestAnimationFrame ||
+        window.msRequestAnimationFrame ||
+        function (callback, element) {window.setTimeout(callback, 1000 / 60);}
+    ,
     extend: function (dest) {
         var sources = Array.prototype.slice.call(arguments, 1);
         sources.forEach(function (source) {
--- a/src/util.js
+++ b/src/util.js
@@ -13,7 +13,7 @@ WaveSurfer.util = {
         return function () {
             var my = this, args = arguments;
             requestAnimationFrame(function () {
-                func.apply (my, args)
+                func.apply (my, args);
             });
         };
     },
--- a/src/util.js
+++ b/src/util.js
@@ -6,8 +6,17 @@ WaveSurfer.util = {
         window.mozRequestAnimationFrame ||
         window.oRequestAnimationFrame ||
         window.msRequestAnimationFrame ||
-        function (callback, element) {window.setTimeout(callback, 1000 / 60);}
+        function (callback, element) { setTimeout(callback, 1000 / 60); }
     ),
+
+    frame: function (func) {
+        return function () {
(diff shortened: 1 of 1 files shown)


src/util.js
```diff
@@ -1,5 +1,14 @@
 /* Common utilities */
 WaveSurfer.util = {
+    requestAnimationFrame: (
+        window.requestAnimationFrame ||
+        window.webkitRequestAnimationFrame ||
+        window.mozRequestAnimationFrame ||
+        window.oRequestAnimationFrame ||
+        window.msRequestAnimationFrame ||
+        function (callback, element) {window.setTimeout(callback, 1000 / 60);}
```
katspaugh: Thanks! Could you add spaces inside the anonymous function? And no need for `window`, `setTimeout` is a global also in Node.

src/util.js
```diff
@@ -6,9 +6,18 @@ WaveSurfer.util = {
...
         window.oRequestAnimationFrame ||
         window.msRequestAnimationFrame ||
-        function (callback, element) {window.setTimeout(callback, 1000 / 60);}
+        function (callback, element) { setTimeout(callback, 1000 / 60); }
     ),
 
+    frame: function (func) {
+        return function () {
+            var my = this, args = arguments;
+            requestAnimationFrame(function () {
```
katspaugh: I guess, it should be `WaveSurfer.util.requestAnimationFrame` defined above?

thijstriemstra:
ping @agamemnus

agamemnus:
Oh sorry. Yes. Fixed.

src/util.js
```diff
@@ -1,5 +1,23 @@
...
+        window.oRequestAnimationFrame ||
+        window.msRequestAnimationFrame ||
+        function (callback, element) { setTimeout(callback, 1000 / 60); }
+    ),
+
+    frame: function (func) {
+        return function () {
+            var my = this, args = arguments;
+            WaveSurfer.util.requestAnimationFrame(function () {
+                func.apply (my, args);
```
katspaugh: No need for the space after `apply`

agamemnus: No need for semicolons either...  :P

katspaugh:
Merged, thanks much!

agamemnus:
There is a problem that I just realized. Now this render is asynchronous, so anything using it actually needs a callback, etc.

agamemnus:
@thijstriemstra 

You can't do drawWave anymore, or whatever. you have to add a callback parameter, because the function will run asynchronously -- on the next animation frame.

There are several dozen lines of code that need to be reworked. It is all interwoven... check out drawPeaks, for example...

https://github.com/agamemnus/wavesurfer.js/blob/master/dist/wavesurfer.js

agamemnus:
(But then everything behind that needs callbacks too. So you have a callback pattern for drawPeaks (where I put the frame() call, and then callback pattern for drawBuffer. And then, although I didn't implement it, you need a callback for everything that calls drawBuffer.

agamemnus:
@katspaugh SOS

thijstriemstra:
A pr that reverts your changes might be necessary then.

agamemnus:
Or just add the callback methods. But maybe for next because this will
break outside code that expects an instant change.

PR #1110, opened 2017-05-31
Make the width/start/end generation functions public.

agamemnus:
I added four functions: `getComputedWidthStartEnd` (couldn't come up with a better name) and `getComputedWidth`/`getComputedStart`/`getComputedEnd`. This brings out code from drawBuffer.

Rationale: sometimes, a plugin might want to know what the width/start/end of a drawer is supposed to be before it is loaded...

I also reworded the comment to "Use the fill container" and put it inside the if... (I think that's what it means..)


src/wavesurfer.js
```diff
@@ -326,21 +326,30 @@ var WaveSurfer = {
...
+            var width = nominalWidth;
+            var start = this.drawer.getScrollX();
+            var end = Math.min(start + parentWidth, width);
         }
+        return { width: width, start: start, end: end };
+    },
+
+    getComputedWidth: function () { return this.getComputedWidthStartEnd().width; },
+    getComputedStart: function () { return this.getComputedWidthStartEnd().start; },
+    getComputedEnd: function () { return this.getComputedWidthStartEnd().end; },
```
katspaugh: I feel like these 5 extra functions inflate the API, while the hypothetical plugins could just use `getComputedWidthStartEnd`.

agamemnus:
Ok. That's probably right. It needs a better name though, but I couldn't
come up with one..

thijstriemstra:
StartEnd.. as in range?

agamemnus:
Hmm. Maybe getComputedWidthAndRange, then ..?

thijstriemstra:
Looks like most of this PR was already merged with https://github.com/katspaugh/wavesurfer.js/commit/e3fb34a5293a2107104a30f843bc3c3213a30040

thijstriemstra:
@agamemnus please rebase this PR and open a new one.

PR #1112, opened 2017-06-03
Add pointerEvents: none to left/right sides.

agamemnus:
Without this change, if the zIndex of the waves is higher than the zIndex of the regions, the waves will stand out better but block the resize controls.

(Figured out how to do two files at once.)

### DIFF
--- a/src/drawer.multicanvas.js
+++ b/src/drawer.multicanvas.js
@@ -23,7 +23,7 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
         this.progressWave = this.wrapper.appendChild(
             this.style(document.createElement('wave'), {
                 position: 'absolute',
-                zIndex: 2,
+                zIndex: 3,
                 left: 0,
                 top: 0,
                 bottom: 0,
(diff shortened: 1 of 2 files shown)


katspaugh:
Nice idea!

thijstriemstra:
@agamemnus can you also add this to the documentation?

agamemnus:
Ok, in a bit. By the way, what is the plan for merging changes into next?
Are you just going to do a bunch of monster merges?

agamemnus:
I'm not sure where I'd add it since it's not an option.

src/drawer.multicanvas.js
```diff
@@ -72,11 +73,12 @@ WaveSurfer.util.extend(WaveSurfer.Drawer.MultiCanvas, {
         entry.wave = this.wrapper.appendChild(
             this.style(document.createElement('canvas'), {
                 position: 'absolute',
-                zIndex: 1,
+                zIndex: 2,
```
glebm: This makes it so that this canvas is above the progress wave, hiding it. See #1136

Issue #1055, opened 2017-04-04
Broken drawing of raw peaks


adw1n:
When the resolution is small enough rendering pre-generated peaks (on the newest version of wavesurfer) is broken.
![broken1](https://cloud.githubusercontent.com/assets/8993001/24680868/e61935f4-1992-11e7-9cff-37d13ae917d0.png)
![broken2](https://cloud.githubusercontent.com/assets/8993001/24680887/eca7979e-1992-11e7-9830-1ddf7f73212e.png)
Once the screen estate is big enough, stuff is back to normal:
![broken3](https://cloud.githubusercontent.com/assets/8993001/24680917/03b8293a-1993-11e7-9670-d8396d4718e4.png)

On wavesurfer version 1.1.1 I had no problems whatsoever:
![working2](https://cloud.githubusercontent.com/assets/8993001/24680945/2d8be2ec-1993-11e7-839d-56cf098b1e62.png)

Browsers:
* FF 47
* Chromium 51.0.2704.79 

Source code of examples:
* broken (wavesurfer 1.3.7)  
http://public.adw1n.com/wavesurfer_issue/broken.html
* same code was working on version 1.1.1  
http://public.adw1n.com/wavesurfer_issue/working.html

Jamby93:
The problem raises whenever you provide more peaks than the canvasWidth.
The drawWave function of Canvas and drawLineToContext function of MultiCanvas has been broken since #924 by @mspae that has changed the rendering loop from looping the entire peaks array to looping only up to "end" or "canvasEnd" respectively which is both cases the canvas width. In this way, the last point drawn x-coordinate is canvasWidth * scale which is less than canvasWidth whenever you have more peaks than pixels.

mspae:
#924 is the v2 port of #909 – v2 hasn't been released yet – not that it matters :)

mspae:
Related? https://github.com/katspaugh/wavesurfer.js/issues/994

X-Raym:
I confirm this issue. V1.2.8 was fine and 1.3.0 is broken.

entonbiba:
closing, will continue on issue #1067 if pr #1072 doesn't fix it 100%

PR #1114, opened 2017-06-06
fix render waveform

entonbiba:
fixes render waveform with peaks.
#1067 #1055

preview: https://codepen.io/entonbiba/pen/PmpXqJ

waveform-error

to

waveform-test


agamemnus:
I don't think `width*end` is correct. I am getting tiny spikes at intermittent periods now.

agamemnus:
In fact, shouldn't it be `end = width  + 1`?

X-Raym:
@taylorvc can you confirm the  `end = width - 1` suggestion by @agamemnus works for you with long audio files ? :)

agamemnus:
@X-Raym: Don't you mean @entonbiba?..

taylorvc:
@X-Raym changing to `end = width -1` does not solve the issue for small or large files. Here is the waveform for a 4.5 hour long file:
image

Here is the waveform for a 20 min file:
image

X-Raym:
@taylorvc Thanks for testing :)
I think it is better to stay with the @entonbiba fix then 👍

agamemnus:
Oh, the spikes problem seemed to coincide with a formatting error (which is now fixed), so probably nothing -- I have been using PHP to output audio files based on a request to the server.

X-Raym:
@agamemnus good ! So, I think this PR is ready to be merge in its current state ? Or is there any problem left ?

agamemnus:
I'm just not sure why it is 9999. Seems arbitrary. I would like to try to figure out the root cause...

X-Raym:
@agamemnus 9999 was arbitrary, but now it is `end = width*end;` (it may seems also arbitrary, but surely more flexible than an absolute value). But finding the cause of the problem may be better indeed. Good luck in your investigation ! :)

entonbiba:
@X-Raym @agamemnus @taylorvc Is the spike thing happening on a specific browser? Not able to replicate it on my end.

agamemnus:
Not an issue anymore. Just a malformed file.

entonbiba:
@agamemnus sounds good

agamemnus:
Here is the problem: getPeaks uses sub-ranges and not width.

The correct could would then be:
````
            var numberOfChannels = this.backend.buffer ? this.backend.buffer.numberOfChannels :
                this.backend.preload ? this.backend.preload.numberOfChannels : 2;
            var subrangeLength = this.backend.peaks ? this.backend.peaks.length / numberOfChannels: width;
            start = 0;
            end = subrangeLength - 1;
            var peaks = this.backend.getPeaks(subrangeLength, start, end);
            this.drawer.drawPeaks(peaks, width, start, end);
````

Other issues/notes:
1) This adds a new parameter that we should pull from somewhere: `this.preload`, with a property of `numberOfChannels`.
2) I am not sure if the fireEvent should fire width or subrangeLength.
3) I suggest also that we rename getPeaks's `length` parameter into `subrangeLength` to differentiate the two.

agamemnus:
Updated code above.

agamemnus:
@entonbiba: updated code above.... thoughts?

X-Raym:
@entonbiba Hi ! any news about this bug ?

entonbiba:
@agamemnus @X-Raym I like the subrangeLength idea. The current quick fix `end = width*end;` should be merged first to at least fix the current issue of no waveform showing. We should open another pr for the subrange length as it would need more testing with larger and smaller file sizes.

I've been looking into `createPeriodicWave` webaudio method as well which looks interesting.

agamemnus:
It definitely hairy though.

X-Raym:
@entonbiba has this been fixed ?

entonbiba:
@X-Raym I opened another pr since this one had too many already merged prs.

X-Raym:
@entonbiba ok thx !

PR #1115, opened 2017-06-07
Update drawer.js -- make sure wrapper has the right parentNode.

agamemnus:
`this.container.removeChild(this.wrapper);` ->
`if (this.wrapper.parentNode == this.container) this.container.removeChild(this.wrapper);`

Useful just in case. For example, I have a function that detaches everything in order for childList/removedNodes mutationObservers to fire.


katspaugh:
Thanks!

PR #1116, opened 2017-06-07
Update drawer.js (same as #1115 except master instead of next)

agamemnus:
Topic. Same as https://github.com/katspaugh/wavesurfer.js/pull/1115.


katspaugh:
Thanks!

Issue #1082, opened 2017-05-07
Example for pre-rendered peaks doesn't actually show how to do it.


agamemnus:
https://wavesurfer-js.org/example/audio-element/

It just says: `wavesurfer.load('example/media/demo.mp3');`. It doesn't show how one would add in the extra parameters.

entonbiba:
@agamemnus those examples haven't been updated in a while. Sometime in the near future we will create more examples and link from the main page.
https://wavesurfer-js.org/examples/ 

For now you can check out this example I created for peak data.
http://codepen.io/entonbiba/pen/VPqvME

agamemnus:
Thanks.

agamemnus:
@entonbiba: I am also unsure how to save that peak data since the exportPCM doesn't work (https://github.com/katspaugh/wavesurfer.js/issues/1083). (I noticed you added a label -- thanks). I want to save the peaks with node.js as the audio files I am using are very big and I need the peaks to load much faster. Would appreciate any ideas.

davidchalifoux:
+1 on the need for updated documentation.

entonbiba:
@agamemnus you can use the following.
https://www.npmjs.com/package/audio-pcm-format

agamemnus:
This is confusing.

What am I supposed to use https://www.npmjs.com/package/audio-pcm-format for?

In another thread/issue, someone mentioned https://github.com/bbc/audiowaveform to get the peaks. I don't know what audio-pcm-format is supposed to do or what exportPCM is supposed to do. If I am using audiowaveform, why do I need the others? I am very lost here.

entonbiba:
@agamemnus that library is for nodejs didn't you ask for that? It will return the pcm json data from the serverside for faster rendering.

agamemnus:
None of this really works how it should.
1) I do not know how to properly install the audiowaveform program.
2) I tried `audio-pcm-format`, but setting a sampleRate of 44100 gave me an out-of-memory error.
I tried a sampleRate of 50 and used the wavesurfer code, but the result had one or two big peaks that made the whole thing really small when normalized.

agamemnus:
Okay. So I took the getPeaks code and I made it into a stream/chunk mechanism. I'm still improving on it and I'll post it later. Ideally, in the future, could just include wavesurfer.js and run getPeaks with a parameter...

entonbiba:
@agamemnus what kind of ram/cpu are you running it on?

[...]

PR #1117, opened 2017-06-07
Compressing and flipping min/max.

agamemnus:
1) Removed a bunch of empty lines for readability...
2) Min and max seemed to be flipped. Practically, it makes no difference.
3) Did `if (min < value) { min = value; } else if (max > value) { max = value; }`, since value cannot be greater than min and less than max at the same time.

### DIFF
--- a/src/webaudio.js
+++ b/src/webaudio.js
@@ -217,51 +217,29 @@ WaveSurfer.WebAudio = {
      */
     getPeaks: function (length, first, last) {
         if (this.peaks) { return this.peaks; }
-
         first = first || 0;
         last = last || length - 1;
-
         this.setLength(length);
-
         var sampleSize = this.buffer.length / length;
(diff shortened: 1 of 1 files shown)


katspaugh:
Took me some minutes to realize this PR doesn't do anything useful. Let's not introduce unnecessary stylistic changes.

agamemnus:
It will allow a better understanding of the code for more changes later
(see
https://github.com/katspaugh/wavesurfer.js/issues/1082#issuecomment-306976656)
and because min and max are flipped...

katspaugh:
I'm not sure why you think `max` and `min` are flipped (and should be reversed). Can you explain?

agamemnus:
I just looked at it again and it turns out that I got psyched out by the
code. They could not be reversed.

agamemnus:
Should*

PR #1120, opened 2017-06-11
requestAnimationFrame should be bound on window...

agamemnus:
Apparently...

https://stackoverflow.com/questions/10743596/why-are-certain-function-calls-termed-illegal-invocations-in-javascript


katspaugh:
Merged, thanks!

PR #1122, opened 2017-06-12
Slight position fixes.

agamemnus:
1] When a user clicks on the right-most pixel, it means the progress should be set at 100%. If they click the left-most, it should be set at 0%.

At present, it is set at 0% when the left-most pixel is clicked, but not 100% when the right-most pixel is clicked. Thus, I changed the code in `handleEvent` from:

`progress = ((clientX - bbox.left + this.wrapper.scrollLeft) / this.wrapper.scrollWidth) || 0;`
to:
`progress = ((clientX - bbox.left + this.wrapper.scrollLeft) / (this.wrapper.scrollWidth - 1)) || 0;`

(I also changed the code on the other side of that if statement.)

Thus, progress can be set by the user to exactly 100%.

2] In the `progress` function, we currently have `var pos = Math.round(progress * this.width) * minPxDelta;`. But this means that sometimes when the user clicks on a wave, the progress separator/"bar" will not be directly to the left of the user's cursor. It may sometimes be one pixel to the left of that. To solve this, I changed the code to `var pos = Math.ceil(progress * this.width) * minPxDelta;`.

(Sorry.)


thijstriemstra:
Perhaps you can place that value 1 in a variable and place a big comment above it explaining why it is subtracted.

agamemnus:
Idk, I can add a comment... but changing the 1 to a variable sounds like overkill...

thijstriemstra:
Well instead of defining it twice (and having to add a comment twice in separate places) I suggested a var.

thijstriemstra:
And if it's that important for correct rendering I don't think it's overkill at all. Me having to figure out why this magic 1 is used later on could be prevented if a proper comment is added.

agamemnus:
Added explanation and reworked code slightly.

src/drawer.js
```diff
@@ -48,18 +48,17 @@ WaveSurfer.Drawer = {
...
-            progress = ((clientX - bbox.left) * this.params.pixelRatio / nominalWidth) || 0;
-
-            if (progress > 1) {
-                progress = 1;
-            }
+            var numerator = (clientX - bbox.left) * this.params.pixelRatio;
+            var denominator = nominalWidth - 1;
         } else {
-            progress = ((clientX - bbox.left + this.wrapper.scrollLeft) / this.wrapper.scrollWidth) || 0;
+            var numerator = (clientX - bbox.left + this.wrapper.scrollLeft);
```
thijstriemstra: you're defining numerator and denominator twice..

agamemnus: "var" can only define once. How would you do it? I suspect an empty variable. I don't like that idea. Two extra superfluous lines of code.

agamemnus: Or, we can convert the if statement to a variable. Would that work?

agamemnus: (and use ternary)

agamemnus: (Can't figure out how to name this...)
````
var IDK = (!this.params.fillParent && nominalWidth < parentWidth);
var numerator   = IDK ? ((clientX - bbox.left) * this.params.pixelRatio) : (clientX - bbox.left + this.wrapper.scrollLeft);
var denominator = IDK ? (nominalWidth - 1)                               : (this.wrapper.scrollWidth - 1);
````

**Comment by agamemnus** [comment]
Ok, one more @entonbiba: shouldn't `this.width` theoretically equal `this.wrapper.scrollWidth`?

**Comment by agamemnus** [comment]
Never mind. It shouldn't...

File: src/drawer.js [hunk 2]
```diff
@@ -45,21 +45,22 @@ WaveSurfer.Drawer = {
...
-            if (progress > 1) {
-                progress = 1;
-            }
+        // If the entire container is not filled and further if the nominal width is less than the parent width...
+        if (!this.params.fillParent && visibleWidth < scrollWidth) {
+            var numerator = (clientX - bbox.left) * this.params.pixelRatio; // Apparently we need to scale this here...
+            var denominator = visibleWidth - 1;
         } else {
-            progress = ((clientX - bbox.left + this.wrapper.scrollLeft) / this.wrapper.scrollWidth) || 0;
+            var numerator = clientX - bbox.left + this.wrapper.scrollLeft;
```
katspaugh: Shouldn't this part be scaled with the pixel ratio, too?

agamemnus: No, because `visibleWidth` is the actual size and (just renamed) `scrollContainerWidth` is a virtual size which doesn't change if the pixelRatio goes up or down. I am not sure about the if statement though.

In general, params.fillParent is a bit difficult to understand.

thijstriemstra:
@agamemnus can you rebase a new pull request on the current master branch?

PR #1125, opened 2017-06-13
Make drawPeaks not run twice if peaks are pre-loaded.

agamemnus:
Topic...


agamemnus:
Similarly, (and I'm not sure about this one), but seekTo was firing twice if peaks were loaded. It should only fire once.

agamemnus:
Now including other dodgy fixes to make sure that, for both `WebAudio` and MediaElement`:

1) ... audio plays once seekTo is fired and the audio has loaded.
2) ... "preload" causes the audio data to start loading (for WebAudio).

X-Raym:
This sounds like a useful tricks. It should crease peak drawing performance a lot !

mspae:
I believe this is obsolete. If you disagree please reopen.

PR #1126, opened 2017-06-14
Add functionality for backgroundColor parameter

### LINKED ISSUE #1118: Background Color Parameter
My project calls for being able to export an image with a background color, which is only do-able if the drawer itself gets changed. I'm in the process of adding this – it's like, really straight-forward. Just wanted to know if this is something I should make a pull request for/was a feature that would be wanted by others.

disdyakis:
While an easy workaround for this would be to simply add a background color
to the parent container, the exportImage function/the canvas toDataURL function,
still only exports what is on the canvas. This change simply draws a rectangle
to fill the canvas if the parameter is set.

No breaking changes in external or internal API.

Resolves #1118


agamemnus:
This should really be a fillColor though.

disdyakis:
my reasoning was that if someone's working with the api, _backgroundColor_ is the most descriptive name for it (imo). and then, working with canvases, you have fill/fillRect/fillStyle, so _fillColor_ could be mildly confusing

agamemnus:
(Sorry, I meant fillStyle.)

Just a minor issue here though is that someone might expect this to set a `backgroundColor` on the wrapper.

Why not just set the backgroundColor and then when exporting, use a computed backgroundColor in the wrapper to draw the output (no need to redraw the image -- just do drawImage)? This would also play nicely with my submitted changes which creates a cutout of the wave and places a background color equal to the waveColor / progressColor.

disdyakis:
oh i thought you meant changing the parameter name lol. and i don't think i entirely get what you're suggesting. getImage just calls toDataURL on the canvas, so i don't see what drawImage would do, but i haven't worked with canvases a lot so i could just be missing something entirely

thijstriemstra:
can you leave out the changes in `/dist`?

agamemnus:
I did but then I was thinking about compatibility with my work and other issues.

So my thought is that before using toDataURL, have the getImage function apply the backgroundColor of the wrapper on the canvas. You can just do a regular fill but with globalCompositeOperation set to destination-over: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/globalCompositeOperation.

disdyakis:
oh i totally didn't know that was a thing. okay so i get how i'd do that but, can you explain how my changes affect your change you mentioned? i don't get what you mean by cutout. sorry if i'm being dense lol

agamemnus:
I'm inverting the wave so that the pixels where the wave becomes transparent and vice-versa. That allows me to put the inverse wave under the region plugin so that I could have different-colored wave segments, for example.

If you fill the canvas, that won't work anymore.

disdyakis:
ah gotcha. okay, i'll totally look into changing this branch with your suggestions, and then i'll do a new pull request without the ./dist changes

agamemnus:
All you need to do is just change that export function to implement a fix to your original problem, right... in fact, later on this could be extended further to take into account background images, etcetera -- although there is a library or two that can convert dom elements to a canvas, so it might be a bit much in that case.

disdyakis:
after implementing these changes the image returned from `exportImage()` was just a solid 
 rectangle of the background color for some reason. after some digging i found this in the canvas element spec for the function `context.putImageData(...)` which i guess might also apply to `getImageData(...)` at least that would explain why i'm only getting back a solid rectangle.

"The globalAlpha and globalCompositeOperation attributes, as well as the shadow attributes, are ignored for the purposes of this method call; pixels in the canvas are replaced wholesale, with no composition, alpha blending, no shadows, etc." -[ html canvas spec](https://html.spec.whatwg.org/multipage/canvas.html#pixel-manipulation)

so this seems like a dead end at that point. i'm just going to update this branch, and take out the changes to `./dist`

agamemnus:
Just create a temporary image first....

disdyakis:
i'm not exactly sure what you mean by that, but i've already spent more time on this relatively small change than i intended lol. this is perfectly fine for what my purposes for the change are, if it doesn't work with your changes and isn't going to be a useful addition to the code base i don't mind if my pull request is denied. or if it's accepted and you want to make it work with your changes, that'd be great too

thijstriemstra:
@disdyakis can you rebase/create a new pull request based on the current master branch?

PR #1129, opened 2017-06-15
setAliases and deepMerge util functions.

agamemnus:
The use cases (will be evident in subsequent pull requests -- later today):

`deepMerge` allows a deep merge of two objects and properties  (ignoring HTML elements).

This is useful in the case of a parameter having an object instead of a string/number/etc..

`setAlias` allows several source properties to match a target property via getters/setters. This is useful in the case where we have a `cursorColor` as a property but might also want to allow the user to alternatively set a cursor's `backgroundColor` (assuming the cursor was an element and its color was set with backgroundColor). Both properties would transform and be transformed by a hidden property; e.g.: `_backgroundColor`.


thijstriemstra:
> The use cases (will be evident in subsequent pull requests -- later today):

Please reference these pull requests so we can see what you're talking about.

agamemnus:
https://github.com/katspaugh/wavesurfer.js/pull/1130 ...

thijstriemstra:

PR #1130, opened 2017-06-15
create styleList and classList parameters + aliases

agamemnus:
classList/styleList holds any wavesurfer styles/classes potentially applicable to DOM sub-objects.

This will make it possible for example to change the cursor style (if/when it is converted to a DOM object per an incoming PR) or modify some of the wave styling.

Blocked by https://github.com/katspaugh/wavesurfer.js/pull/1129.


thijstriemstra:

PR #1131, opened 2017-06-15
drawBars and drawWave share some code... combine it + other optimization.

agamemnus:
drawBars and drawWave share some code on top. I refactored and then offloaded that into a new function called routeAndClear.

I also reworked/streamlined the rest of the code slightly. Will add the single canvas if OK.

Shouldn't affect anything else...


agamemnus:
OK, this needs to be put on hold because the request animation frame change just really is a big change.

thijstriemstra:
I'm also really reluctant to make these changes without any unit tests that verify it works as expected. Can you take a look at that?

agamemnus:
I made way too many changes for now, and partialRender doesn't work... although I am not sure how well it worked before. Let's wait for the smaller changes to go through.

mspae:
This is redundant since #1198 was merged, no? Sorry I didn't see this PR before doing the refactor, otherwise I would have reused some of the stuff. Can we close this now @agamemnus ?

agamemnus:
My code would probably need too much work by now. I'll stick with my version, though...

PR #1132, opened 2017-06-15
Rework structure and allow extra styles + extra classes + cursor alignment

agamemnus:
Blocked by https://github.com/katspaugh/wavesurfer.js/pull/1130. (I am also not sure how the merging will work here, but let's see....)

1) Allow styles/classes to be added for the wave/progressWave. E.G.:
````
   styleList: {
    progressWave: {borderBottom: "8px white solid"}, wave: {borderBottom: "8px lightGrey solid"},
   }
````
2) Cursor is now aligned as "right", "middle", or "left". Previous alignment was to the left by default but now it's in the middle point of where the audio file is. The cursor can be styled with styleList / classList, but previous methods also work.

2) Now the HTML is structured like this:
````
<div class="wrapper">
 <wave>
  <wave><canvas/></wave> <--- progress wave
  <wave><canvas/></wave> <--- regular wave
  <div></div> <--- progress bar
 <wave>
</div>
````
Versus the old way:
````
<div class="wrapper">
 <wave>
  <wave><canvas/></wave> <--- progress wave
  <canvas/> <--- regular wave
 <wave>
</div>
````

PR #1133, opened 2017-06-15
Create invertTransparency option to MultiCanvas.

agamemnus:
Blocked by https://github.com/katspaugh/wavesurfer.js/pull/1132. (I am again not sure how the merging will work here...)

Allows the user to specify an `invertTransparency` option. This basically cuts out the wave and puts a background color around the cutout, allowing the user to put stuff behind the wave (like regions) -- this could allow various effects, like different wave colors for different regions.


thijstriemstra:
I'm not such a fan of `invertTransparency`, it's not immediately clear from the name what this option does. Can you try and come up with a better name?

agamemnus:
Well, it causes the transparent pixels to become solid, and vice-versa. It takes the transparency and inverts it... not sure.

thijstriemstra:
ok but what part get's transparent (or the opposite). (invert)waveTransparency-something?

agamemnus:
The whole canvas transparency is inverted.. invertWave is shorter but it
also inverts bars so it's not accurate, and it's not clear what invert
means here.

Umm.. What about inverseOpacity or invertOpacity?

On Jun 25, 2017 12:38 PM, "Thijs Triemstra" <notifications@github.com>
wrote:

ok but what part get's transparent (or the opposite).
(invert)waveTransparency-something?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/katspaugh/wavesurfer.js/pull/1133#issuecomment-310913272>,
or mute the thread

agamemnus:
Thanks.... will rework with current code in the near future...

thijstriemstra:
> Thanks.... will rework with current code in the near future...

PR #1142, opened 2017-06-18
Join ends of peaks

agamemnus:
I'm not entirely sure that the peaks are joined at the ends. When I zoomed all the way in, it seemed that there were gaps...


agamemnus:
Actually, should be +1 for all. But I'm just guessing here. Not even sure about `var length = peaks.length / 2;`...

thijstriemstra:
> When I zoomed all the way in, it seemed that there were gaps...

Can you provide a screenshot?

agamemnus:
difference

katspaugh:
Merged, thanks!

Issue #1127, opened 2017-06-14
Add a method to redraw wave form


tf:
It would be nice if there was an API method to manually trigger redrawing the wave form. I mainly see two use cases:

**Responsiveness** Projects like `react-wavesurfer` already [use `drawBuffer`](https://github.com/mspae/react-wavesurfer/blob/master/src/react-wavesurfer.js#L76) to redraw the wave form once the container size has changed. As described in [this issue](https://github.com/mspae/react-wavesurfer/issues/43) this causes problems since `drawBuffer` appears [not to clear the canvas](https://codepen.io/tfischbach/pen/PjGQgX) before redrawing causing the waveform to look less crisp once the viewport is resized.

**Updating options** I'd also like an official way to change options like `waveColor` and `progressColor`
for an already initialized wavesurfer player. This is possible already now by calling:
```js
        wavesurfer.params.waveColor = nextProps.options.waveColor;
        wavesurfer.drawer.clearWave();
        wavesurfer.redrawBuffer();
```
It would be nice if there was an official way to achieve this. Something like a `redraw` method could be a step in this direction. Still, I am not sure if mutating `wavesurfer.params` should also be replaced by a method call along the lines of the existing `toggleInteraction`/`toggleScroll`, i.e. `setWaveColor`.

I'd be happty to sumit a PR if we can agree on a direction.

agamemnus:
My PR (here: https://github.com/katspaugh/wavesurfer.js/pull/1133) lets you put elements under the waveform, so you don't need to redraw it. FYI.

tf:
@agamemnus interesting idea. Thanks for the heads up. This would still only work if the wave is placed on a background that matches the color used for the "inverted mask" in the canvas, right? In my case there is a large background image behind the wave.

agamemnus:
Well, that is true. But you could also modify it to make the mask an image instead of a colored rectangle...

tf:
@agamemnus I'm sure there are cases where this would work, but in my case the background image is really dynamic and unrelated to the waveform. So I think the suggested API additions still make sense.

@katspaugh Looking a bit further, I found that calling `drawBuffer` originally worked. Adapting the minimal example that I [posted before](https://codepen.io/tfischbach/pen/PjGQgX) to [use 1.2.8](https://codepen.io/tfischbach/pen/dROxYa) makes it work fine.

[...]

PR #1144, opened 2017-06-20
Ensure canvas elements are cleared before redrawing

tf:
refs #1127

Before #909, calling `drawBuffer` to redraw the wave invoked
`drawer.drawPeaks` which in turn invoked `drawer.setWidth` which
always caused the canvas to be cleared as a side effect of
`drawer.updateSize`.

In #909 `setWidth` was changed to short circuit if the width did not
change. This now causes the waveform to be redrawn on top of the
previous rendition, making the edges of the wave appear less crisp.

This change makes `setWidth`/`setHeight` return a boolean to indicate
whether changes were needed. If not, `drawer.clearWave` is called
manually. So we make sure that the previous wave is always cleared,
but do not perform the possibly performance intensive task of clearing
the canvas twice if it already happened as a side effect of
`setWidth`.

### DIFF
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -88,7 +88,9 @@ WaveSurfer.Drawer = {
     },
 
     drawPeaks: function (peaks, length, start, end) {
-        this.setWidth(length);
+        if (!this.setWidth(length)) {
+            this.clearWave();
+        }
 
         this.params.barWidth ?
             this.drawBars(peaks, 0, start, end) :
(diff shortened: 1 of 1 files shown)


thijstriemstra:
@tf can you add some documentation to these methods, indicating they return a boolean value etc? Your PR description would be useful to use as documentation for that code.

katspaugh:
Thanks! I'll merge, I think it's pretty clear as is.

katspaugh:
Thanks! I'll merge, I think it's pretty clear as is.

PR #1146, opened 2017-06-21
Add refresh method to redraw wave after resize

### LINKED ISSUE #1127: Add a method to redraw wave form

tf:
When the width has changed, only calling `drawBuffer` is not enough since the width of the progress canvas has to be updated as well to account for the new total size.

Since getting the played percents requires interaction with the backend, we need to add a new method.

fixes #1127


thijstriemstra:
Thanks, it's very busy now, I'll take a look this weekend.

tf:
No hurry. I'll by away for two weeks starting tomorrow anyway. Happy to look into any change requests after.

src/wavesurfer.js
```diff
@@ -92,11 +92,6 @@ var WaveSurfer = {
         this.drawer = Object.create(WaveSurfer.Drawer[this.params.renderer]);
         this.drawer.init(this.container, this.params);
 
-        this.drawer.on('redraw', function () {
-            my.drawBuffer();
```
thijstriemstra: why remove this? Shouldn't this be `this.drawer.on('redraw', this.refresh);`?

tf: As noted in the commit message, I could not find any place where `redraw` is fired on drawer:

> The `redraw` event is never fired on the drawer (only the wavesurfer
> object itself). The code has been unused since [resize handling has
> been removed from the drawer][1].
>
> [1]: https://github.com/katspaugh/wavesurfer.js/commit/50a89e477a68792c522672a0c693e5456e7ab672#diff-0333441833ed71523c3686dac5804d80L145

mspae:
How does this relate to the responsive param? is it still necessary?

mspae:
Is merged into master since #1219

PR #1154, opened 2017-06-28
change name of regions methods onDrag() and onResize()

Jp-Rivera:
I was working on documentation for the regions plugin methods and I noticed the function onDrag() and onResize() were very useful. [This codepen](https://codepen.io/danmarcos/pen/wepdEO) shows an example of some uses. 

The naming seems to come based on the context in which the function is called. I think that it makes more sense to drop the "on" in the names of these function. This way it better represents the function's behavior when being called.


plugin/wavesurfer.regions.js
```diff
@@ -424,12 +424,12 @@ WaveSurfer.Region = {
 
                     // Drag
                     if (my.drag && drag) {
```
glebm: What was my.drag checking for previously?

glebm:
Definitely don't think this is worth breaking backwards compatibility on master. Perhaps this should be done only on the `next` branch.

plugin/wavesurfer.regions.js
Jp-Rivera: This is a problem I missed. 'my.drag' is checking if the current region being dragged. The names drag() and resize() will break the events bound to that region. Maybe something more like dragRegion() and resizeRegion()?

Jp-Rivera:
Yes, I think this should be on the next branch. Should I close out this PR and resubmit on the correct branch?

plugin/wavesurfer.regions.js
Jp-Rivera: sorry, this will not break the events to the region as I said before. It just won't define drag() or resize() as a function.

Issue #149, opened 2014-03-20
playback rate / sound pitch


pitrackster:
Changing playback rate is changing sound pitch.

As this behavior could be needed, sometimes we could need to preserve the original pitch (even with artifacts).

yoyomo:
It seems everyone has ignored the OP's problem. 

> Changing playback rate is changing sound pitch.

we don't want that. HTML5 <audio> tag can neatly do this without changing pitch.

sprycoder:
First of all Thanks to team for creating such a useful component for audio. 

I am using it for my project and its really easy to use. But I'm disappointed by the change in sound pitch after increasing playbackRate. I tried with waveSufer.setPlaybackRate, waveSurfer.backend.setPlaybackRate and other few options.

Is it still an open bug or I have to do it different way?

jeremylach:
I'd also be interested in this feature.

katspaugh:
I'll need to do some research on this.

jsphweid:
@katspaugh I've been looking into this and wanted to present this to you.

Someone [created a javascript version of the famous open-source soundtouch libraries](https://github.com/also/soundtouch-js) years ago and I used it to create [this plunker.](https://plnkr.co/edit/ROQyEtPgQGfgZdluRzQy).

I'm no expert on the Web Audio API and cannot figure out how to integrate it smoothly with your fantastic library. This creates a script processing node and I figured you could just connect it in to your context. Obviously it's more complicated than that; I don't know how to make it work with play and pause OR getting things to line up with different playback rates.

Does this look promising?

jsphweid:
@katspaugh Sorry, I can't stop playing around with this.

First off, I made somewhat of a working version with time-stretch preserve pitch: [current website](https://josephweidinger.com/browserTranscribe) & [GitHub repo](https://github.com/jsphweid/browserTranscribe).

But just to summarize, changes to WebAudio js file -> https://www.diffnow.com/?report=w2vuc 
It's kind of a mess, I know. But it works mostly well for my purposes. Assumes 2 channel audio (although I just fixed that on my site).

To be honest, I wonder if a lot of this is going to change when they stop supporting script processor nodes (https://webaudio.github.io/web-audio-api/#idl-def-ScriptProcessorNode) in favor of AudioWorklet nodes.

katspaugh:
@jsphweid if you are willing to finish this (and create a PR), I know @jeremylach is ready to sponsor this feature.

[...]

PR #1157, opened 2017-07-02
feature: preserving pitch when changing tempo (#149)

katspaugh:
A first stab at preserving the pitch when changing the playback rate of an audio clip.

Todo:
 * Currently, it makes the audio mono. Ideally, it should keep all channels.

**Update**:
I went with the solution that @jsphweid suggested. The wavesurfer.js code is unmodified. Soundtouch.js is attached as an external filter.

Basically, the full implementation can be seen here: https://github.com/katspaugh/wavesurfer.js/blob/76da8d6411a593e850bc6dc9c641ded1d5e31645/example/stretcher/app.js

Here's a live demo: http://wavesurfer-js.org/example/stretcher/


katspaugh:
Closing this to create the same PR into the `es5` branch.

infernocloud:
I've filed a bug issue reporting that this example only works in Chrome and audio fails to play in Firefox and Edge while changing the speed. #1307

Issue #1147, opened 2017-06-24
When providing preloaded peaks with MediaElement, the peaks are overwritten once the track loads entirely


matixmatix:
I'm using wavesurfer.js 2.0.0-beta01, `MediaElement` as a backend, and providing preloaded peaks.

When the track starts playing, the provided peaks are shown OK. But after the whole track gets downloaded, peaks generated by wavesurfer.js are shown over preloaded peaks. 

I replicated the issue here: https://codepen.io/newjurgen/pen/KqvYxO (the track is around 20MBs, it can take a little while to load).

How can I prevent this behaviour? 

Thanks for making wavesurfer.js.

matixmatix:
Seems it only happens with `MediaElement`. But I need this option because it enables partial content loading..

matixmatix:
I solved the issue by adding a parameter when initializing: `forceDecode: false`. I thought this parameter defaults to  `false`? Looking at this code, it truly looks so: https://github.com/katspaugh/wavesurfer.js/blob/master/src/wavesurfer.js. But somehow the `forceDecode` gets the value `true`, even when leaving it out when initializing. Not sure how this happens, looks like a bug.

thijstriemstra:
@entonbiba any idea?

entonbiba:
@thijstriemstra it's actually set to true in version 2 by default, I don't see anywhere that it gets the parameter from the initialize. I'll create a pr with a proposed fix.

![image](https://user-images.githubusercontent.com/5193884/27801706-70600bde-5fed-11e7-8d04-52be02f7430e.png)

mspae:
Fixed by the above PR once it's released.

PR #1159, opened 2017-07-03
fix for forceDecode - make default set to false

entonbiba:
should be false by default to prevent overlapping #1147 

ref example:
https://codepen.io/newjurgen/pen/KqvYxO 

image
image


mspae:
This only includes changes in the `dist/wavesurfer.js` file, not the original source. Otherwise I think it's fine.

entonbiba:
@mspae yep

PR #1163, opened 2017-07-10
Add option to prevent removing media element on destroy

tf:
When an existing media element is reused, it makes sense to keep the media element around even when destroying the player.


thijstriemstra:
@tf any news? can you update this branch?

tf:
@thijstriemstra Sorry for not getting back earlier. Rebased the PR.

PR #1164, opened 2017-07-11
Fix variable redeclaration in the timeline plugin

glebm:
Fixes a variable redeclaration in the timeline plugin, allowing it compile in TypeScript without warnings.


glebm:
@katspaugh Changed the target branch to es5 as `next` has been merged into `master`.

Issue #1167, opened 2017-07-14
Big file cause tab crash in chrome


ale-grosselle:
Hi all!
As many library wavesurfer uses xhr to get information.

When try to download a big audio file (I can't give you the exact size because depends on your ram) chrome tab crashes if exceed memory.
Safari and Firefox work fine.

So, I suggest to creare a parameter to limit the wave generation if file is over <parameter> value.
I prepare a PR with this param

#1168

ale-grosselle:
#1168

entonbiba:
@ale-grosselle How big is the actual file and are you using the MultiCanvas render? Can you post some sample code on how you are loading the audio?

ale-grosselle:
My test:
- Use this file: https://qaxthron-cdn.thron.com/static/BGRPNK_170713-lazanzara-128_LHZG9Z.mp3
- In my Virtual Machine, window7, ram  2GB, chrome 59
` wavesurfer.init({
        container: document.querySelector('#waveform'),
        waveColor: '#A8DBA8',
        progressColor: '#3B8686',
        backend: 'MediaElement',
	barWidth: 3,
	waveStyle: 'soundWave'
    });
    // Load audio from URL
    wavesurfer.load('https://qaxthron-cdn.thron.com/static/BGRPNK_170713-lazanzara-128_LHZG9Z.mp3');`  

Pay attention where you will test, for example, in my 8gb mac book air, in chrome all work fine!
In a pc with 1 or 2 gb chrome crash (only chrome).

noakcess:
Figured out the issue!
You cannot stream from AWS S3 directly.

Created a localhost server via nodejs.
Server downloads the file from S3 and stores locally.
Servers back the file to the browser.
No Crashes!  1.5 hour file loaded without issue.

Hopefully this helps figure out why streaming from AWS S3 causes this issue.

PR #1168, opened 2017-07-14
Creates a param to limit wave creation (Avoid crash chrome)

ale-grosselle:
Fix for #1167


katspaugh:
Thanks for your PR, but this seems like a very specific browser workaround rather than a bugfix. I suggest that the application code should do this kind of size checks and notify the user accordingly.

PR #1169, opened 2017-07-17
Feature: soundtouch.js filter to preserve pitch when changing tempo

katspaugh:
I went with the solution that @jsphweid suggested in #149. However, the wavesurfer.js code is unmodified. Soundtouch.js is connected as an external filter.

Here's a live demo: http://wavesurfer-js.org/example/stretcher/


thijstriemstra:
is soundtouch available on npm? I'd rather not include it in the repo as it gets outdated quickly.

katspaugh:
It was last updated 5 years ago 😄

katspaugh:
@jeremylach failed to sponsor the bounty, so I'm closing this.

wcallag3:
I utilized the code in #1157 in app.js from line 20 - 70 to preserve pitch when changing playback rate, however I still get the issue where the play cursor does not update correctly. I'm currently using Wavesurfer 1.1.1 (I know it's super old, but I can't change it right now). I've also noticed that this only works in Chrome and not Firefox. Any insight would be much appreciated.

katspaugh:
@wcallag3 so is your issue with #1157 or with this PR?

katspaugh:
GitHub did something weird when I tried to change the base of the PR to the current master. I'll need to update the code for the version 2.

tyler-g:
@katspaugh I'm not exactly sure how to implement this without copying code from the demo example: http://wavesurfer-js.org/example/stretcher/

Can this be available as a plugin and added to the plugins docs?

katspaugh:
@tyler-g good idea! Will do. Btw the stretcher example was merged in #1214.

tyler-g:
Great!  Thanks

Issue #1148, opened 2017-06-26
Get the current region


dimidd:
Hi, is there an easy way to get the current playing region, i.e. the smallest region that contains the current time? I've thought iterating through `wavesurfer.regions.list` and for each region compare the start and end to the current time. Is there a better way?

agamemnus:
Seems like the way to do it.

thijstriemstra:
Fixed (and/or merged) with #1148.

PR #1180, opened 2017-07-31
Get the current region (#1148)

dimidd:
Get the current region, i.e. the smallest region that contains the current time.
If several such regions exist, we take the first.
Return null if none exist.

Implements https://github.com/katspaugh/wavesurfer.js/issues/1148

Signed-off-by: Dimid Duchovny <dimidd@gmail.com>


mspae:
Can you add a unit test for the function?

dimidd:
Sure, where should the tests be? In the regions plugin itself or in `spec`?


2017-08-18 15:54 GMT+03:00 Martin Spencer <notifications@github.com>:

mspae:
I just saw we don't have any tests yet. Either you create a `regions.spec.js` in `/spec` and add your tests there. Or leave it out and the person that adds all the other tests for the regions plugin (at some point) also adds it.

Issue #1181, opened 2017-08-01
Generated waveform with audiowaveform is half-complete


BriceN:
The command I use :
`audiowaveform -i myfile.mp3 -b 16 -o myfile.json`

What i get in my application : 
![capture2](https://user-images.githubusercontent.com/8823428/28817845-beb2b1d8-76a9-11e7-86dd-d699e1a8e33a.PNG)
(missing part in yellow, the blue peaks match the music perfectly )

Snippet of my json file : 
`{"sample_rate":44100,"samples_per_pixel":4410,"bits":16,"length":2318,"data":[-8,99,-2,91,7,167,-2512,2011,-2447,2009,-2475,1922,-2338,1626,-2356,1624,-2556,1697,-3050,1899,-3175,2268,-2978,2105,-2925,.....`

What is the problem here ? Not enough peaks data are generated ?

Also i use :
` normalize: true` as wavesurfer param
And I filter the peaks data like that :
`                response.data.data.filter(function(value) {
                    return value >= 0;
                })`

entonbiba:
this will be fixed with pr #1182

GitHubRakeshSharma:
I had the same issue. I found that the drawLineToContext method uses the canvas size in px as top limit in the drawing loop. I just changed it to be the local length parameter which is half the length of the converted array of peaks (meaning the one where negative values have been added alternatingly).

mspae:
Should be fixed now. Please check if it works for you now. @GitHubRakeshSharma @BriceN 

To install a npm package directly from github, do:
`npm install katspaugh/wavesurfer.js#master`

GitHubRakeshSharma:
I did import the new version from the masterbranch... BUT: now the dist folder is missing. Can you generate one or could you tell me how to efficiently genrate one myself?

mspae:
@GitHubRakeshSharma My bad, use this:

```
git clone git@github.com:katspaugh/wavesurfer.js.git && cd wavesurfer.js && npm i && npm run build && npm link && cd ../ && npm link wavesurfer.js
```

it …
* clones wavesurfer.js into the directory `wavesurfer.js`
* installs the dependencies in that project
* builds it (the `dist` directory)
* [links it](https://docs.npmjs.com/cli/link) so wavesurfer is resolved to that copy

to update you just `git pull` inside the `wavesurfer.js` directory.

This is just a quick-fix until the next beta release comes out.

GitHubRakeshSharma:
[...]

Issue #1175, opened 2017-07-22
Only part of the waveform is rendered when the browser window isn't maximized


FraserThompson:
### Wavesurfer.js version(s):

Every version since 1.3.0 up until 2.0.0 beta. The behaviour isn't there in 1.2.8.

### Browser version(s):

Chrome 60, Edge, Internet Explorer 11, Firefox

### Code needed to reproduce the issue:
JS:
```
    this.wavesurfer = WaveSurfer.create({
      container: this.waveform,
      waveColor: color,
      height: height,
      hideScrollbar: true,
      normalize: true,
      pixelRatio: "1",
      forceDecode: false,
      backend: "MediaElement",
      progressColor: 'orange',
      barWidth: '2'
    });
```

CSS:
```
.waveform {
	opacity: 1;
	transition: opacity 250ms ease-in-out;
	width: 100%;
	position: relative;
}

#waveform_wrapper {
	height: 60px;
	opacity: 1;
	width: 100%
}
```

HTML:
```
<div id="waveform_wrapper" class="player row" data-artist="{{include.artist}}"">
    <div class="col-xs-12" style="padding-top: 5px;">
        <div class="length-wrapper" style="left: 45px;">
            <span class="currentTime"></span>
        </div>
        <div class="waveform" id="waveform_{{include.machine_name}}"></div>
        <div class="length-wrapper" style="right: 0px;">
            <span class="length"></span>
        </div>
    </div>
</div>
```

### Use behaviour needed to reproduce the issue:

Loading an MP3 and Waveform into Wavesurfer when the browser window isn't maximised (or the developer console is open) will cause it to only render part of the waveform, like this:

![image](https://user-images.githubusercontent.com/6093858/28487190-c80e1f78-6edf-11e7-9f2e-5ecdf605db57.png)

If you maximise the window and then load the waveform it renders fine.

When I inspect I can see the canvas is there and is the correct width, it's just only being partially displayed.

entonbiba:
@FraserThompson  this has already been fixed with pr #1114 but it hasn't been merged yet.

FraserThompson:
Thanks. Is there a workaround I can use until that's merged? What's the actual cause?

matixmatix:
@entonbiba looking forward to that being merged, is there an estimation when it will happen?

entonbiba:
@matixmatix not sure, hopefully soon. I added the fix to version 1.4.0 and attached it here:
[...]

PR #1182, opened 2017-08-01
Fixes cut off waveform rendering

entonbiba:
Opening a new pr since old one had too many merges committed #1114 . This will have to be merged in order to fix the issue below.

fixes render waveform with peaks.
#1067 #1055 #1175 #1181

preview: https://codepen.io/entonbiba/pen/PmpXqJ

waveform-error

to

waveform-test


mspae:
Check this: https://codepen.io/mspae/pen/Jyyxjx

the first waveform does not show the scrollbar (only upon clicking play does it render it)

I haven't tested scrolling and resizing in this scenario

entonbiba:
@mspae it seems it needs a click for it to render the scrollbar for the first one, I'll look into it

mspae:
I've got a fix. Just preparing a PR. Wow. that was arduous :p

mspae:
I'll close this. since #1198 was merged.

X-Raym:
Oh cool ! So this problem is fixed now ? :)

mspae:
@X-Raym should be, see #1181

awenger63:
Hello, I just begining to test your nice wavesurfer (v2.0.0-beta01) with peaks generated by audiowaveform in 16bits. 
When the wave is rendered in load function with peaks this is make 2 time => the fisrt time, just the begining part (like issue #1181 ) is visible and in second time (1 or 2 seconds later) the rest of the wave will be display (I thniks it's a redraw call).
I look change code of this closed issue, and i see it's not the same on actual version the 2.0.0 (line 3965 of wavesurfer.js). I change it to test in a local version to have the same (width*end) and the wave is correctly display in one time with the peaks ! BUT with this same code if i call load without peaks, only with song file mp3 => there is no wave :-(
Do you know if it's possible to have a solution applied who work whith and without peaks ?
Thanks a lot

Issue #1092, opened 2017-05-14
Releasing 2.0.0


mspae:
There has been an extensive refactor of wavesurfer within the last half year which resulted in a version which includes a lot of new features (namely a dynamic plugin system) but also some breaking changes. We have decided to release it as a version two. Changes to either branch will be ported to each other (especially bug fixes of course).

This issue is intended as a place for general discussion and feedback regarding the v2 release.

*Edit 1:* Added change I forgot to mention: MultiCanvas is the default and only inbuilt renderer

The version is _mostly_ stable and functional. Due to the nature of the refactor (an integrated plugin system for adding, initialising and destroying of plugin instances) all plugins are somehow affected. Some plugins are still buggy when used with the dynamic plugin API. Conventional use cases (i.e. cases where you just want to create an instance with some plugins – no dynamic adding or removing – should work just fine) (Below in the plugin API examples, this means 1. should already work, 2. and 3. don't yet always work in all situations and for all plugins)

### Installing the release

To get the current beta release (at the time of writing) do `npm install wavesurfer.js@2.0.0-beta01`

You can find the beta release(s) on the [releases page](https://github.com/katspaugh/wavesurfer.js/releases) or by running `npm show wavesurfer.js | grep beta`

### Developing

To work on the new branch checkout the `next` branch, do `npm install` and then `npm start` – The library directory is exposed under `http://localhost:8080`

Generate the documention by running `npm run doc` (see below) – Tests can be run by calling `npm run test`

**There are still lots of things which need doing.** We welcome and are thankful for any help we can get. Please see right at the bottom of this post how you can help us.

### Changes

(See a [full list of all merged PRs for v2](https://github.com/katspaugh/wavesurfer.js/pulls?q=is%3Apr+is%3Aclosed+label%3Av2))

- **ES6 Syntax:** Refactored all code to use the new ES6/ES2015 syntax (especially arrow functions, classes, spread operators etc.)
- **New build system:** Webpack is being used as a build system to compile the code. It can be imported like this:
```js
// EITHER - accessing modules with <script> tags
var WaveSurfer = window.WaveSurfer;
var TimelinePlugin = window.WaveSurfer.timeline;
var MinimapPlugin = window.WaveSurfer.minimap;

// OR - importing as es6 module
import WaveSurfer from 'wavesurfer.js';
import TimelinePlugin from 'wavesurfer.js/dist/plugin/wavesurfer.timeline.min.js';
[...]

PR #1191, opened 2017-08-18
Ports from es5

mspae:
Adds features and fixes that were added to the es5 branch (previously master) to the new master (previously next).

See #1092


katspaugh:
Thank you so much!

PR #1192, opened 2017-08-19
chore: eslint autofix + bump the version

katspaugh:
I've run `eslint --fix` and updated the version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "wavesurfer.js",
-  "version": "2.0.0-beta01",
+  "version": "2.0.0-beta02",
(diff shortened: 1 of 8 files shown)


mspae:
Weird. eslint complains because of the changes eslint --fix introduced.

I've been meaning to ask: What would you say to the idea of introducing a code formatter.

I've recently updated [react-wavesurfer](https://github.com/mspae/react-wavesurfer) to use prettier, which is an opinionated js code formatter. I've configured eslint and prettier to follow/check the same style and I've added a pre-commit hook to ensure code is formatted before it enters version control.  – that works quite nicely: I propose we do the same for wavesurfer.js

katspaugh:
Good idea! Closing this then.

PR #1199, opened 2017-08-27
Feature: Reset zoom/unzoom

mspae:
* Adds the ability to call `wavesurfer.zoom` without a parameter (or with a falsey parameter) to reset zoom state
* Adds test for the properties set in the zoom method. (Doesn't currently test the right functions are being called)

None.

None.

### DIFF
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -910,16 +910,22 @@ export default class WaveSurfer extends util.Observer {
 
     /**
      * Horizontally zooms the waveform in and out. It also changes the parameter
-     * `minPxPerSec` and enables the `scrollParent` option.
+     * `minPxPerSec` and enables the `scrollParent` option. Calling the function
+     * with a falsey parameter will reset the zoom state.
      *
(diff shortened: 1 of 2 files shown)


src/wavesurfer.js
```diff
@@ -910,16 +910,22 @@ export default class WaveSurfer extends util.Observer {
 
     /**
      * Horizontally zooms the waveform in and out. It also changes the parameter
-     * `minPxPerSec` and enables the `scrollParent` option.
+     * `minPxPerSec` and enables the `scrollParent` option. Calling the function
+     * with a falsey parameter will reset the zoom state.
      *
-     * @param {number} pxPerSec Number of horizontal pixels per second of audio
+     * @param {number?} pxPerSec Number of horizontal pixels per second of
```
katspaugh: By the way, a nullable type is _prefixed_ with a `?`, I think (according to http://usejsdoc.org/tags-type.html).

mspae: aah true. I'll correct and merge.

PR #1210, opened 2017-09-18
Visual regression testing

mspae:
1. Visual regression tests for wavesurfer core library using [gemini](https://github.com/gemini-testing/gemini) (See below for an explanation of what this is)
2. Added `duration` parameter to `ws.load` function in order to correctly render according to `minPxPerSec` when using `WebAudio` and peaks. (Wavesurfer doesn't know how long the audio is until it decodes – This allows us to explicitly pass the duration until the audio is decoded)
3. Both backends fire `waveform-ready` (when the waveform is visible) and `ready` (when the audio is decoded/playable)
4. Added `ws.hasWave` flag, analogous to the `ws.isReady` flag
5. Updated the regions and timeline plugin to initialise on the `waveform-ready` event and `hasWave` flag.
6. Fixed minor bugs that I discovered in the process in order to create correct reference screenshots:
    - Responsive
      - The responsive parameter didn't accept `0` because it is a falsey value. This meant that a debounce timeout of `0` was not possible.
      - Resizing function (when `responsive` is set) should only be called if `scrollParent` is not true
      - Progress was incorrectly scaled when the waveform was resized. This was fixed by using the `redraw` event
    - Other
      - `Redraw` was triggered at the end of `drawBuffer` – AFAIK this is wrong since in `createDrawer` we listen for this event and call `drawBuffer` when it is fired. (`redraw` is a "command" event, not a "notification" event) – If such an event is required we should add a new event `redrew` (or similiar)

- There might be breakage due to the change of events. The events should behave like this:
   - `waveform-ready` a waveform has been drawn with a certain set of peaks. This is not called when scrolling or zooming, but is called when new peaks arrive, for example when the audio was decoded.
  - `ready` the audio has been decoded and is playable (webaudio) or the audio is playable because the media element has fired the [`canplay`](https://developer.mozilla.org/en-US/docs/Web/Events/canplay) event

See above.

### Todos/Notes:
- The testing workflow is documented in the readme, but I've also documented it at the bottom of this text.
- Creating reference screenshots and testing sometimes fails for some reason (probably to do incomplete audio decoding etc.) – rerunning the tests fixes the issue. But this should be fixed before ...
- We can write [browserstack](https://www.browserstack.com/) to incorporate this into our CI system
- We should also write (visual) tests for the plugins

([source](https://css-tricks.com/visual-regression-testing-with-phantomcss/))

#### Visual regression testing workflow

To set up visual regression testing do:
```bash
npm install -g gemini
npm install -g selenium-standalone
selenium-standalone install
```
Then to update reference images:
```bash
# start the selenium server (needs to run in a seperate tab)
selenium-standalone start

# start the webpack server (needs to run in a seperate tab)
npm start

# start the gemini ui
gemini-gui
```


mspae:
PS: I know this is larger than PRs should be.

cat gif

thijstriemstra:
@mspae this branch has conflicts.

katspaugh:
Wow, this is great!

mspae:
Okay, so the status on this is:

I will split it up into smaller PRs (some of which are already open/merged) – also I will be a little more conservative with the whole "change rendering events etc." shenanigans.

Already open or merged part-PRs:
- responsive bugs
- redraw/redrew

TBD PRs:
- Duration param for WebAudio
- Tests

mspae:
@katspaugh @thijstriemstra So once #1239 is merged I'll open a PR purely for the tests.

thijstriemstra:
@mspae can you update this PR so it merges cleanly against the master branch?

thijstriemstra:
No feedback.

This pull request is heavily out of date and should be merged forward. Please do that or open a new pr.

PR #1214, opened 2017-09-29
feature: soundtouch.js filter to preserve pitch when changing tempo (…

katspaugh:
The same as #1157 but rebased on master.


mspae:
Should be able to merge when #1233 is merged. (or cherry-pick the commit before)

infernocloud:
Where did this soundtouch.js come from? Can someone link the official repo for this 3rd party script?

katspaugh:
@infernocloud the repo is here: https://github.com/also/soundtouch-js/blob/master/src/js/soundtouch.js

PR #1220, opened 2017-10-04
Fix: changed the fired redraw event after drawBuffer to redrew

mspae:

Changed the fired `redraw` event at the end of `drawBuffer` to `redrew` otherwise there would be a hypothetical loop since `drawBuffer` is called when `redraw` is fired:
```javascript
// in createDrawer:
this.drawer.on('redraw', () => {
  this.drawBuffer();
  this.drawer.progress(this.backend.getPlayedPercents());
});
```

* `redraw` is fired in order to make the waveform rerender
* `redrew` is fired when the waveform was rerendered.

* Hypothetically this could break code which relies on the "second" redraw event (= at the end of drawBuffer).

* See above. AFAIK none of the internal code actually broke.

### Todos/Notes:
* Contains parts of #1210

/

### DIFF
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -890,7 +890,7 @@ export default class WaveSurfer extends util.Observer {
      * Get the correct peaks for current wave viewport and render wave
      *
      * @private
-     * @emits WaveSurfer#redraw
+     * @emits WaveSurfer#redrew
      */
     drawBuffer() {
         const nominalWidth = Math.round(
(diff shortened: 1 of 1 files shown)


thijstriemstra:
Is this `redrew` used anywhere else?

mspae:
It's a new event I created to be able to differentiate between:

- `The waveform should be redrawn! (Command)` → `redraw`
- `The waveform was just redrawn. (Informational)` → `redrew`

As such it isǹ't being used anywhere as yet. `Redraw` is being used extensively but as I mentioned above I think breakage is unlikely.

mspae:
@thijstriemstra @katspaugh How should we proceed with this? I'm afraid the backward-compatible solution I proposed in #1244 won't work here.

thijstriemstra:
@mspae good question, I'm also running into this issue now.

katspaugh:
I can't think of a use-case which would break either.
Could we call the new event `redrawn`, though? As in a past participle.

thijstriemstra:
no feedback, closing.

Issue #771, opened 2016-07-08
Uncaught (in promise) DOMException: The play() request was interrupted by a call to pause().


Fiorello:
Hi, i'm trying to use wavesurfer.js with MediaElement backend, and external audio tag.
When i seek into the wave while playing i get this error:

`Uncaught (in promise) DOMException: The play() request was interrupted by a call to pause().`

Maybe it depends by the size of the audio file, 16 minutes about 4 MB.
When i use loop function, with regions plugins, the browser tab freeze for a while, generating an huge amount of this error.

I've tried to use your "audio-element"and preload peaks with my audio file, and the error is the same.
![screen shot 2016-07-08 at 17 18 44](https://cloud.githubusercontent.com/assets/3278942/16692289/72d1a7fc-4530-11e6-9361-67c3e4699a04.jpg)

Can you help me, please?

mkadutskyi:
I have something the same. When I have this option backend: 'MediaElement', I can't seek. It just paused. And throughs the same error

mspae:
I can confirm that this also happens in the annotations example, which uses the MediaElement backend to get around CORS restriction for the remote audio file.

entonbiba:
@Fiorello @mkadutskyi @mspae may you check with the latest version, the following fix was applied #977 which might be related.

mspae:
@entonbiba Yes awesome! – I found https://github.com/katspaugh/wavesurfer.js/issues/994 while checking.

vvillalo:
How was your xhr configuration in the waveconfig?

PR #1229, opened 2017-10-13
Play/pause now properly return Promises

ffxsam:

This should fix errors such as "Uncaught (in promise) DOMException: The play() request was interrupted by a new load request", as developers can now use the returned Promise to wait for subsequent media actions.

N/A - though the documentation should be updated to let developers know that `play`, `pause`, and `playPause` now return a Promise as per the HTML5 audio spec (this PR should also work for WebAudio API as well, but please verify!)

Not sure. Shouldn't be any.

### Todos/Notes:

Might be a good idea to review WebAudio and HTMLAudio specs and check which methods return Promises, and return those from Wavesurfer's methods.

Issue #771 could be related.


mspae:
Great idea!

The [media element play method](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/play) returns a promise, the [pause method](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/pause) is currently not standardised and therefore we cannot rely on it returning a promise.

In general I think it is a good idea to return the play promise as you said. I think adjusting the JSdoc comments above the functions are still missing. (Maybe also with an `@example` block with an example of how to use the promise)

Regarding the implementation for WebAudio the specs should be useful: [AudioContext.resume](https://webaudio.github.io/web-audio-api/#dom-baseaudiocontext-resume) and [AudioBufferSourceNode.start](https://webaudio.github.io/web-audio-api/#dom-audiobuffersourcenode-start)

ffxsam:
That's fine if `pause` doesn't return a promise. In the code I've written, it should just return `undefined`.

I can update the JSDoc blocks when I get some spare time again.

Issue #1228, opened 2017-10-13
Error raised when trying to seekTo on Firefox


BriceN:
Hello,
When I try to click on any part of the waveform on firefox, this error is raised : 

`TypeError: Value being assigned to HTMLMediaElement.currentTime is not a finite floating-point value.`

Looking at the sources, it seems that in this function : 

`            this.drawer.on("click", function(e, progress) {
                setTimeout(function() {
                    console.log(progress);
                    my.seekTo(progress);
                }, 0);
            });`

the value of progress is `Infinity`

At this point, I don't know what to do to solve this, any ideas ?

mspae:
Thank you for reporting. We'll look into it as soon as possible.

BriceN:
Note that I'm using the regions plugin

mspae:
Duplicate of https://github.com/katspaugh/wavesurfer.js/issues/745

mspae:
Nevertheless this should be fixed. Which browser version are you using? Can you post some code?

mspae:
Let's discuss in the original issue.

mspae:
Apparently not an (exact) duplicate. Reopening

mspae:
So there will be a better error message when this happens, but we still haven't found out why this appears.

awenger63:
Hi, I have same issue but on Chrome (Version 63.0.3239.108), not on fireFox (57.0.2) ! with v2.0.1

`Uncaught TypeError: Failed to set the 'currentTime' property on 'HTMLMediaElement': The provided double value is non-finite.
    at MediaElement.seekTo (wavesurfer.js?ln=js:2090)
    at WaveSurfer.seekTo (wavesurfer.js?ln=js:3949)
    at wavesurfer.js?ln=js:3683`

It cause by this.backend.getDuration() is Infinity
And it appear not each time.... load with and without peak (json), not for all song, not for same song. it's look like same process to load have different result depend of the moment. 
I suggest this depend of webBrowser bufferread, I see this [https://www.chromestatus.com/feature/5671401352593408](url), FireFox in development!
So, sometime buffer is complete and Duration can be define, sometime no !
I search to change webBrowser buffer size but without success 👎 
I don't know if this explain the issues and if this impact getDuration() in the futur.
If someone have any idea ?

MileanCo:
Im also experiencing this issue in Chrome...

MileanCo:
Also experiencing these issues in Chrome..
[...]

PR #1234, opened 2017-10-28
Fix #745: Check seekTo progress parameter and return understandable error

### LINKED ISSUE #745: Failed to set the 'currentTime' property on 'HTMLMediaElement'
big wav file.1:20:32.wrong:
wavesurfer.js:1257 Uncaught TypeError: Failed to set the 'currentTime' property on 'HTMLMediaElement': The provided double value is non-finite.

mspae:
- Checks whether the progress parameter passed to `wavesurfer.seekTo` is
  - a number
  - not NaN
  - between 0 and 1

/

/

### Todos/Notes:
Also includes the build fixing commit from #1233 in order to be able build.

- "Fixes" #745 by providing a readable error message.
- "Fixes" #1228

### DIFF
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -739,6 +739,16 @@ export default class WaveSurfer extends util.Observer {
      * wavesurfer.seekTo(0.5);
      */
     seekTo(progress) {
+        if (
+            typeof progress !== 'number' ||
+            !isFinite(progress) ||
+            progress < 0 ||
+            progress > 1
+        ) {
+            return console.error(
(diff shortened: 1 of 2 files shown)


src/wavesurfer.js
```diff
@@ -739,6 +739,16 @@ export default class WaveSurfer extends util.Observer {
      * wavesurfer.seekTo(0.5);
      */
     seekTo(progress) {
+        if (
```
thijstriemstra: can you add a comment here, this is hard to read.

Issue #1237, opened 2017-10-29
FR: Prevent HTML Audio Tag Injection (lazyload option)


X-Raym:
Hi,

This is a follow-up a the issue discussed on the [Bug: Browsers don't respect setting preload to "none" in wavesurfer.load()](https://github.com/katspaugh/wavesurfer.js/issues/973#issuecomment-340260055) issue.

## Context

This issue concerns using MediaElement backend and Cached Peaks.

## Problem

Some browsers ignore the preload=none attribute of audio tags.
This means that in some cases, audio files are loaded even if not needed.

## Proposition

Only inject audio HTML tags when needed (aka, after a custom event).
For this, we can imagine having a `lazyload` wavesurfer init option attribute, to init the object.
Initializing the object also require knowing its length before hand. (attribute integration proposition by @mspae : https://github.com/katspaugh/wavesurfer.js/commit/1025d7263e32f087a3404012dbf0301e5f95b383 )
This will not be a problem with cached data (the length can be added in the cache file).

This will increase wavesurfer performance and some browser/OS (like Safari+MacOS) a lot ! :)
Thanks for taking a look !

mspae:
@X-Raym I've opened a PR on the duration parameter.

thijstriemstra:
@X-Raym PR was merged and released, closing issue.

X-Raym:
@thijstriemstra Thanks, but has the `lazyload` attribute been integrated or only the `duration` ? As far as I remember, we needed both for this FR, and @mspae pull request was just for `duration`. Correct me if I'm wrong.

@mspae Do we need both attribute for this FR ?

thijstriemstra:
I see. Can you make a pull request?

X-Raym:
@thijstriemstra I personnaly can't, it has become a bit out of my league :S Maybe @mspae had ideas about how to implement it to finalize this FR ?

thijstriemstra:
>  it has become a bit out of my league

That sounds like a problem with the library, can you elaborate?

X-Raym:
@thijstriemstra What I meant is that I don't know how to implement this `lazyload` option, wavesurfer.js is a pretty complex script.

X-Raym:
@katspaugh how has this been solved?

katspaugh:
Please feel free to reopen this if this is still an issue. Browsers might've gotten better since 2017.

PR #1239, opened 2017-10-31
Feature: Duration parameter for the load function

mspae:
Adds an optional duration parameter to `wavesurfer.load` which can be used to explicitly set the duration of the audio (for both backends). This is useful to render the correct wave size in cases where the audio is not decoded:

WebAudio backend with peaks provided currently can't know the duration of the audio and thus always renders it at 100% width initially and then jumps to the correct size on first interaction (when audio is decoded) –
 this can be fixed by using the duration parameter (See one use case in #1210)

/

/

### Todos/Notes:
- Subsequent calls of `load` overwrite the previous duration
- Once the audio is decoded (a buffer is available in `WebAudio.buffer`) its real length is used

- Improved duration parameter feature initially proposed in #1210 
- Is required to implement lazy-loading of audio element (see #1237 )
- Contains #1233 in order to be able to build

### DIFF
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -1026,8 +1026,10 @@ export default class WaveSurfer extends util.Observer {
      * the audio to render the waveform if this is specified
      * @param {?string} preload (Use with backend `MediaElement`)
      * `'none'|'metadata'|'auto'` Preload attribute for the media element
-     * @param {?number} duration The duration of the audio. This is required to
-     * render the peaks data correctly without having to decode the audio
+     * @param {?number} duration The duration of the audio. This is used to
+     * render the peaks data in the correct size for the audio duration (as
+     * befits the current minPxPerSec and zoom value) without having to decode
+     * the audio.
      * @example
      * // using ajax or media element to load (depending on backend)
      * wavesurfer.load('http://example.com/demo.wav');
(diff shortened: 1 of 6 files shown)


src/wavesurfer.js
```diff
@@ -1022,10 +1022,12 @@ export default class WaveSurfer extends util.Observer {
      * Loads audio and re-renders the waveform.
      *
      * @param {string} url The url of the audio file
-     * @param {?number[]|number[][]} peaks Wavesurfer does not have to decode the audio to
-     * render the waveform if this is specified
+     * @param {?number[]|number[][]} peaks Wavesurfer does not have to decode
+     * the audio to render the waveform if this is specified
      * @param {?string} preload (Use with backend `MediaElement`)
      * `'none'|'metadata'|'auto'` Preload attribute for the media element
+     * @param {?number} duration The duration of the audio. This is required to
```
thijstriemstra: I'm not a fan of the word `required` here as this is an optional thing. Could you rephrase this a little?

mspae: Done. I've tried to explain it a little more in the docs and removed the param docs on the private methods for brevity.

mspae:
@thijstriemstra bumped them to `2.0.0-beta03`

thijstriemstra:
Cool. Branch still has conflicts.

PR #1245, opened 2017-11-03
Apply prettier style to mediaelement.js

tf:
My earlier PR #1163 introduced 2 prettier warnings. Linting errors
apparently do not make the build fail.


thijstriemstra:
@tf can you change the build so it fails on lint errors? cc @mspae

PR #1246, opened 2017-11-03
Ensure Travis build fails when webpack build fails

tf:
Commands in package.json scrpts were chained using "&". This causes
parallel execution which might be desired, but causes exit code of all
but last command to be ignored. Use "&&" operator instead to execute
commands sequentially and fail as soon as one does not succeed.

In particular, this ensures the build fails when prettier finds
problems.


tf:
@thijstriemstra applied the requested changes to the build setup @mspae

mspae:
@tf Thanks!

Issue #1267, opened 2017-12-07
Incorrect progress shown if the media is playing in a background-tab


jonas2515:
### Wavesurfer.js version(s)
master-branch

### Browser version(s):
Newest Firefox and Chrome

If using the media element as a backend, the waveform gets updated [when a new frame is drawn](https://github.com/katspaugh/wavesurfer.js/blob/master/src/mediaelement.js#L63).

So when the user starts playback and switches to another tab, the framerate for the current tab is reduced for energy saving reasons and the waveform gets updated at a slower interval.

And since the waveform doesn't get updated if the player isn't playing, finishing the playback while another tab is active and then switching back to the wavesurfer tab can lead to a progress bar "being stuck" somewhere in the middle of the waveform.

I didn't check if this is also the case with the WebAudio backend, as a fix for this one I propose subscribing to the pause event of the media element and updating the waveform one last time.

thijstriemstra:
can you also check if it's the case for the webaudio backend?

jonas2515:
I don't think it is. In the WebAudio backend, the audioprocess event is fired by ScriptProcessorNode in the browser and just forwarded by wavesurfer.

thijstriemstra:
Merged in 2.0.2.

PR #1268, opened 2017-12-07
Fix: MediaElement backend: Update progress on pause events (#1267)

jonas2515:
The progress might not get updated if the playback finishes while another tab is active, this happens because in modern browsers the framerate is reduced to save energy.

Fix it by updating the progress one more time using the pause event instead of relying on a high framerate.

This shouldn't break anything and is just a bugfix.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 wavesurfer.js changelog
 =======================
 
+next (unreleased)
+-----------------
+
+- MediaElement backend: Update progress on pause events (#1267) 
+
 2.0.1 (18.12.2017)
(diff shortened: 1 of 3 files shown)


jonas2515:
I changed it to have consistent behaviour for the WebAudio and the MediaElement backend.

Muting is done by setting the volume in the backend to 0. Since I implemented volume persistence for the MediaElement backend in commit 6b897c4, muting also persists after loading new media. For the WebAudio backend this already was the case before, so setting this.isMuted to false was already wrong.

I actually wanted to do different pull requests for the last two commits, but Github automatically added them.

PR #1271, opened 2017-12-12
Adding barGap as an optional parameter for drawBar render (Proposal for #1058)

### LINKED ISSUE #1058: Set bar spacing as well as width
Wavesurfer.js version(s): 1.3.2
Browser version(s): Firefox 52

I've got some good-looking waveforms being drawn as bars with the barWidth property, but I'd like to be able to set the space between the bars as well; would some kind of barSpace property be possible?

acollier88:
This uses the existing implementation if the parameter is not supplied, otherwise it uses the user defined value with minimum of pixel ratio.

fixes #1058

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ wavesurfer.js changelog
 next (unreleased)
 -----------------
 
+- Added `barGap` parameter to set the space between bars (#1058)
 - Replace deprecated gain.value setter (#1277)
 - MediaElement backend: Update progress on pause events (#1267) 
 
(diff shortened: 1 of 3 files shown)


thijstriemstra:
@acollier88 you also need to add it to the default params (https://github.com/katspaugh/wavesurfer.js/blob/master/src/wavesurfer.js#L179) and document it there.

acollier88:
@thijstriemstra  i added documentation for it however I don't feel that it needs a default value.  Since this is designed to be an optional parameter it falls back to prior behavior if undefined; by providing a default legacy behavior is not preserved.

thijstriemstra:
@acollier88 without it being in the list, how will users even know the option is there? You can set it to undefined and adjust code accordingly

PR #1274, opened 2017-12-18
update dev dependencies and add prepublish command

thijstriemstra:
refs #1092 

Running `npm publish` will build the `dist` folder and publish it.


katspaugh:
Thanks! I'll republish right away.

PR #1290, opened 2018-01-08
Include credentials when fetching audio urls.

jishi9:
Add options to set withCredentials to true on outgoing requests. This allows accessing audio resources hosted on a different domain.

None

None

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -109,6 +109,7 @@ WaveSurfer.util = {
             ajax.fireEvent('error', e);
         });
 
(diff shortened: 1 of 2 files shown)


jishi9:
Note this is a PR for the 'es5' branch (probably the reason why the CI fails)

thijstriemstra:
@jishi9 please make a pr against the master branch instead.

thijstriemstra:
#1310 fixed this.

Issue #1064, opened 2017-04-14
Bad region dragging behaviour with cursor 


GianlucaCal:
Hi, 

I noticed a problem in my project (which you can also see from here http://codepen.io/katspaugh/pen/obmOXV) when I drag or resize any region the cursor follow along. I fixed the resize behaviour by putting this code in the region plugin.

```
if (drag || resize) {
                   
                    if (resize==="end"){
                    my.wavesurfer.params.interact=false;
                    }

                    drag = false;
                    resize = false;
```
I could have putted `my.wavesurfer.params.interact=false;` right after the first `if` but then there would be no possibility to click inside the region. I noticed that there is no difference between drag or click event inside the region, could that be the problem??

Thanks

entonbiba:
@GianlucaCal can you try with this sample and see if you experience the same issue.

http://codepen.io/entonbiba/pen/zwZyOg

GianlucaCal:
not working either.

**Before dragging the region**

![1](https://cloud.githubusercontent.com/assets/24889920/25611508/a5397562-2f27-11e7-8850-b6e7e5a92d38.PNG)

**After dragging the regione the cursor follow**

![2](https://cloud.githubusercontent.com/assets/24889920/25611537/b99a0346-2f27-11e7-9967-142f55210f2f.PNG)

**Resizing the region the cursor follow the risize**

![3](https://cloud.githubusercontent.com/assets/24889920/25611555/c6b9e186-2f27-11e7-8eef-86867d06bec4.PNG)

entonbiba:
@GianlucaCal yea I'll make it so when you drag it, it doesn't set the cursor on drag end.

thijstriemstra:
#1295 seems to fix this, closing.

Issue #572, opened 2015-11-04
seekTo and seek event fires when creating a region


jbeyer05:
Thanks in advance for a wonderful tool.  I have a fairly simple project with regions.  When a new region is created, wavesurfer seems to seekTo to the point of the mouseup action and fires a corresponding seek event.
This would seem to be unintended behavior.  Or is this the intended behavior?

jbeyer05:
One other area where there seems to be similar unintended behavior is when the user drags a region.  When the user drags a region, a click and seekTo event is fired at the area where the user stops dragging the region.

jbeyer05:
One more comment.  If this is indeed unintended behavior, I have created a fairly ugly hack, but I'm not sure how to integrate cleanly.  When the region event listener eventUp( ) senses the end of a completed region, I log the time in a global variable.  In the 'click' event listener in wavesurfer.js, I compare the timestamp in the global variable to the current time.  If it happened in the last 200ms, I do not call seekTo( ).

katspaugh:
Try adding [`stopPropagation`](https://developer.mozilla.org/en/docs/Web/API/Event/stopPropagation) in the `mouseup` handler instead.

David-Crty:
Hello,

 I have tried the stopPropagation function on mouseup but not working.
```javascript
var wave = $('#wave');
var wavesurfer = WaveSurfer.create({
    container: '#wave'
 });
wave.on('mouseup', '.add-comment', function (event) {
    console.log('mouseup');
    event.stopPropagation();
});
```
I see the log mouseup when i click on .add-comment but the seek event fire all the time :(

thijstriemstra:
#1295 seems to fix this, please re-open or leave comment if this is not fixed in 2.x.

PR #1295, opened 2018-01-12
Regions Plugins - prevent Click when creating / updating region

ArTiSTiX:
When creating / updating regions, cursor seeks to mouseup position.
Such behaviour should be prevented, and must be part of application logic (e.g. by listening to `region-updated` or `region-created`, seekTo start of region might be useful).
Thus, interactions with regions plugin must be ignored by drawer.

# Fixes
* Prevent once following click event after updating or creating a region with a preventClick util function
* Added mouse up listener on body when creating region to stop resize if cursor leaves waveform
* Do not fire region update-end event if nothing changed

# Notes:
The only way to prevent click events to "bubble" after a `mouseup` is to capture following click event from body (or a parent to the listener), and stop its propagation.

So i added a preventClick util to expose it to future plugins. It should work when called from any mouseDown/mouseMove/mouseUp events

Reference/Idea: https://stackoverflow.com/questions/33364540/prevent-click-event-from-mousedown

# Related Issues:
https://github.com/katspaugh/wavesurfer.js/issues/1064 - Bad region dragging behaviour with cursor
https://github.com/katspaugh/wavesurfer.js/issues/572 - seekTo and seek event fires when creating a region

# Testing
Only tested on Chrome 63

### DIFF
--- a/src/plugin/regions.js
+++ b/src/plugin/regions.js
@@ -11,6 +11,7 @@ class Region {
         this.wavesurfer = ws;
         this.wrapper = ws.drawer.wrapper;
         this.style = ws.util.style;
+        this.util = ws.util;
 
         this.id = params.id == null ? ws.util.getId() : params.id;
         this.start = Number(params.start) || 0;
@@ -296,6 +297,7 @@ class Region {
                 let drag;
                 let maxScroll;
                 let resize;
+                let updated = false;
                 let scrollDirection;
                 let wrapperRect;
 
@@ -369,7 +371,11 @@ class Region {
                         drag = false;
                         scrollDirection = null;
                         resize = false;
+                    }
 
+                    if (updated) {
+                        updated = false;
+                        this.util.preventClick();
                         this.fireEvent('update-end', e);
                         this.wavesurfer.fireEvent('region-update-end', this, e);
                     }
@@ -394,11 +400,13 @@ class Region {
 
                         // Drag
                         if (this.drag && drag) {
+                            updated = updated || !!delta;
                             this.onDrag(delta);
                         }
 
                         // Resize
                         if (this.resize && resize) {
+                            updated = updated || !!delta;
                             this.onResize(delta, resize);
                         }
 
@@ -748,6 +756,7 @@ export default class RegionsPlugin {
             scrollDirection = null;
 
             if (region) {
+                this.util.preventClick();
                 region.fireEvent('update-end', e);
                 this.wavesurfer.fireEvent('region-update-end', region, e);
             }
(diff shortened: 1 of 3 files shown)


thijstriemstra:
@roboswank can you take a look at this?

thijstriemstra:
@ArTiSTiX this PR has some conflicts now, can you take a look at that?

ArTiSTiX:
@thijstriemstra I will have a look tomorrow, and test it with all new changes from 2.0.1 (the version i use).
I think the conflict can be resolved by accepting both additions (in any order, it's just a variable declaration).

Thanks for considering this merge.

ArTiSTiX:
I just noticed mouse events now ends as soon as the cursor leaves the container.
It might be a recent planned behaviour, but I think it's more user-friendly to let him drag off the wave-form.

Didn't had time to inspect concerned changes/issues today.

thijstriemstra:
> I just noticed mouse events now ends as soon as the cursor leaves the container.
It might be a recent planned behaviour, but I think it's more user-friendly to let him drag off the wave-form.

@roboswank can you give some feedback on that?

ArTiSTiX:
@thijstriemstra @roboswank Sorry, i remembered wrongly. 
✅ Leaving container doesn't end region dragging, but outer moves doesn't update (i thought it used to receive mousemoves but it never worked like that, however it could be an improvement).
✅ Re-entering still drags
✅ mouseup/touchend ends dragging even if out of container (i fixed it in this PR, wasn't working when creating regions)
❓ Moving mouse could drag region out of container by creating the move listener on document instead of the wrapper

thijstriemstra:
@ArTiSTiX I do not use the regions plugin, nor do I pretend to understand it's code, so if you can make the necessary adjustments in this pull request, we can take a look.

ArTiSTiX:
I prefer keeping this PR light, only fixing related issues and not implementing new behaviour.
I will file an dedicated issue (open to discussion) about the possibility of dragging out of container to improve user experience.
I plan to work on other improvements requested on regions plugins (e.g. non-overlapping regions).

src/plugin/regions.js
```diff
@@ -11,6 +11,7 @@ class Region {
         this.wavesurfer = ws;
         this.wrapper = ws.drawer.wrapper;
         this.style = ws.util.style;
+        this.util = ws.util;
```
thijstriemstra: can you swap line 13 and 14 and use this.style = this.util.style?

src/util/prevent-click.js
```diff
@@ -0,0 +1,8 @@
+function preventClickHandler(e) {
```
thijstriemstra: Can you add a docstring to this module? e.g.

```
/**
 * Description here
 *
 * @return {function}
 */
```

johndebord:
On mobile, if a region is moved or adjusted, the user has to double-tap to get anything to work.
What's the recommended work-around for this?

PR #1300, opened 2018-01-15
add support for sink id

### LINKED ISSUE #1293: Ability to select audio output device
### Wavesurfer.js version(s):
^1.1.11 but seems like it affects all

### Browser version(s):
Chrome

As far as i can tell there's no way to specify the audio output device for wavesurfer. My application has a speaker selector and as such I need to be able to set the output to the default device/bluetooth headset etc. 

Is there any supported / unsupported way to accomplish this?

johnryan:
fixes #1293

### DIFF
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -791,7 +791,7 @@ export default class WaveSurfer extends util.Observer {
     /**
      * Set the playback volume.
      *
-     * @param {number} deviceId String value representing underlying output device
+     * @param {string} deviceId String value representing underlying output device
      */
     setSinkId(deviceId) {
         this.backend.setSinkId(deviceId);
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -791,6 +791,15 @@ export default class WaveSurfer extends util.Observer {
     /**
      * Set the playback volume.
      *
+     * @param {number} deviceId String value representing underlying output device
+     */
+    setSinkId(deviceId) {
+        this.backend.setSinkId(deviceId);
+    }
+
+    /**
+     * Set the playback volume.
+     *
      * @param {number} newVolume A value between 0 and 1, 0 being no
      * volume and 1 being full volume.
      */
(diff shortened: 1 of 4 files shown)


thijstriemstra:
thanks @johnryan. some problems reported by build: https://travis-ci.org/katspaugh/wavesurfer.js/builds/329266525#L631 can you fix those?

src/webaudio.js
```diff
@@ -275,6 +275,27 @@ export default class WebAudio extends util.Observer {
...
+                return;
+            }
+            audio.autoplay = true;
+            var dest = this.ac.createMediaStreamDestination();
+            this.gainNode.disconnect();
+            this.gainNode.connect(dest);
+            audio.src = URL.createObjectURL(dest.stream);
+            audio.setSinkId(deviceId);
+        }
+    }
```
thijstriemstra: @johnryan why are you creating a new Audio instance here? Can the regular one not be (re-)used?

johnryan: @thijstriemstra which instance are you referring to? The webaudio apis don't support the `setSinkId` concept so I create an HTMLAudioElement and attach the stream to that.

thijstriemstra: I see: https://stackoverflow.com/a/43069474

```
Unfortunately, setting the destination audio device of a webaudio graph isn't yet implemented, and the api for this is not yet finalized.

What you can do for now, is connect the webaudio graph to an HTML element, and set the sinkid of the element (currently works on Chrome only)
```

Can you add a summary of that in a multiline comment so it's clear why you're doing that?

thijstriemstra: And I would also throw errors instead of showing the warning, console.log is not used inside the library itself.

src/wavesurfer.js
```diff
@@ -788,6 +788,15 @@ export default class WaveSurfer extends util.Observer {
         this.drawer.progress(0);
     }
 
+    /**
+     * Set the playback volume.
+     *
+     * @param {number} deviceId String value representing underlying output device
```
thijstriemstra: if it's a string, document param as string.

src/webaudio.js
johnryan: @thijstriemstra done

thijstriemstra: Awesome, thanks.

PR #1303, opened 2018-01-17
fix deprecated value setter for playback rate

### LINKED ISSUE #1302: Chrome deprecation warning for AudioParam value setter
This line https://github.com/katspaugh/wavesurfer.js/blob/master/src/webaudio.js#L523 triggers a deprecation warning in Chrome:

```
[Deprecation] AudioParam value setter will become equivalent to AudioParam.setValueAtTime() in M65, around March 2018  See https://webaudio.github.io/web-audio-api/#dom-audioparam-value for more details.```

Similar to #1277

thijstriemstra:
fixes #1302

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@ next (unreleased)
 -----------------
 
 - Added support for selecting different audio output devices using `setSinkId` (#1293)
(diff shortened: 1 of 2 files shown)


thijstriemstra:
ping @katspaugh

Issue #1038, opened 2017-03-13
authorization headers enhancement


sht5:
Hello,
i am looking for a way to add an authorization header to the requests wavesurfer makes when loading a url. i want to load a url dependent on authorization.
thanks

katspaugh:
You can load the audio on your own (using XMLHttpRequest or the Fetch API), and then pass the response body to [this internal method](https://github.com/katspaugh/wavesurfer.js/blob/master/src/wavesurfer.js#L361) called `loadArrayBuffer`.

sht5:
actually its also possible to just send a jwt as a query parameter on the
url so this solves the problem for me

TrevorHinesley:
Is there no better way to do this? I have to have authorization as well, and the tokens are long-lived so passing them as a query parameter isn't secure. It needs to be a header.

Can you give an example of how to load the file ourselves?

TrevorHinesley:
For anyone wondering any easy fix for this, I overrode the `WaveSurfer.util.ajax` method like so:

```
// Other than where specified, this should be the same as:
//   https://raw.githubusercontent.com/katspaugh/wavesurfer.js/master/src/util.js
//
// Override ajax function to add our auth headers
WaveSurfer.util.ajax = function (options) {
    var ajax = Object.create(WaveSurfer.Observer);
    var xhr = new XMLHttpRequest();
    var fired100 = false;

    xhr.open(options.method || 'GET', options.url, true);

    // START added code
    xhr.setRequestHeader("Authorization", "Bearer <your-auth-token>")
    // END added code

    xhr.responseType = options.responseType || 'json';

    xhr.addEventListener('progress', function (e) {
        ajax.fireEvent('progress', e);
        if (e.lengthComputable && e.loaded == e.total) {
            fired100 = true;
        }
    });

    xhr.addEventListener('load', function (e) {
        if (!fired100) {
            ajax.fireEvent('progress', e);
        }
        ajax.fireEvent('load', e);

        if (200 == xhr.status || 206 == xhr.status) {
            ajax.fireEvent('success', xhr.response, e);
        } else {
            ajax.fireEvent('error', e);
        }
    });

    xhr.addEventListener('error', function (e) {
        ajax.fireEvent('error', e);
    });

    xhr.send();
    ajax.xhr = xhr;
    return ajax;
}
```

Worked like a charm :)

TrevorHinesley:
[...]

PR #1310, opened 2018-01-24
Support for XHR options

janhartmann:
This pull request expands the `load(url)` method to support object initialisation. This means you will be able to add headers and other request specific properties to the load function.

Example:

```js
this.wavesurfer.load({
    url: '/url/to/file.wav',
    requestHeaders: [
        { 
            key: "Authorization": 
            value: "my-token"
        }
    ]
});
```

Other options within `ws.util.ajax` will also be overridable.

- I've added the support for adding request headers on the the XHR request.
- I've handling the support for having the `url` variable both as a string and as a object.
- I've added a preload check whether the object as a minimum contains a `url` property.

Should not be any.

Should not be any.

### Todos/Notes:
We could argue that `url` is not a suitable name for the variable anymore, since it can now be a string, object and an element. 

Authorization headers enhancement #1038
#1100

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ wavesurfer.js changelog
 next (unreleased)
 -----------------
 
+- Added `xhr` option to configure util.ajax for authorization (#1310, #1038, #1100)
 - Fix `setCurrentTime` method (#1292)
 - Fix `getScrollX` method: Check bounds when `scrollParent: true` (#1312)
 - Minimap plugin: fix initial load, canvas click did not work (#1265)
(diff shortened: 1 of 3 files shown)


janhartmann:
Wauw, that a huge pull request. But yes, I see.

Not sure I got the time on my hands right now for looking into that. Feel free to close this merge request! :-)

thijstriemstra:
The pull request is huge, but the actual diff is tiny..

thijstriemstra:
@janhartmann refactored the changes and added an `xhr` parameter. Instead of passing the object to `load` use that wavesurfer option instead and supply it with an object, e.g.

```javascript
var xhr = {
    requestHeaders: [
        { 
            key: "Authorization": 
            value: "my-token"
        }
    ]
}
```

Also incorporated the `withCredentials` option from #1100.

janhartmann:
This is perfect for our needs, thank you so much!

PR #1313, opened 2018-01-26
Keep scrollX in bounds

### LINKED ISSUE #1312: In safari on mac the waveform disappears when scrolling via the mouse wheel
### Wavesurfer.js version(s):
2.0.3

### Browser version(s):
Safari v11.0.1 on OSX v10.13.1

Reproducible for me every time in Safari on macbook touchpad and mac mini magic trackpad.
Create a waveform with the param `scrollParent: true` and scroll along the x axis using the "mouse wheel".
If you scroll to the very end or back to the beginning the waveform will disappear and an error will log with further scrolling.
The error occurs in `fillRect` due to the index being out of bounds (there not being an entry canvas at -1 or >=canvases.length).
Safari has an elastic scroll when using the "mouse wheel" and allows for scrollX to be out of bounds.

To fix drawer.js line 223
```javascript
    getScrollX() {
        var maxScroll = this.params.scrollParent
            ? ~~(
                  this.wrapper.scrollWidth * this.params.pixelRatio -
                  this.getWidth()
              )
            : this.getWidth();

        return Math.min(
            maxScroll,
            Math.max(
                0,
                Math.round(this.wrapper.scrollLeft * this.params.pixelRatio)
            )
        );
    }
```

Will create a PR.

roboswank:
Keep scrollX from going beyond the container or below 0.

### Notes:
The only disadvantage I see to this check happening here, is if someone is counting on special behavior when using elastic scroll beyond the container extents. In which case `drawBuffer` needs to be checking that `start`/`end` stay in bounds.

fixes #1312

### DIFF
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -220,7 +220,7 @@ export default class Drawer extends util.Observer {
      * @return {number}
      */
     getScrollX() {
-        return Math.round(this.wrapper.scrollLeft * this.params.pixelRatio);
+        return Math.min(this.getWidth(), Math.max(0, Math.round(this.wrapper.scrollLeft * this.params.pixelRatio)));
     }
 
     /**
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -220,8 +220,11 @@ export default class Drawer extends util.Observer {
      * @return {number}
      */
     getScrollX() {
+        var maxScroll = this.params.scrollParent
+            ? this.wrapper.scrollWidth - this.getWidth()
+            : this.getWidth();
         return Math.min(
-            this.getWidth(),
+            maxScroll,
             Math.max(
                 0,
                 Math.round(this.wrapper.scrollLeft * this.params.pixelRatio)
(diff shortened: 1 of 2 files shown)


thijstriemstra:
@roboswank can you also add this to the changelog and put a comment or two above your change explaining what it's doing?

roboswank:
@thijstriemstra Will do.

thijstriemstra:
Looks good to me, maybe @katspaugh can have the final say.

src/drawer.js
```diff
@@ -220,7 +220,23 @@ export default class Drawer extends util.Observer {
...
     getScrollX() {
-        return Math.round(this.wrapper.scrollLeft * this.params.pixelRatio);
+        const pixelRatio = this.params.pixelRatio;
+        let x = Math.round(this.wrapper.scrollLeft * pixelRatio);
+
+        // In cases of elastic scroll (safari with mouse wheel) you can
+        // scroll beyond the limits of the container
+        // Calculate and floor the scrollable extent to make sure an out
+        // of bounds value is not returned
+        // https://github.com/katspaugh/wavesurfer.js/issues/1312
```
thijstriemstra: you dont have to include the full url, just a ticket reference should be fine.

roboswank: Gotcha, will update.

PR #1318, opened 2018-02-04
Regions plugin - Scroll when near edges of container

### LINKED ISSUE #430: Dragging a region doesn't work outside of scrollbar
When I try to drag a region, with a part of that region inside view and a part outside of the view, I expect that the scrollbar moves with my mouse so I can select more that what's currently in view. Maybe the current behavior (selecting only inside view) is the expected behavior for others?

roboswank:
When dragging, if mouse or region edge is near the container's edges, scroll in that direction.

Added checks, within region mouse move events, to scroll the container when within a certain threshold of the container's edges.

New region plugin options available:
`scroll` - defaults to true
`scrollSpeed` - defaults to 1
`scrollThreshold` - defaults to 10

fixes #430

### DIFF
--- a/src/plugin/regions.js
+++ b/src/plugin/regions.js
@@ -32,6 +32,10 @@ class Region {
         this.minLength = params.minLength;
         this._onRedraw = () => this.updateRender();
 
+        this.scroll = params.scroll !== false && ws.params.scrollParent;
+        this.scrollSpeed = params.scrollSpeed || 1;
+        this.scrollThreshold = params.scrollThreshold || 10;
+
         this.bindInOut();
         this.render();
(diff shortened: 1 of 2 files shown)


src/plugin/regions.js
```diff
@@ -639,12 +780,25 @@ export default class RegionsPlugin {
                 region = this.add(params || {});
             }
```
roboswank: Not sure if this really needs to be updated on each interaction - but moved it to `eventDown` so it is not being called on every mouse move.

thijstriemstra:
@roboswank can you add a changelog item for this?

Issue #1319, opened 2018-02-05
form with direction RTL have render issues with TimeLine + Zoom and other


awenger63:
### Wavesurfer.js version(s): 
v2.01

### Browser version(s):
Chrome 64.0.3282.140 , Firefox quantum 58.0.1

### Code needed to reproduce the issue:
Add dir="RTL" on your form and test seek, play ..
I do it on this code  : [https://github.com/katspaugh/wavesurfer.js/issues/new](url)
`<div **dir="RTL"**><div id="waveform"></div>
<div id="waveform-timeline"></div>
</div> ....`

Workarround for me => always put dir="LTR" in wave div !
In this case wavesurfer is alway LTR for all.

thijstriemstra:
@awenger63 what are you suggesting exactly? Any improvements that can be made to this library?

awenger63:
First suggestion can be  : have a default value for dir = "LTR" than wavesurfer always ready when it'is implemented in a RTL form (like the workarround). It's for make a trace that I have created this issue and if someone have same issue.

Second, it's depend of RTL country habits. In this case, users would like the wave on the other side ... This need require to reverse or inverse all functionnality (plugins too) and it should be a little complexe. Today it'snt necessary for my actual end user, wavesurfer's always the same direction. Perhaps nobody need, than first suggestion can be the solution to not inherit dir property.

is it answer the questions ?

thijstriemstra:
@awenger63 can you make a pull request and change the actual code so it's easier to see your intention?

thijstriemstra:
also see #1540

awenger63:
Thanks @thijstriemstra , it will be better than my workarround to have a solution for LTR if it' need for user country usage. Use a param to make it like an option is very friendly because some user (LTR env) would like stay with wave RTL !

awenger63:
Nice, I will test it with the end users when they want to put it in place

Issue #1296, opened 2018-01-14
Right to Left Support


samanmohamadi:
Thanks for the Great work.
in my project, I need the wave to be right-to-left.
is there any simple way to achieve this?
Does replacing all `right` and `left` in the project do the job?

thijstriemstra:
@samanmohamadi good question.. @katspaugh you have an idea how hard this would be to implement?

katspaugh:
Simply adding `style="transform: rotateY(180deg)"` to the container should do the trick.

thijstriemstra:
Does that work for you @samanmohamadi? If it does, adding an example would be nice.

samanmohamadi:
It works but there is an issue with calculating progress when mouse click.
a little change in  [handleEvent](https://github.com/katspaugh/wavesurfer.js/blob/cec39f9abf48600f58c9d1e0be7a383b0213e09f/src/drawer.js#L86) function will fix it.

katspaugh:
A PR is welcome, @samanmohamadi!

samanmohamadi:
I'm am in the middle of a forced project and my solution to this issue is a little dirty and it not ready for a pr. I will do that as soon as possible.

thijstriemstra:
Maybe related to #1319

PR #1331, opened 2018-02-20
Suggestion for always wavesurfer work on LTR independant of inheritance

awenger63:
see #1319 : with this change, wavesurfer.js always work and stay Left to Rigth (like audio) with correct functionnality if it's use in HTML form with dir ='RTL' since a global RTL implementation #1296 ?

### DIFF
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -295,7 +295,7 @@ export default class WaveSurfer extends util.Observer {
         }
 
         /** set dir always ltr, only way 'today' to work independant of inheritance */
-        this.container.dir='ltr';
+        this.container.dir = 'ltr';
 
         if (this.params.maxCanvasWidth <= 1) {
             throw new Error('maxCanvasWidth must be greater than 1');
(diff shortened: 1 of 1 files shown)


thijstriemstra:
@awenger63 build fails, see https://travis-ci.org/katspaugh/wavesurfer.js/builds/343809271#L513

awenger63:
Thanks @thijstriemstra, need to improve.

thijstriemstra:
@awenger63 can you also reword the comment:

```
/** set dir always ltr, only way 'today' to work independant of inheritance */
```

maybe using google translate, because i don't understand what it means..

awenger63:
No google translate, but a 'shortly' description, perhaps some word are implicit ! in french google translation it's compréhensible ;-)

Set the 'dir' property always at "ltr" to force the value. This is the only solution today to wavesurfer and his pluging work correctly when is it implemented in a RLT form and it inherit this property.

Are you agree with this comment ? Is it clear like this ? If Yes, I will make the change.

thijstriemstra:
@awenger63 can you turn this into a new wavesurfer.js option and make the default 'ltr'? I prefer that over hardcoding it.

PR #1338, opened 2018-02-25
Add isReady methods

### LINKED ISSUE #1333: [Enhancement] isReady method
Hey !

Thanks for wavesurfer it's awesome.
I was thinking about adding an isReady() function wich simply return true or false. The event 'ready' is working and I can recreate the isReady function by myself I know but that implies I can not unsub the event from another callback.

If you have any idea. Thanks ! :)

Guisch:
Add an isReady method to know if the wavesurfer is loaded and ready or not.

### Todos/Notes:
Need to be tested

fixes #1333

### DIFF
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -901,7 +901,7 @@ export default class WaveSurfer extends util.Observer {
      * @return {boolean}
      */
     isReady() {
-        return this.isMuted;
+        return this.isReady;
     }
 
     /**
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -895,6 +895,16 @@ export default class WaveSurfer extends util.Observer {
     }
 
     /**
+     * Get the current ready status.
+     *
+     * @example const isReady = wavesurfer.isReady();
+     * @return {boolean}
+     */
+    isReady() {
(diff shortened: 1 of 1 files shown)


src/wavesurfer.js
```diff
@@ -894,6 +894,16 @@ export default class WaveSurfer extends util.Observer {
...
     }
 
+    /**
+     * Get the current ready status.
+     *
+     * @example const isReady = wavesurfer.isReady();
+     * @return {boolean}
+     */
+    isReady() {
+        return this.isMuted;
```
thijstriemstra: @Guisch what does `isMuted` have to do with this? :)

Guisch: Ahah, thanks Copy/Paste

thijstriemstra:
@Guisch if there's a `isReady` attribute, why add a method that returns this property, why not use this property directly?

Guisch:
Because it's a private attribute ? (it's the same logic for isMuted attribute and getMute function).
https://github.com/katspaugh/wavesurfer.js/blob/07306849a71db2f02c607bd53f8d06003b315c8a/src/wavesurfer.js#L356

PR #1343, opened 2018-03-03
Tiled Renderer plug-in

jamiefaye:
```javascript
/*
The TiledRenderer works with a pool of Canvas objects,
automatically re-imaging them and positioning them	as needed.
It maintains a large 'div' element which is held open to the maximum
length of the song.	As the scroll position changes, the 
TiledRenderer determines if a new area is about to emerge into
view	and if one is, recycles the canvas that is farthest
away and repositions it.

The TiledRenderer can zoom down to the sample level. At a
given threshold, it changes the drawing technique	from peak
rendering to sample rendering. Since the sample data is already
present, the TiledRenderer avoids generating or using the
peak data, preferring instead to use the sample data directly.
If the zoom level is such	that multiple samples would map to
a single pixel, the 'nearest neighbor' sample is used.

In order to take control of the use of the 'peaks' arrays,
I created a special version of the drawBuffer method that I
patch-in after creating the wavesurfer object, as shown in the
example below.

I have not (yet) gone through and made sure that all the
APIs work. For example, the getImage call only returns the
images from the canvases that actually exist. I probably missed
other stuff.
*/

import WaveSurfer from './js/wavesurfer.js';
import {TiledRenderer, tiledDrawBuffer} from './js/plugins/wavesurfer.tiledrenderer.js';

// Example use:
var surfer = WaveSurfer.create({
	container:		'#waveform',
	waveColor:		'violet',
	progressColor:	'purple',
	splitChannels:	true,
	interact:		false,
	fillParent:		false,
	scrollParent:	true,
	partialRender:  false,
	renderer:		TiledRenderer,
});  

// Patch in our override to the drawBuffer function.
surfer.drawBuffer = tiledDrawBuffer;
//
//
// Optional params you can set:

// Number of canvases to cache.
// params.canvasLimit	 		// Default: 6: Don't set this below 4 or the cat will chase its tail!

// Use the sample-drawer if the minPxPerSec is >= sampleSpeed, otherwise use peaks/bars.
// params.sampleSpeed	// default is 2560px per second;
```


thijstriemstra:
ping @jamiefaye

thijstriemstra:
added an example but scrolling nearly crashes the browser..

thijstriemstra:
Going to close the pull request because there is no feedback (and PR is broken as-is).

PR #1344, opened 2018-03-04
More tests

thijstriemstra:
Add more test coverage.


spec/util.spec.js
```diff
@@ -1,8 +1,111 @@
...
+                    }
+                ]
+            }
+        };
+        var instance = WaveSurfer.util.ajax(options);
+        instance.on('success', (data, e) => {
+            // with credentials
+            expect(e.target.withCredentials).toBeTrue();
+
+            // XXX: find a way to retrieve request headers
```
thijstriemstra: @janhartmann any idea for this one? seems setting request headers is possible but i can't find a public property that returns them during this test..

PR #1345, opened 2018-03-05
two new events 'mute' and 'volume'

traburiss:
add new event 'mute'  when the function toggleMute is call and return the mute status
add new event 'volume'  when volume is changed


thijstriemstra:
@traburiss thanks, looks good. Made some minor changes, will merge when build passes.

PR #1396, opened 2018-06-17
Update isReady attribute

Guisch:
The isReady attribute is set to false at the creation of a wavesurfer. Then, It is set to true when a song is loaded but it is never set to false when emptying the wavesufer. As a result, after the first song is loaded, isReady return true even if the wavesurfer is empty.

### DIFF
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -1449,6 +1449,7 @@ export default class WaveSurfer extends util.Observer {
             this.stop();
             this.backend.disconnectSource();
         }
+        this.isReady = false;
         this.cancelAjax();
         this.clearTmpEvents();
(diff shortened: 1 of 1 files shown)


thijstriemstra:
Thanks!

thijstriemstra:
@Guisch I think it should also be set to false in `destroy`?

Guisch:
Yes !

thijstriemstra:
@Guisch can you also do a PR for that? (and verify it's false in the `destroy` unit test)?

Guisch:
Done (except for the unit test, sorry i do not see what you'r talking about)

thijstriemstra:
cool. yea the unit test for `destroy` is missing, can you open a new ticket for that?

PR #1397, opened 2018-06-19
timeline plugin: add parameters for start time value and format

### LINKED ISSUE #1380: change start second and format on timeline plugin
New needs on Timeline plugin :
 - change start second by real value instead of 0. This case is for use this plugin for a recording, I would like to begining the Timeline by the rigth hour of the record begin.
 - change Timeline format to add hour of the day corresponding at the record date.
see this example screenshot :
image

Could you created a pull request for this, I have make local code can be share with you ? Thanks

awenger63:
fixes #1380 

Make startSeconds like a parametre can be change initial value.
Add formatTime parameter to have possibility to change default format

No change on current using timeLine !

Can be call when create wavesurfer like this to change startSecond and/or cahnge formatTime :
 
`plugins: [TimelinePlugin.create({container: '#waveform-timeline', startSeconds: 3600, formatTime: 'HH:MM:SS'})]`

### DIFF
--- a/src/plugin/timeline.js
+++ b/src/plugin/timeline.js
@@ -16,7 +16,9 @@
  * performance for large files
  * @property {string} fontFamily='Arial'
  * @property {number} fontSize=10 Font size of labels in pixels
- * @property {function} formatTimeCallback=→00:00
+ * @property {number} startSeconds=0 Start time in second for begining timeline rule
+ * @property {string} formatTime='MM:SS' default time format for timeline rule, can be change to'HH:MM:SS'
+ * @property {function} formatTimeCallback=→00:00 (or 00:00:00 depend of formatTime property)
  * @property {?boolean} deferInit Set to true to manually call
  * `initPlugin('timeline')`
  */
@@ -132,14 +134,26 @@ export default class TimelinePlugin {
                 fontFamily: 'Arial',
                 fontSize: 10,
                 zoomDebounce: false,
-                formatTimeCallback(seconds) {
+                startSeconds : 0,
+                formatTime : 'MM:SS',
+                formatTimeCallback(seconds, format) {
                     if (seconds / 60 > 1) {
                         // calculate minutes and seconds from seconds count
-                        const minutes = parseInt(seconds / 60, 10);
+                        minutes = parseInt(seconds / 60, 10);
                         seconds = parseInt(seconds % 60, 10);
                         // fill up seconds with zeroes
                         seconds = seconds < 10 ? '0' + seconds : seconds;
-                        return `${minutes}:${seconds}`;
+                        if (format === 'HH:MM:SS') {
+                            // calculate hours from minutes
+                            var hours = parseInt(minutes / 60, 10);
+                            minutes = parseInt(minutes % 60, 10);
+                            // fill up minutes and hours with zeroes
+                            minutes = minutes < 10 ? '0' + minutes : minutes;
+                            hours = hours < 10 ? '0' + hours : hours;
+                            return `${hours}:${minutes}:${seconds}`;
+                        } else {
+                            return `${minutes}:${seconds}`;
+                        }
                     }
                     return Math.round(seconds * 1000) / 1000;
                 },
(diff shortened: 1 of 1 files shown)


src/plugin/timeline.js
```diff
@@ -16,7 +16,9 @@
  * performance for large files
  * @property {string} fontFamily='Arial'
  * @property {number} fontSize=10 Font size of labels in pixels
- * @property {function} formatTimeCallback=→00:00
+ * @property {number} startSeconds=0 Start time in second for begining timeline rule
```
thijstriemstra: there are 2 typos: `second` and `begining`

src/plugin/timeline.js
```diff
@@ -16,7 +16,9 @@
...
+ * @property {string} formatTime='MM:SS' default time format for timeline rule, can be change to'HH:MM:SS'
```
thijstriemstra: 2 typos: `change` should be `changed` and missing space after `to`

src/plugin/timeline.js
```diff
@@ -16,7 +16,9 @@
...
+ * @property {function} formatTimeCallback=→00:00 (or 00:00:00 depend of formatTime property)
```
thijstriemstra: `depending on` instead of `depend of`

and can you remove the arrow: `→`

src/plugin/timeline.js
awenger63: OK

src/plugin/timeline.js
awenger63: OK

src/plugin/timeline.js
awenger63: OK, Arrow exist in original file, it's not me this time ;-)

thijstriemstra:
Not sure if you're ready yet but there are still some review comments that are not addressed @awenger63

awenger63:
@thijstriemstra you'r right. Sorry I'm little busy.
I suggest (have realise on my side) to add also another parameter. I would like your great view and translation validation. His name could be "isRealFirstInterval: true if the interval of the rule begin on the real startSeconds or in the first timeInterval rounded "

To explain you the difference of this 2 possibilities, see screenshot (begin and end timeLine with zoom) :
If is RealFirstInterval (default value) with a startSeconds = 3768 (01:02:48)
image

If is not RealFirstInterval with same condition, interval will be recalculate depend of timeInterval value and zoom process like this :
image

What do you think about this change ?
The naming convention for the new paramaters looks like good ?
Thanks per advance for your return.
Regards

thijstriemstra:
> I suggest (have realise on my side) to add also another parameter. 

Please open a new ticket/PR for that and keep this PR on topic.

awenger63:
OK, I undestrand, but to work I need startSeconds parmeter to work, and without this change I think the solution for add start time is cannot be complet. How can I do this PR, it's a new branch from this current PR ? is it possible to do that and merge it after ?

thijstriemstra:
I'm sorry, I don't understand what you're saying. I just review code and approve it makes sense, or reject if it doesn't. (I do not use the timeline plugin myself).

awenger63:
OK, I will try to explain. In fact, first need #1380 it's to change start time. I do this in the current PR and it work for me.
But when I have use it in real case, I see when Start Time isn't a rounded value, it will be great to change also the first interval to have a better render on the rule below (see screenshot example). Sometime, depend of interval precision,  It's better to see a rounded time % 10 or 5 without the real second.
It's for that I upgrade the code for implement this possibility.

Then either I update this PR, either I need to create a PR linked at this PR like a new branch (I don't know if it's possible and how I can do this, I need to improve myself with Git usage and sharing code)

I would like also, before your review in Git, your suggestion concerning the name and the comment I would like to use for this new parameter and functionnality.

It's understandable ?

awenger63:
@thijstriemstra I update your review comment, and I create a new PR from this PR to complet the solution. how and when can I merge them ? Thanks per advance.

thijstriemstra:
@awenger63 can you take a look at master branch, some improvements for the timeline plugin were added that (I think) you also tried to address in this PR.

thijstriemstra:
@awenger63 this should be fixed with #1458

PR #1420, opened 2018-08-06
logo proposal

entonbiba:
Svg and png versions if required will upload jpg version as well. 

Wavesurfer.js Logo 
- [x] png version
- [x] svg version
- [] update docs to include logo

to be decided:
- [] move files to folder specific (/logo/) or leave in main root folder


entonbiba:
@katspaugh not sure if it should be included in the master branch or just the gh-pages branch only?

katspaugh:
@entonbiba thank you! Could you also commit the non-textual part of the logo as a separate file?

As for the textual part, I think it should be all lower-case and `.js` should probably be the same size as the rest (to differentiate better from the well-known Wavesurfer project).

entonbiba:
@katspaugh I created white/black text versions and also symbol only versions all in png and svg formats.

entonbiba:
@katspaugh I updated the docs, #1426

PR #1435, opened 2018-08-21
fix cursor plugin destroy

Bobrovskih:

in cursor plugin
move variable wrapper to this,
else when call destroy method its throw error
cannot read property removeListener of undefined


thijstriemstra:
thanks!

PR #1441, opened 2018-08-26
regions plugin: display regions before file load

TimSC:
Pre-generated peaks waits for the audio length, either from loaded audio or from explicitDuration. This PR adds explicitDuration as an argument for the WebAudio constructor. This allows it to be defined up front.

I've also updated the regions plugin initialization to make use of this information so the "backend-created" event is used to add regions, rather than the "ready" event which only fires when a song has finished loading.

### DIFF
--- a/src/plugin/regions.js
+++ b/src/plugin/regions.js
@@ -665,6 +665,7 @@ export default class RegionsPlugin {
 
     destroy() {
         this.wavesurfer.un('ready', this._onReady);
+        this.wavesurfer.un('backend-created', this._onBackendCreated);
         this.disableDragSelection();
         this.clear();
     }
--- a/src/plugin/regions.js
+++ b/src/plugin/regions.js
@@ -177,34 +177,34 @@ class Region {
 
     /* Update element's position, width, color. */
     updateRender() {
-        // Duration varies during loading process, so don't overwrite important data
+        // duration varies during loading process, so don't overwrite important data
         const dur = this.wavesurfer.getDuration();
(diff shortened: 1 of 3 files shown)


thijstriemstra:
thanks @TimSC this looks fine. can you also add a unit test in `spec/wavesurfer.spec.js` for this?

TimSC:
I will try but I'm not that great at javascript!

src/plugin/regions.js
```diff
@@ -177,31 +177,34 @@ class Region {
...
     updateRender() {
+        // Duration varies during loading process, so don't overwrite important data
         const dur = this.wavesurfer.getDuration();
         const width = this.getWidth();
 
-        if (this.start < 0) {
-            this.start = 0;
-            this.end = this.end - this.start;
+        var start_limited = this.start;
+        var end_limited = this.end;
```
katspaugh: Could you write these variables with camelCase?

src/wavesurfer.js
```diff
@@ -187,6 +187,7 @@ export default class WaveSurfer extends util.Observer {
         cursorColor: '#333',
         cursorWidth: 1,
         dragSelection: true,
+        explicitDuration: null,
```
katspaugh: I would call it simply `duration` and mark as optional in the docs.

thijstriemstra:
ping @TimSC

TimSC:
Great thanks so much! I was a bit out of my depth using javascript.

PR #1448, opened 2018-08-29
Change the time format for seconds from ceil to floor 

galexynye:
In the Angular Player example, when the time was suppose to hit a new minute and display 01:00 or 02:00 it would display 00:00 for one second before displaying 01:01 or 02:01. By changing the format function it now displays each minute properly and the length is the actual file length.


thijstriemstra:
thanks!

galexynye:
:)

PR #1463, opened 2018-09-18
plugins: fix microphone plugin on MS Edge 

### LINKED ISSUE #627: Microphone plugin issue in Edge browser
When testing the microphone example in Edge 20 I get the following error on [this line](https://github.com/katspaugh/wavesurfer.js/blob/master/src/webaudio.js#L499) of webaudio.js:

```
Invalid AudioBuffer. You may not use an AudioBuffer from an 'audioprocess' event.
```

Commenting out that line seems to fix the problem:

``` diff
diff --git a/src/webaudio.js b/src/webaudio.js
index eef8f4b..0e40480 100755
--- a/src/webaudio.js
+++ b/src/webaudio.js
@@ -267,7 +267,7 @@ WaveSurfer.WebAudio = {
         this.source.stop = this.source.stop || this.source.noteOff;

         this.source.playbackRate.value = this.playbackRate;
-        this.source.buffer = this.buffer;
+        //this.source.buffer = this.buffer;
         this.source.connect(this.analyser);
     },
```

Now, why is this line there, and is it needed for any browsers, can we live without? Or should it be disabled on edge only?

mronkain:
This PR just makes the microphone plugin work on Edge: the audio data received from onaudioprocess is first copied to a local AudioBuffer which is passed on instead of the event.inputBuffer. Note that there is a lot of blinking whenever a new buffer is received.

There are no breaking changes.

fixes #627

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,10 +4,12 @@ wavesurfer.js changelog
 2.0.7 (unreleased)
 ------------------
 
+- Fix microphone plugin on MS Edge browser (#627)
 - Add wavesurfer.js logo, created by @entonbiba (#1409)
 - Library version number is now available as `WaveSurfer.VERSION` (#1430)
 - Fix `setSinkId` that used deprecated API (#1428)
 - Set `isReady` attribute to false when emptying wavesufer (#1396, #1403)
+- Fix cursor plugin `destroy` (#1435)
 
 2.0.6 (14.06.2018)
 ------------------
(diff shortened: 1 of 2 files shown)


mronkain:
@thijstriemstra I can add browser detection if you want but I didn't intend this fix as pure workaround since the onaudioprocess [specs](https://developer.mozilla.org/en-US/docs/Web/Events/audioprocess) define that the returned buffer isn't guaranteed to be valid outside the scope of the event. So my reasoning is that even though using the buffer elsewhere doesn't cause issues on other browsers (at the moment), it is better to follow the spec and do the local copy in all cases.

thijstriemstra:
> it is better to follow the spec and do the local copy in all cases.

if it's working without the workaround in other browsers, I don't see why we'd do all that extra work to follow the spec. If Edge needs it, fine, but I'd rather not have the browser do unnecessary work for the sake of the spec.

mronkain:
> > it is better to follow the spec and do the local copy in all cases.
> 
> if it's working without the workaround in other browsers, I don't see why we'd do all that extra work to follow the spec. If Edge needs it, fine, but I'd rather not have the browser do unnecessary work for the sake of the spec.

Ok, understood. I'll do the changes some time in the next few days.

thijstriemstra:
thanks!

Issue #1459, opened 2018-09-12
Intermittent SyntaxError Thrown on iOS


DrLongGhost:
### Wavesurfer.js version(s):
2.0.5 

### Browser version(s):
Seen with 4 different users on the following devices:

Mobile Safari: Version: 11.0
iOS: Version: 11.4.1, 11.2 and 11.2.5
Devices: iPhone and iPad

### Code needed to reproduce the issue:
Unfortunately this issue is not easily reproducible. It arises from having Wavesurfer deployed in a Production environment where large numbers of people are using it. I am using Wavesurfer in a Vue.js app in which Wavesurfer is completely destroyed and then re-created potentially multiple times without a full page reload. For all browsers except iOS Safari, this works great. However, a small number of users appear to be having intermittent issues rendering Wavesurfer on iOS devices.

I can provide the following information based on the stack traces captured by our error reporting/logging.

Users are seeing the following error:

SyntaxError: The string did not match the expected pattern

From this code:

{snip} ndow.OfflineAudioContext||window.webkitOfflineAudioContext)(1,2,e)),window.WaveSurferOfflineAudioContext}}]),s(t,[{key:"init",value:function {snip}

As shown [here](https://github.com/WebKit/webkit/blob/master/LayoutTests/webaudio/offlineaudiocontext-constructor-expected.txt), the error appears to be caused by a falsey value being passed to the getOfflineAudioContext() method in the webaudio.js file.

Beyond this, I do not know what exactly is causing this false value to be passed in and I cannot replicate the error myself. As a result of a description from a user, the syntax error appears to result in Wavesurfer failing to load the desired audio file.

I do see one possible problem in the code in the spot where getOfflineAudioContext() is called:

this.offlineAc = this.getOfflineAudioContext(
  this.ac ? this.ac.sampleRate : 44100
);

The above code is arguably failing to do proper validation because it is checking that "this.ac" is not falsey but it then proceeds to use "this.ac.sampleRate" without explicitly checking to see whether "this.ac.sampleRate" is falsey. It is possible that with this extra validation put in, it would fallback to the default 44100 sample rate and Wavesurfer would render. It is also possible that there are other issues with the AudioContext and this change would not resolve those other issues, whatever they may be. I may try this hotfix myself and see what the result is.

[...]

PR #1467, opened 2018-09-26
 Made a guard condition more robust in WebAudio

DrLongGhost:
The current code does insufficient checking when applying the default sample rate, which can result in errors being throw in iOS.

None

None

### Todos/Notes:
This PR will hopefully remove a class of failures in iOS but it is also possible that this fix will merely result in recording breaking at a different spot in the code. Even if this happens, the new place where the code breaks may be more illustrative of the underlying problem and/or provide an additional data point to help track it down.

Full details on the iOS issue are in ticket  #1459


DrLongGhost:
I don't know why that build is failing but I assume it's unrelated to my change, based on the tiny scale of my change and content of the logging re: the build failure.

PR #1472, opened 2018-10-01
Fix order of arguments for PluginClass.constructor

higuri:
#### Before
* The PluginClass (base class defined in src/wavesurfer.js) has `constructor(ws, params)`
* \*Plugin (concrete classes defined in src/plugin/*.js) have `constructor(params, ws)`
* The WaveSurfer object instantiates plugins with `new *Plugin(params, ws)`
#### After
* The PluginClass has `constructor(params, ws)`

None

None

### Todos/Notes:
None

None

### DIFF
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -133,10 +133,10 @@ class PluginClass {
     /**
      * Construct the plugin
      *
-     * @param {Object} ws The wavesurfer instance
      * @param {Object} params={} The plugin params (specific to the plugin)
+     * @param {Object} ws The wavesurfer instance
      */
-    constructor(ws, params) {}
+    constructor(params, ws) {}
     /**
      * Initialise the plugin
      *


higuri:
Now I reviewed these changes and I came to think that it is better to change concrete Plugin classes instead of the base Plugin class. I mean `constructor(ws, params)` is better than `constructor(params, ws)`, because `params` is an optional argument that can be omitted.

Is it okay if I change all constructor of concrete Plugin classes and the line that instantiates them.
I'm sorry for this poor PR. 😓

thijstriemstra:
@higuri 

> I came to think that it is better to change concrete Plugin classes instead of the base Plugin class.

I'm not sure, your fix looks ok to me. In the end it's not going to matter much what the var is called, javascript will not treat them differently, no matter what they're called.

higuri:
Then, would you merge the current commit?

The motivation of this PR is that those who make new plugins may implement their constructors which don't work as intended, since the constructor of the plugin base class is defined differently from the actual calling way.
This can be solved with the current commit, so it is okay to keep it.

thijstriemstra:
Thanks!

Issue #1183, opened 2017-08-01
audio.st - prerender peaks via online api,


entonbiba:
To be updated.

The api is available here for testing via browser, you may upload your audio files which will be destroyed once converted. I've limited the file upload to 20 MB for now just to test the waters.
[https://waveform.audio.st/](https://waveform.audio.st/)

**my todo**:
-  [x] create api short url
-  [x] create api website
-  [x] create api service
-  [x] allow wav, mp3 and flac audio types
-  [x] save decoded file data as json
-  [x] save decoded file data as png
-  [x] save decoded file data as svg
-  [ ] save uploaded audio as base64
-  [ ] save uploaded audio as hls chunks
-  [ ] test with free apikeys
-  [ ] create free and pro api keys plans
-  [ ] increase upload size
-  [ ] create plugin to reference waveform generated data
-  [ ] update link in github pages to waveform.audio.st
-  **other features, please comment**
 
when the online api service is released it might fix #392 #1171 #1161 #737 #668 #316 #159 #22 #9 #1082 #488 #497 #1094 #1218 #1348 #2394

Preview:

![image](https://user-images.githubusercontent.com/5193884/32705764-684c99c0-c7e5-11e7-8ede-f177a81471f2.png)

xiaoxiaodexueyin:
great!

matixmatix:
I'm interested! If you're using Ruby, I can help you with that - I have a module that takes a file and makes wavesurfer compatible peaks.

akreal:
Here is also C++ implementation: https://github.com/beschulz/wav2json

entonbiba:
@xiaoxiaodexueyin @matixmatix @akreal I added some features on what the api will offer, any additional features you think would make sense to add?

mspae:
Also you can check this out: https://github.com/bbc/audiowaveform

entonbiba:
will be updating the service soon, if you guys have any suggestions of feature requests let me know.

PR #1505, opened 2018-11-10
add waveform.al to projects links

entonbiba:
add waveform.al api service link

What this service offers:
- allows to create waveform json audio data and images from audio files that can be used with wavesurfer.js
- no audio files are stored on server, audio json data will be available via cdn
- wavesurfer-js.org is already whitelisted with CORS for unlimited use

You can request features here: #1183


katspaugh:
Cool site!

PR #1541, opened 2018-12-22
replace usage of util.ajax with util.fetchFile

### LINKED ISSUE #1365: Replace XMLHttpRequest with fetch API
https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

entonbiba:
I've created a fetch api file in the utils folder. We should leave the ajax.js file as is and introduce a new fetch.js file to support the fetch api. Currently it creates a blob and returns a blob. If needed I can add an option to specify what type it should return being `blob` or `arraybuffer`.

The function is called `dofetch` to avoid conflicts with browser default `fetch` and will check if Fetch API is supported by `self.fetch`.

todo:
- [x] create utils fetch.js file
- [x] check if any other files that need to be updated for Fetch API to work
- [x] add custom return type to support arraybuffer?

The values that can be added to the options will be as follows:
```
{
            method: "GET"),
            requestHeaders: [{
  	      header: {
                 "Content-Type":"application/json"
              }
            }]
            mode: "cors",
            credentials: "same-origin",
            cache: "default"
}
```

fixes #1365 
refs #1100
@katspaugh


src/util/fetch.js
```diff
@@ -0,0 +1,59 @@
...
+            credentials: (options.credentials || "same-origin"),
+            cache: (options.cache || "default")
+        };
+
+        // do the fetch
+        fetch(fetchRequest,fetchOptions).then(function(response) {
+          if (response.ok) {
+            return response.blob();
+          } else{
+            instance.fireEvent('HTTP error status ', response.status)
```
katspaugh: I think here you might need to throw an exception? Otherwise the next `then` will try to run.

entonbiba: I can add it, but currently it will return the `instance.fireEvent` event as it catches the error and it won't go to the next `then`. The error message will say `Failed to fetch` by default if the url cannot be resolved.

src/util/fetch.js
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        // set default fetch options
+        const fetchOptions = {
+            method: (options.method || "GET"),
+            headers: fetchHeaders,
+            mode: (options.mode || "cors"),
+            credentials: (options.credentials || "same-origin"),
+            cache: (options.cache || "default")
+        };
```
thijstriemstra: there's a `xhr` option that should be used here.

src/util/fetch.js
katspaugh: If you don't throw an error in this `then`, it will try to run the next `then` (which will fail because the `blobResponse` will be undefined).

src/util/fetch.js
```diff
@@ -0,0 +1,59 @@
...
+            instance.fireEvent('success', fetchRequest.response, objectURL);
+        }).catch(function(error) {
+            instance.fireEvent('error', error.message)
+        });
+
+        // return the fetch
+        instance.fetchRequest = fetchRequest;
+        return instance;
+    } else{
+        instance.fireEvent('error', 'fetch api not supported');
```
katspaugh: For consistency, it would be nice to return the instance in all cases.
Otherwise, the caller won't be able to receive this error message.

src/util/fetch.js
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * @param {Options} options Description
+ * dofetch({
+ *   url: "https://www.waveform.al/sample.json"
+ * })
+ *
+ * @returns {Object} Observer instance
+ */
+export default function dofetch(options) {
+    if (self.fetch) {
```
katspaugh: How about an early return here?
```
const instance = new Observer();
if (typeof fetch !== 'function') {
  Promise.resolve(null).then(() => {
    instance.fireEvent('error', 'fetch api not supported');
  });
  return instance;
}
```

thijstriemstra:
@entonbiba any updates on this? would be nice to get this into v3.0.0. I don't think we need a lot/any backward compat for fetch to be honest.

katspaugh:
Btw, fetch doesn't throw for error codes like 40x and 50x. One needs to explicitly check for it.
An example (in TypeScript):

```
export async function fetchJson(url: string, params?: any): Promise<IJson> {
  const resp = await fetch(url, params) as Response;
  if (resp.status >= 400) {
    const text = await resp.text();
    throw Error(text);
  }
  return resp.json();
}
```

src/util/fetch.js
thijstriemstra: fixed.

src/util/fetch.js
thijstriemstra: fixed

thijstriemstra:
@entonbiba I completed the implementation, can you take a look? This is the last thing before 3.0.0 can be released IMO.

src/util/fetch.js
```diff
@@ -0,0 +1,196 @@
...
+export default function fetchFile(options) {
+    if (!options) {
+        throw new Error('fetch options missing');
+    }
+    if (!options.url) {
+        throw new Error('fetch url missing');
+    }
+    const instance = new Observer();
+    let fetchHeaders = new Headers();
+    let fetchRequest = new Request(options.url);
```
katspaugh: These two references are never reassigned, so they could be `const`.

src/util/fetch.js
```diff
@@ -0,0 +1,196 @@
...
+            return new Response(
+                new ReadableStream({
+                    start(controller) {
+                        read();
+
+                        /**
+                         * start reading
+                         * @private
+                         */
+                        function read() {
```
katspaugh: Could this function be defined somewhere outside to reduce the nestedness?

karma.conf.js
```diff
@@ -1,11 +1,13 @@
...
 
+const fs = require('fs');
+
 process.env.BABEL_ENV = 'test';
 process.traceDeprecation = true;
 
 require('@babel/register');
-var webpackConfig = require('./build-config/webpack.prod.main.js');
-var ci = process.env.TRAVIS || process.env.APPVEYOR;
+let webpackConfig = require('./build-config/webpack.prod.main.js');
```
chrisparton1991: Looks like `webpackConfig` can be `const`.

spec/util.spec.js
```diff
@@ -88,6 +88,171 @@ describe('util.ajax:', function() {
...
+            expect(instance.response.headers.has('Content-Type')).toBeTrue();
+            expect(instance.response.headers.get('Content-Type')).toEqual(
+                'text/plain'
+            );
+
+            done();
+        });
+    });
+});
+
```
chrisparton1991: All of the `let`s in this file can be changed to `const` too. Not a big deal though.

src/util/fetch.js
```diff
@@ -0,0 +1,196 @@
...
+ *
+ * request.on('error', e => {
+ *     console.warn('fetchFile error: ', e);
+ * });
+ */
+export default function fetchFile(options) {
+    if (!options) {
+        throw new Error('fetch options missing');
+    }
+    if (!options.url) {
```
chrisparton1991: Could save a line by making this `else if (!options.url)`

chrisparton1991:
Looks good. I added a couple of nitpicks, but nothing that's actually a problem 👍

src/util/fetch.js
thijstriemstra: sounds good, I'll probably put it in a separate class or something.

PR #1584, opened 2019-03-13
add backgroundColor option

### LINKED ISSUE #1118: Background Color Parameter

thijstriemstra:
fixes #1118

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 wavesurfer.js changelog
 =======================
 
+2.2.1 (unreleased)
+------------------
+
+- Add `backgroundColor` option (#1118)
+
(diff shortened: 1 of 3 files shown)


thijstriemstra:
This doesn't change the canvas background color but the container color, so `getImage` will still be transparent (which is a good thing IMO, if someone wants to include color in the canvas background we should reconsider a new option, and a new PR).

Or we should ignore this container background color setting and go for canvas background color only. Or make that v2 of this option.

PR #1601, opened 2019-03-27
doc: expand audiowaveform FAQ item

thijstriemstra:
improve documentation


thijstriemstra:
@entonbiba I'm not sure if the correct `audiowaveform` parameters are used in this example, do you know what the best settings are for that `audiowaveform` tool?

aburai:
Best settings depends very much on usage, imho.

PR #1605, opened 2019-03-31
cursor plugin: add showTimeYPosition option

thijstriemstra:


thijstriemstra:
@katspaugh not sure about the option name and option values, any thoughts or is it clear enough?

src/plugin/cursor.js
```diff
@@ -13,6 +13,9 @@
  * @property {boolean} showTime=false Show the time on the cursor.
  * @property {object} customShowTimeStyle An object with custom styles which are
  * applied to the cursor time element.
+ * @property {string} showTimeYPosition='fixed' Use 'relative' to make the time on
```
katspaugh: `showTimeYPosition` sounded like a boolean value until I read the description.
Maybe `followCursorY {boolean}`?

thijstriemstra:
@katspaugh can you adjust the coveralls coverage percentage? So it does not show up as failed when a tiny percentage, e.g. 0.3%, coverage decrease is detected? I think 1 or 1.5% is a better threshold.

katspaugh:
@thijstriemstra I've set it to 1.5% 👌

PR #1616, opened 2019-04-10
Use fake AudioContext when not supported

### LINKED ISSUE #1614: IE11: Fallback to MediaElement not working, it always tries to access AudioContext
README [says](https://github.com/katspaugh/wavesurfer.js/blob/master/README.md#browser-support) that the player will fall back on audio element (without graphics) on browser that don't support WebAudio (i.e. IE 11), but still it doesn't look to work at all if WebAudio is missing.

IE 11, for example, throws an error because wavesurfer tries to access `AudioContext` or `window.webkitAudioContext`, I think from here: <https://github.com/katspaugh/wavesurfer.js/blob/ff5d8b442bc503890068b5a206eaf9142fffbf1f/src/webaudio.js#L76-L77>

You can test the issue just by opening the [audio element example page](https://wavesurfer-js.org/example/audio-element/index.html) on your site in IE 11. Here's the error in the browser console (pardon the Italian locale, the red text says something like "Invalid action for the object"):

byzanz-record

Maybe the fallback mechanisms has been dropped in newer releases? Thanks.

**Wavesurfer.js version(s):** 2.2.1  
**Browser version(s):** Internet Explorer 11

lorenzos:
As [requested here](https://github.com/katspaugh/wavesurfer.js/issues/1614#issuecomment-481666875).

fixes #1614

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@ wavesurfer.js changelog
 - Fix `wavesurfer.isReady`: make it a public boolean, the
   broken `isReady` method is removed (#1597)
 - Add support for `Blob` output type in `wavesurfer.exportImage` (#1610)
+- Fix fallback to Audio Element in browsers that don't support Web Audio (#1614)
 - Cursor plugin:
   - add `formatTimeCallback` option
   - add `followCursorY` option (#1605)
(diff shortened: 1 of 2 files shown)


thijstriemstra:
looks good to me, thanks.

Issue #1600, opened 2019-03-26
Support for OffscreenCanvas


thijstriemstra:
> rendering of `OffscreenCanvas` is fully detached from the DOM and therefore offers some speed improvements over the regular canvas as there is no synchronization between the two.

See https://developers.google.com/web/updates/2018/08/offscreen-canvas and https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas

PR #1617, opened 2019-04-10
refactor: add drawer.canvasentry

thijstriemstra:
This PR adds a `CanvasEntry` class used by the `MultiCanvas` renderer. Previously the renderer used simple objects to store the canvas and rendering context in, this PR creates a dedicated class for these canvases. This way it should be easier to eventually implement `OffscreenCanvas` support (#1600).

Also disabled eslint in webpack, because it's really annoying to fix lint errors while you're working on wavesurfer.js. eslint will run when you try to commit the changes, a much better moment to fix lint errors.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,7 @@ wavesurfer.js changelog
 - Add `wavesurfer.getActivePlugins()`: return map of plugins
   that are currently initialised
 - Update progress when seeking with HTML media controls (#1535)
+- Refactor `MultiCanvas` and add `CanvasEntry` class (#1617)
 - Fix `wavesurfer.isReady`: make it a public boolean, the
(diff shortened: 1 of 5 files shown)


thijstriemstra:
Perhaps this should also be released as a v3.0 since it could break some existing custom renderers.

thijstriemstra:
> Btw, can the original drawer.js potentially also be implemented using an Entry?

not sure, since multicanvas extends drawer.. If I find a way to improve drawer, I'll open another PR.

chrisparton1991:
Sorry I didn't get a chance to look at this before it was merged @thijstriemstra.

It looks great though, definitely a good idea to encapsulate the complexity of managing the canvas layers :+1:

PR #1618, opened 2019-04-13
configure the VS Code Debugger for the development

toolowner:
Suggestion on using VS Code - Debugger for Chrome for the development

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,7 +1,4 @@
 {
-    // Use IntelliSense to learn about possible attributes.
-    // Hover to view descriptions of existing attributes.
-    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
     "version": "0.2.0",
     "configurations": [
         {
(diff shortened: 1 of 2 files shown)


thijstriemstra:
What about adding this json file to the repostitory @toolowner?

toolowner:
Good idea :+1:  I will add it.

I wanted to do that but I was afraid that some people don't prefer that way.

.vscode/launch.json
```diff
@@ -0,0 +1,21 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
```
thijstriemstra: it's weird there are comments in this json file (rendering it invalid), did you add those manually or?

toolowner: That comment was generated by VS Code.

toolowner: I will delete it.

PR #1621, opened 2019-04-17
Added npm prepare script

moritzvieli:
Added a new npm script "prepare" in the package.json, which simply redirects to the build-script. The prepare script is described by npm as follows:

> Run both BEFORE the package is packed and published, on local npm install without any arguments, and when installing git dependencies (See below).
https://docs.npmjs.com/misc/scripts

This is sometimes required, if the dependency is not installed over npm directly but over github.

None

None

### Todos/Notes:
None

None


thijstriemstra:
thanks.

PR #1622, opened 2019-04-17
multicanvas: use intersection observer to reduce canvas drawing

thijstriemstra:
Update `MultiCanvas` to use the Intersection Observer API (https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API).

This allows us to detect when one or more canvases are visible, and only render those canvases when they're visible. When you scroll in a song (and view their canvases), the intersection observer detects this and will render those canvases only. Once a canvas has been rendered it will not be rendered again.

So for example, with a max canvas width of 10,000 px, and heavy zooming on a large track, it could easily create 40 canvases (of 10k px) and render all of those when you issue the zoom command. This is unnecessary and only rendering the canvases visible to the user (the viewport) is enough.

I built it into the `MultiCanvas` renderer but this functionality could also/should be moved to a subclass of `MultiCanvas`.

It's enabled by default, unless the browser does not support the API.

Browser support: https://caniuse.com/#feat=intersectionobserver


src/drawer.multicanvas.js
```diff
@@ -74,12 +74,37 @@ export default class MultiCanvas extends Drawer {
...
+        this.supportsIntersectionObserver = 'IntersectionObserver' in window;
     }
 
     /**
      * Initialize the drawer
      */
     init() {
+        if (this.supportsIntersectionObserver) {
+            // observe canvas entry intersections
+            let intersectionOptions = {
```
katspaugh: I guess this could be a `const`?

src/drawer.multicanvas.js
```diff
@@ -482,6 +522,36 @@ export default class MultiCanvas extends Drawer {
...
 
+    /**
+     * Called when an entry intersection goes above or below threshold.
+     *
+     * @param {array} entries
+     * @private
+     */
+    onIntersectionChange(entries) {
+        entries.forEach(entry => {
+            let canvasEntry = this.getCanvasEntryByElement(entry.target);
```
katspaugh: Same here.

thijstriemstra: im actually mutating canvasEntry one line below that so const won't work here.

src/drawer.multicanvas.js
thijstriemstra: yes, fixed.

thijstriemstra:
@Stoyvo could you give this a try?

Stoyvo:
@thijstriemstra Not seeing minimap working with this change

Stoyvo:
@thijstriemstra This change has brought a degraded performance. Using an mp3 track, ~4 minutes, while zoomed I have noticed additional stutter every few seconds. Current master branch does not have this stutter.

Additionally, while zooming in, there seems to be additional delay. Master branch also does not have this issue.

thijstriemstra:
thanks for the feedback @Stoyvo

katspaugh:
The idea to lazy-render canvases in the viewport is really good, it saves CPU and battery resources and probably results in a more responsive app (if implemented right).
I'll close this PR for inactivity, but I suggest we return to this idea in the next version.

PR #1623, opened 2019-04-18
timeline plugin: add offset parameter

moritzvieli:
Sometimes it's required to adjust the timeline to the audio. For example, the timeline displays musical bars with a specific time signature. However, the audio does not start at 0 seconds, but maybe at 0.3 seconds. This new parameters allows passing an offset in seconds to the timeline plugin to adjust the timeline to the audio. The value may also be negative.

None

None

### Todos/Notes:
What about documentation? If you can point me to a direction, I can check it. 🙂

None


moritzvieli:
@thijstriemstra done. Is it still missing somewhere?

thijstriemstra:
@moritzvieli yea probably here: https://github.com/katspaugh/wavesurfer.js/blob/master/example/timeline/index.html#L96

src/plugin/timeline.js
```diff
@@ -376,7 +379,8 @@ export default class TimelinePlugin {
             this.params.secondaryLabelInterval
         );
 
-        let curPixel = 0;
+        let curPixel =
```
thijstriemstra: you can simply use `this.params.offset` here (like line 376).

moritzvieli: True, thanks! 👍

moritzvieli:
> @moritzvieli yea probably here: https://github.com/katspaugh/wavesurfer.js/blob/master/example/timeline/index.html#L96

@thijstriemstra True, thanks! I have not found any additional places to add the new param. Is it OK now?

kjakobson:
Hi @thijstriemstra, first of all thanks for all the work you guys do! When do you expect release with this PR?

thijstriemstra:
> When do you expect release with this PR?

@kjakobson opened #1651 for that

PR #1624, opened 2019-04-18
Added the whole dist-folder to gitignore

moritzvieli:
Maybe I missed something, but why is the .gitignore that complicated regarding the dist-folder? After building Wavesurfer, I always have a dirty working tree with some .js and .min.js in the dist-folder. If it makes sense the way it is, maybe it's worth adding a comment to the .gitignore file to avoid confusing future contributors. 😜

None

None

### Todos/Notes:
None

None


thijstriemstra:
it's related to npm. for my projects i use something like https://github.com/collab-project/videojs-record/blob/master/.npmignore where wavesurfer uses https://github.com/katspaugh/wavesurfer.js/blob/master/.npmignore

I prefer the 'blank' npmignore approach, e.g.

```
# Intentionally left blank, so that npm does not ignore anything by default,
# but relies on the package.json "files" array to explicitly define what ends
# up in the package.
```

moritzvieli:
@thijstriemstra Sorry, I don't get it. .npmignore should be used to ignore files for NPM deployment, .gitignore should be used to ignore files in the git repository. The dist-folder should not be pushed to the repo. Do you agree? Therefore I'd suggest to add the whole folder to .gitignore. This should have no impact in NPM.

thijstriemstra:
@moritzvieli like I said, we want to use a blank npmignore and use the `files` array in package.json to specify what files are included. For example: https://github.com/collab-project/videojs-record/blob/master/package.json#L44

moritzvieli:
@thijstriemstra Maybe I'm too dumb to understand... but how is .npmignore related to .gitignore? I'm only talking about .gitignore. After each build, my working tree is dirty, like this:

image

I have to delete the files manually each time and this is obviously not correct.

> So basically double-check if npmignore is empty and whether the dist folder is included in files in package.json.

.npmignore is not empty, but it's not either on the master here on github. Additionally, there is no files array in package.json. But as said, I don't get, how this is related to the problems caused by .gitignore.

thijstriemstra:
It's confusing for sure, I made the changes in master. Thanks for bringing this up.

moritzvieli:
@thijstriemstra OK, perfect! Thanks a lot! 👍

PR #1632, opened 2019-04-25
regions plugin: snap to grid options

moritzvieli:
There are use cases, where I want my regions to be able to snap to a grid. E.g. when I am editing a file with a fixed time signature. This PR introduces two new parameters for the regions plugin: "snapToGridInterval" and "snapToGridOffset". They both behave the same as the parameters "timeInterval" and "offset" in the timeline plugin.

None

None

### Todos/Notes:
I don't know, whether it is a good idea to add the new parameters to the example, because it makes probably no sense there. But I can do it, if you want.

None


src/plugin/regions.js
```diff
@@ -816,9 +825,16 @@ export default class RegionsPlugin {
             }
 
             const end = this.wavesurfer.drawer.handleEvent(e);
+            const startUpdate = this.wavesurfer.regions.util.getRegionSnapToGridValue(
+                start * duration
+            );
+            const endUpdate = this.wavesurfer.regions.util.getRegionSnapToGridValue(
+                end * duration
+            );
+            console.log(startUpdate);
```
thijstriemstra: no console statements

src/plugin/regions.js
```diff
@@ -867,4 +883,25 @@ export default class RegionsPlugin {
 
         return min;
     }
+
+    /**
+     * Get the snapped to grid value.
+     *
+     * @param {*} value the value to snap to the grid, if needed
+     * @param {*} params the regions plugin params
+     */
```
thijstriemstra: im sure there's a better type than `*` to document for these params.

thijstriemstra:
And can you also this to the changelog?

src/plugin/regions.js
moritzvieli: Like this?

moritzvieli:
> And can you also this to the changelog?

@thijstriemstra Thanks a lot for your input! I fixed the mentioned things. Please let me know, if there is anything else required.

CHANGES.md
```diff
@@ -17,8 +17,9 @@ wavesurfer.js changelog
   - add `followCursorY` option (#1605)
 - Spectrogram plugin: fix `ready` listener when loading multiple
   audio files (#1572)
-- Regions plugin: fix `ready` listener when loading multiple
-  audio files (#1602)
+- Regions plugin:
+  - fix `ready` listener when loading multiple audio files (#1602)
+  - add `snapToGridInterval` and `snapToGridOffset` options
```
thijstriemstra: include PR nr

src/plugin/regions.js
```diff
@@ -319,8 +319,9 @@ class Region {
                     );
 
                     // Update time
-                    const time =
-                        this.wavesurfer.drawer.handleEvent(e) * duration;
+                    const time = this.wavesurfer.regions.util.getRegionSnapToGridValue(
```
thijstriemstra: add comment what value `time` is expected to be

src/plugin/regions.js
```diff
@@ -342,8 +343,9 @@ class Region {
                         : null;
 
                     e.stopPropagation();
-                    startTime =
-                        this.wavesurfer.drawer.handleEvent(e, true) * duration;
+                    startTime = this.wavesurfer.regions.util.getRegionSnapToGridValue(
```
thijstriemstra: add comment what value `startTime` is expected to be

src/plugin/regions.js
```diff
@@ -622,6 +628,9 @@ export default class RegionsPlugin {
         this.params = params;
         this.wavesurfer = ws;
         this.util = ws.util;
+        this.util.getRegionSnapToGridValue = value => {
```
thijstriemstra: why put this in `util`?

src/plugin/regions.js
```diff
@@ -867,4 +882,25 @@ export default class RegionsPlugin {
 
         return min;
     }
+
+    /**
+     * Get the snapped to grid value.
```
thijstriemstra: please expand this sentence, not clear to me what this does

src/plugin/regions.js
moritzvieli: I need access to it from the Region and from the RegionsPlugin. Is there a better place?

src/plugin/regions.js
moritzvieli: startTime has not been introduced with this PR. I tried to explain it better.

src/plugin/regions.js
moritzvieli: time has not been introduced with this PR. I tried to explain it better.

thijstriemstra:
@moritzvieli this branch has conflicts that must be resolved first.

moritzvieli:
@thijstriemstra The conflicts have been resolved.

swampthang:
I'm hoping this is something that will allow me to set an incremental snap value so that when a region's handle is dragged or the entire region is dragged, it will snap to those values. However, I haven't found how to set this up. Can someone point me to or create an example of how to set this up? Really would appreciate it!

PR #1633, opened 2019-04-25
regions plugin: allow drawing new regions over old ones, when the old ones are not resizable and not draggable

moritzvieli:
When a region is not resizable and not draggable, I don't want to stop the event propagation, because the event is not handled. In this case, it makes it impossible to draw new regions over (maybe fixed background regions). With this PR, the event propagation only stops, if the region is resizable or draggable and actually handles the event.

None

None

### Todos/Notes:
None

None


src/plugin/regions.js
```diff
@@ -341,7 +341,9 @@ class Region {
                         ? e.targetTouches[0].identifier
                         : null;
 
-                    e.stopPropagation();
+                    if (this.drag || this.resize) {
```
thijstriemstra: add comment why you put in this if statement

moritzvieli:
@thijstriemstra it's done.

PR #1635, opened 2019-04-28
mediaelement: fix volume updates

### LINKED ISSUE #1615: MediaElement mute not updated
When you unmute the volume of an existing `<audio>` element, `wavesurfer.isMuted` is still `true`.

thijstriemstra:
fixes #1615

### DIFF
--- a/src/mediaelement.js
+++ b/src/mediaelement.js
@@ -38,6 +38,8 @@ export default class MediaElement extends WebAudio {
         /** @private */
         this.volume = 1;
         /** @private */
+        this.isMuted = false;
+        /** @private */
         this.buffer = null;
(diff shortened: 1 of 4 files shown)


src/mediaelement.js
```diff
@@ -337,7 +346,7 @@ export default class MediaElement extends WebAudio {
      * @return {number} value A floating point value between 0 and 1.
      */
     getVolume() {
-        return this.volume || this.media.volume;
+        return this.volume;
```
thijstriemstra: this was a bugfix actually because it would never return `this.volume = 0` (resolves to `false`, unless `this.media.volume` was coincidentally 0 as well)

PR #1640, opened 2019-05-02
Dev flexibility

Stoyvo:
dotenv - Allow customizing build output directory, useful for larger applications that utilize this library
watch - Watch the src directory for changes, trigger the specific build command
build:main:plugins - Build library without htmlinit, for applications not using it

N/A

N/A

### Todos/Notes:
N/A

N/A

### DIFF
--- a/.env
+++ b/.env
@@ -1,1 +1,1 @@
-OUTPUT_DIR="dist"
+OUTPUT_DIR="../../wavesurfer"
--- a/.env
+++ b/.env
@@ -1,1 +1,1 @@
-OUTPUT_DIR="../../wavesurfer"
+OUTPUT_DIR="dist"
--- a/.env
+++ b/.env
@@ -1,0 +1,1 @@
+OUTPUT_DIR="dist"
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@
 .chrome
 .build_cache
 .pydevproject
+.env
 
 package-lock.json
 .package.json.swp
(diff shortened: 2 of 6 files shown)


Stoyvo:
> I'd rather not use dotenv and keep things simple. Instead of specifying `dist` with dotenv, why not move it to the `common` build file and define `dist` there so it can be reused.

The goal is to simply have a "configuration" where the library provides a default, but changes to this config are untracked in git. Other than a defined configuration file in the root, no other ideas came to mind.

> webpack-dev-server already watches and rebuilds files. Why not improve that instead of adding additional commands in npm?
I'll take a look into utilizing this a bit more. Was trying to avoid the unnecessary open port. I can remove the dependancy of `watch` and implement https://webpack.js.org/configuration/watch#watch

thijstriemstra:
> I'll take a look into utilizing this a bit more. Was trying to avoid the unnecessary open port. I can remove the dependancy of watch and implement https://webpack.js.org/configuration/watch#watch

Sounds good.

thijstriemstra:
ping @Stoyvo

Stoyvo:
@thijstriemstra haven't forgotten, just overwhelmed at the moment. Will have an update hopefully this week or weekend

thijstriemstra:
thanks for the update. we want to release 3.0 soon (#1651) and would be nice to get this in there as well.

thijstriemstra:
it's now 2 months later and no update, so closing this. Please re-open with changes or open a new PR.

PR #1654, opened 2019-05-22
spectrogram plugin: allow user to specify a colorMap

### LINKED ISSUE #1436: Colour the spectrogram
I want to color the language spectrum, so it might look better, right?I don't know where to start. I hope you can help me

shershey:
Hi,

The following change is a proposal for how to allow users to specify a colormap to be used when drawing a spectrogram. This fixes #1436 .

The spectrogram.js file has been modified to support a colorMap property on the params.  If unspecified, the spectrogram will be drawn as a black and white image as before. If specified, the colorMap is expected to be a 2D array with 256 entries with rgb and alpha values between 0 and 1. 

This shouldn't break the external API.

This shouldn't break the internal API.

### Todos/Notes:
I'm not a colormap expert but it seems some colormap libraries provide values between 0 and 1 and others between 0 and 255. I chose between 0 and 1 since there appears to be a  [colormap](https://www.npmjs.com/package/colormap) javascript library that provides values between 0 and 1. If anyone has opinions on this, let me know.

I didn't see any spectrogram tests that I should update.

I considered modifying the spectrogram example to include the "hot" colormap, but this would either require:
a) Including the colormap npm
b) Hard coding the "hot" colormap as an array of arrays.

bug #1436


shershey:
thijstriemstra, is that last change what you were recommending? I modified the spectrogram/app.js file to use the utils.ajax command to load a new hot-colormap.json file and pass the resulting array of arrays to the spectrogram.

thijstriemstra:
looks good. how did you generate the (json) colormap? documenting this is essential.

shershey:
I generated it using python/matplotlib/color maps. What’s the correct place to document that?  Should I paste the code as a comment here?  Or include those few lines of python code as a comment in the spectrogram/app.js file?

thijstriemstra:
it's great you used python, but a npm/js method would be ideal. if you don't have that, we can document the python method. I would add it to the FAQ and link it in the example's html and js. The faq can be edited in the `gh-pages` branch: https://github.com/katspaugh/wavesurfer.js/blob/gh-pages/faq/index.html

example/spectrogram/app.js
```diff
@@ -51,4 +51,13 @@ document.addEventListener('DOMContentLoaded', function() {
     })();
 
     wavesurfer.load('../media/demo.wav');
+}
+
+// Init & load
+document.addEventListener('DOMContentLoaded', function() {
+    WaveSurfer.util
```
thijstriemstra: add comment what's being loaded

jamesshannon:
Thanks for doing this.

Looks like you can also cwreate the JSON using the linked colormap module:

````
var colormap = require("colormap")

let colors = colormap({
    colormap: 'jet',
    nshades: 255,
    format: 'float',
    alpha: 1
})
````
returns
image

and then create the JSON from there. FWIW, I'm not a huge fan of JSON due to the size and parsing overhead, but it's probably the most practical.

**Comment by thijstriemstra** [comment]
> but it's probably the most practical.

yes, for this example I think it is.

File: src/plugin/spectrogram.js [hunk 2]
```diff
@@ -310,7 +313,24 @@ export default class SpectrogramPlugin {
             if (!this.container) {
                 throw Error('No container for WaveSurfer spectrogram');
             }
-
+            if (params.colorMap) {
+                if (params.colorMap.length < 256) {
+                    throw Error('Colormap must contain 256 elements.');
```
thijstriemstra: throw **new** Error

CHANGES.md
```diff
@@ -20,8 +20,9 @@ wavesurfer.js changelog
   - Add `formatTimeCallback` option
   - Add `followCursorY` option (#1605)
   - Remove deprecated `enableCursor` method (#1646)
-- Spectrogram plugin: fix `ready` listener when loading multiple
-  audio files (#1572)
+- Spectrogram plugin:
+  - fix `ready` listener when loading multiple audio files (#1572)
+  - Allow user to specify a colorMap (#1654)
```
thijstriemstra: use the original ticket nr

shershey:
I used @jamesshannon's idea to generate the colormap, regenerated it, and documented it in spectrogram/index.html. Given that I was verbose in the spectrogram/index.html, I'm assuming I don't need to edit the FAQ.

src/plugin/spectrogram.js
shershey: I made the requested change.

example/spectrogram/app.js
shershey: I added a comment.  Not sure if that was what you were looking for.

CHANGES.md
```diff
@@ -20,8 +20,9 @@ wavesurfer.js changelog
   - Add `formatTimeCallback` option
   - Add `followCursorY` option (#1605)
   - Remove deprecated `enableCursor` method (#1646)
-- Spectrogram plugin: fix `ready` listener when loading multiple
-  audio files (#1572)
+- Spectrogram plugin:
+  - fix `ready` listener when loading multiple audio files (#1572)
```
thijstriemstra: fix should be uppercase

example/spectrogram/app.js
```diff
@@ -14,7 +14,8 @@ document.addEventListener('DOMContentLoaded', function() {
         plugins: [
             WaveSurfer.spectrogram.create({
                 container: '#wave-spectrogram',
-                labels: true
+                labels: true,
+                colorMap
```
thijstriemstra: this seems to be a typo, and should be `colorMap: colorMap`

example/spectrogram/index.html
```diff
@@ -98,7 +98,26 @@ <h4>Options</h4>
                       <li><code>fftSamples</code> - number of FFT samples (<code>512</code> by default). Number of spectral lines and height of the spectrogram will be a half of this parameter.</li>
                       <li><code>frequenciesDataUrl</code> - URL to load spectral data from.</li>
                       <li><code>labels</code> - Whether or not to display frequency labels.</li>
+                      <li><code>colorMap</code> - Specifies the colormap to be used when rendering the spectrogram.</li>
```
thijstriemstra: use 'a' instead of 'the' and use 'optional' somewhere

CHANGES.md
shershey: You mean the original bug? (#1436 )

shershey: Done

CHANGES.md
shershey: This should be fixed now.

example/spectrogram/app.js
shershey: Ah. I'm used to typescript where that would be valid. It actually did work in chrome but I see that it is not guaranteed to work for all versions of javascript.  Fixed.

example/spectrogram/index.html
shershey: I originally intended to and then noticed that the author did the opposite.  They specify "required" for all required parameters and don't add "optional" to the optional parameters.  I was trying to be consistent.  I could:
a) leave as is.
b) add optional to this one param.
c) add optional to all params.
Preference?

shershey:
I believe I've addressed all the current comments.

example/spectrogram/index.html
thijstriemstra: it's fine like this then.

example/spectrogram/index.html
```diff
@@ -98,7 +98,25 @@ <h4>Options</h4>
...
                     </ul>
+                    
+                    <h4>Generating a colorMap</h4>
+                    The <a href="https://www.npmjs.com/package/colormap">colormap npm</a> can be used to generate a
+                     colormap. In this example, the colormap has been saved to a json file and is retrieved by the
+                     client before initializing the spectrogram.
+                    
+                    <noindex><p>
+<pre><code>colormap = require('colormap');
+colors = colormap({
```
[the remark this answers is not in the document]
shershey: Done.

shershey:
Excellent! I see the changes are approved. I haven't actually done a pull request before. What are the next steps and is there anything else I need to do?

PR #1663, opened 2019-06-09
cursor plugin: hide the cursor elements before first mouseover if hideOnBlur is set

xtravar:
Both the cursor and the time element are visible before the first hover, but then hide when cursor moves away if hideOnBlur is true. The most obvious manifestation of this bug is a small 4x4 square on the left upon load. (This may be because I'm loading peak data before the audio and it's not noticeable otherwise.)

This change ensures the elements are hidden and require a mouse enter event if hideOnBlur is true.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,6 +21,7 @@ wavesurfer.js changelog
   - Add `formatTimeCallback` option
   - Add `followCursorY` option (#1605)
   - Remove deprecated `enableCursor` method (#1646)
+  - Hide the cursor elements before first mouseover if `hideOnBlur` is set (#1663)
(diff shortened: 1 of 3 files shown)


thijstriemstra:
thanks @xtravar. is this also an issue when not using the hideOnBlur option?

xtravar:
Yup. Once the cursor moves in and out, it's fine. It's just the initial load. In both cases, once the mouse enters, the duration is updated and the label is fine. The hideOnBlur option simply also has that other lesser noticeable artifact.

thijstriemstra:
@xtravar can you also take a look at that issue?

PR #1673, opened 2019-06-17
regions plugin: calculate the duration at event time to allow predefined regions to be dragged and resized

### LINKED ISSUE #1669: Cannot Resize Regions added during .create
I'm looking to utilize Wavesurfer and the `Regions` plugin to create an initial region, and then allow the user to resize it.  I'm able to create the region, and it appears on the instance as expected:

wavesurfer

I also explicitly set `resize` and `drag` to true on creation:

```
      container: '#waveform',
      backend: 'MediaElement',
      waveColor: 'violet',
      hideScrollbar: true,
      autoCenter: false,
      progressColor: 'purple',
      plugins: [
        WaveSurfer.timeline.create({
          container: '#wave-timeline'
        }),
        WaveSurfer.regions.create({
          regions: [
            {
              id: 0,
              start: self.currentTime,
              end: self.currentTime + 15,
              color: 'hsla(400, 100%, 30%, 0.5)',
              resize: true,
              drag: true
            }
          ]
        })
      ]
    })
```

However, **I'm unable to resize the created region**.  If I enable `dragSelection` in the region options, I can create new regions and resize them, but not initially created regions.

Finally, if I log the created region:
```console.log(wavesurfer.regions.list[0])```

The region indeed reports `drag` and `resize` as being set to true:

```
**attributes: {}
color: "hsla(400, 100%, 30%, 0.5)"
data: {}
drag: true
element: region.wavesurfer-region
end: 15.603342
firedIn: false
firedOut: false
handlers: {remove: Array(3), out: Array(1)}
id: 0
loop: false
maxLength: undefined
minLength: undefined
resize: true
scroll: false
scrollSpeed: 1
scrollThreshold: 10
start: 0.603342
style: ƒ (e,t)
util: {…}
wavesurfer: t {handlers: {…}, defaultParams: {…}, backends: {…}, util: {…}, params: {…}, …}
wrapper: wave
_onRedraw: ƒ ()
__proto__: Object
```

What am I missing? Some of the examples do indeed allow for created regions to be resized.

iamsteadman:
Not sure if this is present in all circumstances, but it appeared that predefined regions (those defined when instantiating a `WaveSurfer` instance) couldn't be dragged or resized. I tracked it down to the regions' mouse events, and found that the waveform duration was only being calculated once, essentially upon creation of the region, which may have been happening before the media was fully loaded. That's why user-created regions could be dragged and resized, because they were created once the duration had been obtained.

This PR simply moves the obtaining of the duration to the relevant events. I understand there might be a performance issue, so happy to be tackled on this. I'm not an expert in the library by any means, so please accept my naive PR in good faith :)

fixes #1669


iamsteadman:
Thanks for sorting this, @thijstriemstra. Had a day off today so hadn’t got to this 🙂

PR #1694, opened 2019-07-14
update site for wavesurfer.js 3.0.0

thijstriemstra:
@katspaugh the update instructions for the site were incorrect, I fixed that and updated the site in this PR.


api/file/src/mediaelement.js.html
```diff
@@ -383,7 +404,7 @@
      * @param {number} length How many subranges to break the waveform into.
      * @param {number} first First sample in the required range.
      * @param {number} last Last sample in the required range.
-     * @return {number[]|number[][]} Array of 2*&lt;length&gt; peaks or array of
+     * @return {number[]|Number.&lt;Array[]&gt;} Array of 2*&lt;length&gt; peaks or array of
```
chrisparton1991: If I'm reading this correctly, the type is `Number.<Array[]>`. I'm not familiar with that typing syntax, is it correct?

chrisparton1991:
Looks good from what I can see, though it's a pretty giant PR and I might have missed something.

api/file/src/mediaelement.js.html
thijstriemstra: yes that's their doc syntax for a multi dimensional number array. annoying but cest la vie

thijstriemstra:
> though it's a pretty giant PR and I might have missed something.

yea it's insane when the api docs etc get re-generated but without an approved review I can't push this to the live site. So thanks for your review.

PR #1702, opened 2019-07-26
add autoCenter option

### LINKED ISSUE #1699: Ability to configure auto center feature
### Wavesurfer.js version(s):
2.1

Firefox and Chrome, Fedora

{autoCenter = true}

With autoCenter enabled, when seeking to left side of waveform, cursor only centers very slowly.

Is there a way or an option to center the cursor immediately on playback? I tried it with calling the seekAndCenter function at the seek Event, but of course this only leads to an infinite calling of the seek event.

zapfdk:
Added parameters in options to change the rate of centering if active and added option to immediately center instead of gradual centering.

### Todos/Notes:
- Rate at which the waveform is centered is now dependent from viewWidth, length of waveform and the factor specified in the options
- For very long files, the scrolling event is still triggered too infrequently so the cursor slowly stutters forward while centering. Any idea for triggering the event more frequently in this case?

Fixes #1699

### DIFF
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -203,8 +203,13 @@ export default class Drawer extends util.Observer {
 
         // if the cursor is currently visible...
         if (!immediate && -half <= offset && offset < half) {
-            // we'll limit the "re-center" rate.
-            const rate = 5;
+            // set rate at which waveform is centered
+            let rate = this.params.autoCenteringRate;
+
+            // make rate dependable on on width of viewwidth and length of waveform
+            rate /= half;
+            rate *= maxScroll;
+
             offset = Math.max(-rate, Math.min(rate, offset));
             target = scrollLeft + offset;
         }
@@ -314,7 +319,10 @@ export default class Drawer extends util.Observer {
 
             if (this.params.scrollParent && this.params.autoCenter) {
                 const newPos = ~~(this.wrapper.scrollWidth * progress);
-                this.recenterOnPosition(newPos);
+                this.recenterOnPosition(
+                    newPos,
+                    this.params.autoCenterImmediately
+                );
             }
 
             this.updateProgress(pos);
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -206,7 +206,7 @@ export default class Drawer extends util.Observer {
             // set rate at which waveform is centered
             let rate = this.params.autoCenteringRate;
 
-            // make rate dependable on on width of viewwidth and length of waveform
+            // make rate depend on on width of viewwidth and length of waveform
             rate /= half;
             rate *= maxScroll;
 
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -204,7 +204,7 @@ export default class Drawer extends util.Observer {
         // if the cursor is currently visible...
         if (!immediate && -half <= offset && offset < half) {
             // set rate at which waveform is centered
-            let rate = this.params.autoCenteringRate;
+            let rate = this.params.autoCenterRate;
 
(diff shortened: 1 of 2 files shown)


thijstriemstra:
thanks for your pr.

> For very long files, the scrolling event is still triggered too infrequently so the cursor slowly stutters forward while centering. Any idea for triggering the event more frequently in this case?

how/where is it triggered?

zapfdk:
> how/where is it triggered?
I'm still not completely familiar with the code, but from what I read I think the recentering should be triggered on every `audioprocess` event, but it only seems to be triggered on every scroll event.

src/drawer.js
```diff
@@ -203,8 +203,13 @@ export default class Drawer extends util.Observer {
 
         // if the cursor is currently visible...
         if (!immediate && -half <= offset && offset < half) {
-            // we'll limit the "re-center" rate.
-            const rate = 5;
+            // set rate at which waveform is centered
+            let rate = this.params.autoCenteringRate;
+
+            // make rate dependable on on width of viewwidth and length of waveform
```
thijstriemstra: typo, use `depend on`

src/wavesurfer.js
```diff
@@ -26,6 +26,10 @@ import PeakCache from './peakcache';
  * initialized ScriptProcessorNode or leave blank.
  * @property {boolean} autoCenter=true If a scrollbar is present, center the
  * waveform around the progress
+ * @property {number} autoCenteringRate=5 If autoCenter is active, rate at which the
+ * waveform is centered
+ * @property {boolean} autoCenterImmediately=false If autoCenter is active, immediately
+ * center waveform around progress
```
thijstriemstra: not sure what 'around progess' means. either re-word or remove that part.

src/wavesurfer.js
```diff
@@ -204,6 +208,8 @@ export default class WaveSurfer extends util.Observer {
         audioScriptProcessor: null,
         audioRate: 1,
         autoCenter: true,
+        autoCenteringRate: 5,
```
thijstriemstra: I would call it `autoCenterRate`, to make it similar to the other autoCenter options.

thijstriemstra:
Also see #1706 (toggle autocenter).

src/wavesurfer.js
zapfdk: I just took over the wording from the original 'autoCenter' option, but I can change that to make it clearer.

thijstriemstra: ok. you changed the docs but not the actual code as well.

thijstriemstra:
ping  @zapfdk

zapfdk:
> ping @zapfdk

Sorry, wasn't near my laptop for a week. 
The names should be fixed now.

dhandarbeit-tk:
Could you please pull this? Thanks you very much

thijstriemstra:
@dhandarbeit-tk i still need to test this but am on a holiday. perhaps @katspaugh or @chrisparton1991 can take a look?

src/drawer.js
```diff
@@ -203,8 +203,13 @@ export default class Drawer extends util.Observer {
 
         // if the cursor is currently visible...
         if (!immediate && -half <= offset && offset < half) {
-            // we'll limit the "re-center" rate.
-            const rate = 5;
+            // set rate at which waveform is centered
+            let rate = this.params.autoCenterRate;
+
+            // make rate depend on on width of viewwidth and length of waveform
```
katspaugh: Oh, the typo is still there.

PR #1716, opened 2019-08-02
regions plugin: use isResizing and isDragging to filter events in region-updated listener

### LINKED ISSUE #1713: How to call onDrag & onResize function in wavesurefer regions ?
Hi all

I need regions `start and end `value for **region resize**.

How to call onDrag & onResize function in wavesurefer regions ?

REF : https://github.com/katspaugh/wavesurfer.js/issues/1555

cc - @katspaugh @MaxwellGover

aburai:
fixes #1713

add regions props isResizing and isDragging to filter events in region-updated listener

### DIFF
--- a/src/plugin/regions.js
+++ b/src/plugin/regions.js
@@ -23,8 +23,9 @@ class Region {
                 : Number(params.end);
         this.resize =
             params.resize === undefined ? true : Boolean(params.resize);
-        this.isResizing = false;
         this.drag = params.drag === undefined ? true : Boolean(params.drag);
+        // reflect resize and drag state of region for region-updated listener
+        this.isResizing = false;
         this.isDragging = false;
         this.loop = Boolean(params.loop);
         this.color = params.color || 'rgba(0, 0, 0, 0.1)';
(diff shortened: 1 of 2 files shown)


src/plugin/regions.js
```diff
@@ -23,7 +23,9 @@ class Region {
                 : Number(params.end);
         this.resize =
             params.resize === undefined ? true : Boolean(params.resize);
+        this.isResizing = false;
         this.drag = params.drag === undefined ? true : Boolean(params.drag);
+        this.isDragging = false;
         this.loop = Boolean(params.loop);
```
thijstriemstra: can you add some comment for both the isResizing and isDragging (here or elsewhere)?

src/plugin/regions.js
```diff
@@ -121,7 +123,6 @@ class Region {
             );
         }
 
-        const width = this.wrapper.scrollWidth;
```
thijstriemstra: why remove this?

aburai: unused variable

src/plugin/regions.js
aburai: ok, i will do this. hope this solution works for the issue opener

PR #1742, opened 2019-09-13
spectrogram plugin: fix variable name in click handler

taironemagalhaes:
I noticed that when handling click events performed in the spectrogram, the click position/progress was always evaluating to 0. After some investigation, I realized that the property `this.scrollWidth` did not exist in the context it was being used (spectrogram.js, line 433), and I believe the correct property name is actually `this.width`, since after that modification the position value was aparently evaluating correctly.


thijstriemstra:
thanks @taironemagalhaes, looks good to me (I can't find `scrollWidth` var either).

Issue #354, opened 2015-02-15
Timeline plugin drawn incorrectly on MacBook retina display 


mahtar:
Only half of the timeline is drawn on MacBook retina display.
Also the canvas is upscaled so it gets blurry.

I edited line 99 to get the whole timeline drawn, from this:

    var width = Math.round(this.drawer.wrapper.scrollWidth / this.wavesurfer.params.pixelRatio);

to this:

    var width = Math.round(this.drawer.wrapper.scrollWidth / this.wavesurfer.params.pixelRatio) * window.devicePixelRatio;

I didn't have time to make the canvas sharp on retina but I'm assuming all the width and height calculations need to take pixel ratio into consideration.

thijstriemstra:
Can you provide a screenshot, I wonder what this looks like..

mahtar:
Here you go
![wavesurfer js timeline plugin](https://cloud.githubusercontent.com/assets/483026/6230231/3fbcd316-b6be-11e4-99d2-2448a4298d75.png)

thijstriemstra:
Right. It looks the same on my Samsung Galaxy S4 mini with Firefox.

mahtar:
I'm guessing it will be the same on all high DPI screens where pixels are doubled by browsers for lower resolution images.

katspaugh:
Should be resolved now. Please reopen, if I missed something.

PR #1743, opened 2019-09-13
minimap plugin: fix left/width calculations for regions on retina/4k screens

martynasj:
When drawing regions in the minimap plugin, the position calculated is incorrect on 4k or retina screens

None

None

https://github.com/katspaugh/wavesurfer.js/issues/354


thijstriemstra:
thanks @martynasj, looks good to me.

PR #1744, opened 2019-09-14
a workaround to fix "playLoop will not play a region on clips with duration <15s" #1626

### LINKED ISSUE #1626: playLoop will not play a region on clips with duration <15s
Create one clip 5 seconds long, select a region and use  wavesurfer.regions.list[e.id].playLoop(); 

Outcome: File plays from the start.

Create one clip 16 seconds long, select a region and use  wavesurfer.regions.list[e.id].playLoop();  

Outcome: Clip plays the dragged region loop

huanle0610:
When judge the end of region playing, use approximate value.
Lines of code,  should no other effects.

fixes #1626

### DIFF
--- a/src/plugin/regions.js
+++ b/src/plugin/regions.js
@@ -230,7 +230,7 @@ class Region {
     bindInOut() {
         this.firedIn = false;
         this.firedOut = false;
-        const approx = (a, b) =>  (Math.abs(a -b) < 1e-1);
+        const approx = (a, b) =>  (Math.abs(a -b) < 1e-2);
         const onProcess = time => {
             if (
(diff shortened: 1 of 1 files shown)


src/plugin/regions.js
```diff
@@ -230,13 +230,14 @@ class Region {
     bindInOut() {
         this.firedIn = false;
         this.firedOut = false;
-
+        const threshold = 5e-2;
```
thijstriemstra: @huanle0610 can you add a comment for the `treshold` var, explaining why it's that specific value?

thijstriemstra:
closing in favor of #1746

PR #1746, opened 2019-09-20
regions plugin: fix playLoop and loop option

### LINKED ISSUE #1626: playLoop will not play a region on clips with duration <15s

adjkldd:
fixes #1626


thijstriemstra:
thanks @adjkldd. Can you also add a description for these changes in this pull request?

liujinliang01:
Could anybody kindly help with region play() and playLoop() functions please?
The recently created region does not return the right start and end values for. I am using Firefox on Win10 professional.
My code is as follows:
	wavesurfer.on('region-created', function (region) {
		/*the following will clear all other regions, 
		leaving only the most recently created one
		which is exactly what I want. I don't need more than one regions.
		*/
		wavesurfer.regions.clear();		
		id = region.id;
		alert(id) // upto now it is OK, meaning I can get the id of the recently created region
		
		/*but the following calls don't get the right results: 
		region.start returns 0, i.e. the begining of the whole file 
		rather than the beginning of the region 
		and region.end returns the end of the whole audio file, 
		which are not right.*/
		
		alert(region.start)
		alert(region.end)
		
		region.playLoop(); // so, this will play from the begining till the end of the whole file
		
		region.play(); // this also will play from the begining till the end of the whole file
		
		//wavesurfer.play(region.start,region.end);
	});

PR #1756, opened 2019-09-30
plugins: restore elan-wave-segment plugin

### LINKED ISSUE #1745: Plugin elan-wave-segment.js is missing
### Wavesurfer.js version(s):
3.0.0

Chrome 77

Run example `example/elan-wave-segment`

image

thijstriemstra:
fixes #1745

### DIFF
--- a/example/elan-wave-segment/app.js
+++ b/example/elan-wave-segment/app.js
@@ -1,10 +1,4 @@
-var wavesurfer;
-
-// Create elan instance
-//var elan = Object.create(WaveSurfer.ELAN);
-
-// Create Elan Wave Segment instance
-//var elanWaveSegment = Object.create(WaveSurfer.ELANWaveSegment);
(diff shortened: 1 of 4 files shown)


katspaugh:
@thijstriemstra is this blocked in some way?

A general note: I think the elan plugin and this segment plugin are both cool little apps in their own right, and they don't have to be plugins. They aren't really modifying wavesurfer.js in any way and can be built on top of it as regular apps.
Perhaps moving them to another repo would be a good idea.

thijstriemstra:
> is this blocked in some way?

yea; the actual work of bringing it back is blocking it ;)

katspaugh:
> yea; the actual work of bringing it back is blocking it ;)

Can you elaborate? Does the plugin code need to be updated because of the latest updates in wavesurfer.js?

If you're not planning to work on it, let's close the PR.

PR #1760, opened 2019-10-03
Waveform with rounded bars

### LINKED ISSUE #953: Waveform with rounded bars
screen shot 2017-02-01 at 13 47 44

Hey, I'm working on a project and the design for the waveform is fairly custom, each bar having border-radius/rounded ends and the gaps being larger than that of the demo's.

How would I go about doing this?

Thanks,

kubk:
Added an optional `barRadius` parameter for specifying the radius of rounded bars. Added `drawRoundedRect` method for rendering rounded bars on Canvas.

No BC break.

No BC break.

fixes #953 

### Notes
If this will be merged I will document this feature in `gh-pages` branch

### Example
image

### DIFF
--- a/src/drawer.canvasentry.js
+++ b/src/drawer.canvasentry.js
@@ -187,7 +187,7 @@ export default class CanvasEntry {
         }
 
         if (radius) {
-            this.drawRoundedRectangle(ctx, x, y, width, height, radius);
+            this.drawRoundedRect(ctx, x, y, width, height, radius);
         } else {
             ctx.fillRect(x, y, width, height);
         }
@@ -205,9 +205,9 @@ export default class CanvasEntry {
      * @param {number} radius Radius of the rectangle
      *
      * @return {void}
-     * @example drawRoundedRectangle(ctx, 50, 50, 5, 10, 20)
+     * @example drawRoundedRect(ctx, 50, 50, 5, 10, 20)
      */
-    drawRoundedRectangle(ctx, x, y, width, height, radius) {
+    drawRoundedRect(ctx, x, y, width, height, radius) {
         if (height === 0) {
             return;
         }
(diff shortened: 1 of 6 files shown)


src/drawer.canvasentry.js
```diff
@@ -171,12 +179,59 @@ export default class CanvasEntry {
...
+     * @param {number} radius Radius of the rectangle
+     *
+     * @return {void}
+     * @example drawRoundedRectangle(ctx, 50, 50, 5, 10, 20)
+     */
+    drawRoundedRectangle(ctx, x, y, width, height, radius) {
+        if (height === 0) {
+            return;
+        }
+        if (height < 0) {
```
thijstriemstra: can you add a comment here, describing why this check is added?

kubk: Hi @thijstriemstra , as I understand it, a waveform consists of a peaks with float values from -1 to 1. The peak value is used to calculate bar height: https://github.com/katspaugh/wavesurfer.js/blob/8ba3328102da9746ce1fd390c0801f01edaaf612/src/drawer.multicanvas.js#L295
That's why a bar height can be negative. We need to use absolute height value in order to correctly calculate rounded rectangle coordinates because the [formula](https://github.com/katspaugh/wavesurfer.js/issues/953#issuecomment-510230749) expects the height to be positive. Does that make sense?

thijstriemstra: makes sense, a good comment to add there. I will add it and merge this PR, thanks.

kubk:
@thijstriemstra Thank you! :+1:

thijstriemstra:
> If this will be merged I will document this feature in gh-pages branch

thanks @kubk, looking forward to that PR as well

PR #1762, opened 2019-10-04
regions plugin: change position of region end handler

viliuslinge:
Currently start handler is located inside the region while end handler is outside of the region boundary.
PR solves the issue by relocating the end handler to be a part of whole region block

No breaking changes

No breaking changes

### Todos/Notes:
-

-


thijstriemstra:
thanks @viliuslinge, looks good to me.

PR #1767, opened 2019-10-07
Allow to use HTML5 audio tag with API Web Audio

### LINKED ISSUE #1763: Load big audio files with WebAudio backend and MediaElementAudioSourceNode
I tried to load an audio file with a duration of three hours, but the browser crashes because internally the library uses AudioBuffer to decode the audio data and AudioBuffer can not be used with huge files. An only solution that i found is to load the file with MediaElement backend, but i know that this backend handle the audio resource without Api Web Audio, so i will lost all my functionalities. 

From the API Web Audio, there is a possibility to load an audio as an HTML tag, but using as a source node the MediaElementAudioSourceNode, instead of AudioBufferSourceNode which decodes internally the audio data causing the crash of the browser in case of big files. Using  MediaElementAudioSourceNode, it can can be connected to other nodes of the Audio graph normally, using the Audio Context, so filters and other functionalities can be applied (also it can be connected to the AnalyserNode to draw the waveform). 

Is there a possibility to implement this option or is there another way to load big audio files with the possibility to have audio filters and other features?

It is a short example:

```javascript
            const audioContext = new AudioContext();
            let audio = document.querySelector('audio');
            let mediaStreamSourceNode =  audioContext.createMediaElementSource(audio);

            let gainNode =  audioContext.createGain();
            gainNode.gain.value =1;

            mediaStreamSourceNode.connect(gainNode).connect(audioContext.destination);
```

marizuccara:
Introduced the WebAudioMedia backend which allows to load audio as HTML5 audio tag and use it with WebAudio backend. 

Setting the WebAudioMedia backend,there is a possibility to load audio of big dimensions, using the API Web Audio features. In this way, filters and other functionalities can be performed as with WebAudio backend without decoding internally audio data, that caused crash of the browser. 

Added the WebAudioMedia backend. The API to use to load an audio or playback it, are the same as with the MediaElement backend. 

Added a file WebAudioMedia  that extends WebAudio. Instead of using the AudioBufferSourceNode to push the source to the AudioContext, it was used the MediaElementSourceNode which takes in input the audio tag. Since an HTML5 file is loaded, there were used APIs of MediaElement backend. 

### Todos/Notes:

It was created another backend to test this functionality and show you my proposal. In fact, I have duplicated the MediaElement code because the resource is handled as HTML5 audio tag. For other operations, there were used WebAudio APIs. 

fixes #1763


src/webaudio-mediasource.js
```diff
@@ -0,0 +1,459 @@
...
+        this.on('play', onAudioProcess);
+
+        // Update the progress one more time to prevent it from being stuck in
+        // case of lower framerates
+        this.on('pause', () => {
+            this.fireEvent('audioprocess', this.getCurrentTime());
+        });
+    }
+    /** @private */
+    createScriptNode() {
```
katspaugh: Isn't this method exactly the same in the parent WebAudio class?

src/webaudio-mediasource.js
```diff
@@ -0,0 +1,459 @@
...
+            }
+        }
+        this.scriptNode.connect(this.ac.destination);
+    }
+    /**
+     * Create the gain node needed to control the playback volume.
+     *
+     * @private
+     */
+    createVolumeNode() {
```
katspaugh: This one can also be inherited w/o duplication.

src/webaudio-mediasource.js
```diff
@@ -0,0 +1,459 @@
...
+        if (this.ac.createGain) {
+            this.gainNode = this.ac.createGain();
+        } else {
+            this.gainNode = this.ac.createGainNode();
+        }
+        // Add the gain node to the graph
+        this.gainNode.connect(this.ac.destination);
+    }
+    /** @private */
+    createAnalyserNode() {
```
katspaugh: Same here. Probably some others, too. I didn't check further.

marizuccara:
> Thanks for this PR!
> What was your incentive to implement this new backend?
> My guess would be some people might want to use a media element but control it via Web Audio?

If I load a big audio file with WebAudio backend, because it internally decodes the AudioBuffer, the browser crashes. Using HTML5 tag, the audio data are not decoded, so in case of a big audio file browser does not crash. Moreover, in this way I can use it with API Web Audio and add filters, like control left and right channels, pannerNode, equalizer, etc

src/webaudio-mediasource.js
marizuccara: Yes, i've added only createSource method which allow me to create MediaElementSourceNode (instead of AudioBufferSourceNode) and handle audio resource with API Web Audio. I've duplicated some code, only to show you my solution in a different file,  to give more clarity to my purpose.

katspaugh:
> Using HTML5 tag, the audio data are not decoded, so in case of a big audio file browser does not crash.

Even with HTML5, the audio data is still [decoded](https://github.com/katspaugh/wavesurfer.js/blob/master/src/wavesurfer.js#L1411-L1432) to draw the waveform.

marizuccara:
> Even with an HTML5 media element, the audio data is still [decoded](https://github.com/katspaugh/wavesurfer.js/blob/master/src/wavesurfer.js#L1411-L1432) to draw the waveform.

But giving peaks, the waveform is drawn with them

katspaugh:
Yes, correct.

marizuccara:
> Yes, correct.
With WebAudio backend, even if i give peaks to draw the waveform, it internally decode the audio to transform it to an AudioBuffer and put it to AudioBufferSourceNode. So the waveform is drawn without any problems, but in case of big audio files, the browser crashes because it decodes the data in memory. 
If we give an HTML5 tag with peaks, the waveform is drawn without decoding. Moreover, giving the source (HTML tag) to MediaElementSourceNode, it does not decode audio data in memory, so it can be played without any problems (like with MediaElement backend). But in this way we can add filters and have more controls with audio, because we can use API Web Audio

katspaugh:
Cool, gotcha.
I guess this new backend a good option then. More powerful than a standalone media element, but more efficient than the buffer source.
Let's clean up the PR and merge it. Any idea how to get that coverage back to 79%?

src/wavesurfer.js
```diff
@@ -753,7 +759,6 @@ export default class WaveSurfer extends util.Observer {
         this.fireEvent('interaction', () => this.play(start, end));
         return this.backend.play(start, end);
     }
-
```
katspaugh: I suggest we restore this new line to avoid unnecessary changes.

src/wavesurfer.js
```diff
@@ -1407,7 +1415,6 @@ export default class WaveSurfer extends util.Observer {
             });
         }
     }
-
```
katspaugh: Same here.

marizuccara:
> Cool, gotcha.
> I guess this new backend a good option then. More powerful than a standalone media element, but more efficient than the buffer source.
> Let's clean up the PR and merge it. Any idea how to get that coverage back to 79%?

Maybe, we can remove private to functions like createScriptNode, createVolumeNode and Analyser, so we can extend them normally

src/webaudio-mediasource.js
```diff
@@ -0,0 +1,459 @@
...
+     * Construct the backend
+     *
+     * @param {WavesurferParams} params Wavesurfer parameters
+     */
+    constructor(params) {
+        super(params);
+        /** @private */
+        this.params = params;
+        /** @private */
+        this.params = params;
```
aburai: duplicated code

src/webaudio-mediasource.js
aburai: i worry about if we need a new backend to handle this, i need time to check it

src/webaudio-mediasource.js
marizuccara: How does this work? Can I modify code deleting duplicated code or in this phase do you make optimizations?

katspaugh: You can commit changes in your local repo and push to your GitHub fork's master.

src/mediaelement.js
```diff
@@ -178,6 +184,15 @@ export default class MediaElement extends WebAudio {
...
         this.setVolume(this.volume);
+        if (this.params.mediaElementWebAudio) {
+            this.createMediaElementSource(media);
+        }
+    }
+    createMediaElementSource(mediaElement) {
+        this.sourceMediaElement = this.ac.createMediaElementSource(
+            mediaElement
+        );
+        this.sourceMediaElement.connect(this.analyser);
```
katspaugh: An alternative option would be to keep WebAudioMedia as a separate backend and just inherit from the MediaElement backend. With just these two methods (`init` and `createMediaElementSource`).

marizuccara: Yes, it can be also an alternative if we want handle audio with MediaElement backend only as HTML5 tag.  It is a business logic choice, I can restore to WebAudioMedia backend and inherit from MediaElement!

katspaugh: I think it would be cleaner as a separate class. 👍

marizuccara: But putting createMediaElementSource to this new class, i have to override _load method because it uses createMediaElementSource method.

src/webaudio.js
```diff
@@ -135,9 +128,7 @@ export default class WebAudio extends util.Observer {
         this.peaks = null;
         /** @private */
         this.playbackRate = 1;
-        /** @private */
         this.analyser = null;
-        /** @private */
         this.scriptNode = null;
         /** @private */
         this.source = null;
```
thijstriemstra: @marizuccara can you replace `/** @private */` with a meaningful comment for each variable?

thijstriemstra: can you also add `@since` in new classes and/or methods? https://jsdoc.app/tags-since.html

src/mediaelement.js
katspaugh: In ES6, you can use [`super`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/super) to refer to the parent class methods.
So your method would looks smth like:
```
_load() {
  super._load();
  this.createMediaElementSource(media);
}
```

marizuccara: ah thanks!

katspaugh:
We just need to add the new backend type to the docs now. The docs are on another branch, [`gh-pages`](https://github.com/katspaugh/wavesurfer.js/tree/gh-pages), so it should be another PR.

marizuccara:
> can you also look at simple unit tests (see existing MediaElement tests)?

Have I run your karma tests or write other tests?

marizuccara:
> > > can you also look at simple unit tests (see existing MediaElement tests)?
> > 
> > 
> > Have I run your karma tests or write other tests?
> 
> before pushing pull requests one should always run the tests with `npm run test` or `yarn test`
> but in this case we add something new that should have test coverage
> an example is found under `spec/mediaelement.spec.js`
> if you need any help, let me know, when i have time to test the new backend, i can have a look at the tests and/or append some tests that make sense

Thanks. I've created  `spec/webaudio-media.spec.js` which has the same tests like MediaElement backend. These are the results: 

 
image

thijstriemstra:
thanks @marizuccara can you commit those as well so the CI service can run the tests as well?

src/wavesurfer.js
```diff
@@ -1322,6 +1329,8 @@ export default class WaveSurfer extends util.Observer {
                 return this.loadBuffer(url, peaks, duration);
             case 'MediaElement':
                 return this.loadMediaElement(url, peaks, preload, duration);
```
thijstriemstra: this line can be removed if it's the same call for both cases.

src/webaudio-media.js
```diff
@@ -0,0 +1,56 @@
+import MediaElement from './mediaelement';
+
+/**
+ * WebAudioMedia backend
```
thijstriemstra: can you add a description for the backend here (possibly copied from the discussion on this PR)?

src/webaudio-media.js
```diff
@@ -0,0 +1,56 @@
...
+ * @since Version 3.1.0
```
thijstriemstra: new version would be 3.2.0

src/webaudio-media.js
```diff
@@ -0,0 +1,60 @@
...
+     */
+    _load(media, peaks) {
+        super._load(media, peaks);
+        this.createMediaElementSource(media);
+    }
+
+    /**
+     * Create MediaElementSource node
+     *
+     * @since Version 3.1.0
```
thijstriemstra: incorrect version nr

thijstriemstra:
>      Thanks. I've created spec/webaudio-media.spec.js which has the same tests like MediaElement backend.

Having a base test class that can be used for both mediaElement and this new backend would be useful. Thoughts?

marizuccara:
> > ```
> >  Thanks. I've created spec/webaudio-media.spec.js which has the same tests like MediaElement backend.
> > ```
> 
> Having a base test class that can be used for both mediaElement and this new backend would be useful. Thoughts?
> 
> Also: will `wavesurfer.loadBlob` work with this backend?

No, `wavesurfer.loadBlob` does not work with this backend because it does not decode audio data, and with `wavesurfer.loadBlob` audio data are decoded. So this method works only with WebAudio backend.

src/webaudio-media.js
marizuccara: Ah, sorry, I forgot to edit this version number!
I've documented also `mediaType` and `mediaControls` properties.

marizuccara:
> > > ```
> > >  Thanks. I've created spec/webaudio-media.spec.js which has the same tests like MediaElement backend.
> > > ```
> > 
> > 
> > Having a base test class that can be used for both mediaElement and this new backend would be useful. Thoughts?
> > Also: will `wavesurfer.loadBlob` work with this backend?
> 
> No, `wavesurfer.loadBlob` does not work with this backend because it does not decode audio data, and with `wavesurfer.loadBlob` audio data are decoded. So this method works only with WebAudio backend.

Maybe is it possible to implement a shared file which has the tests for both backends and include this file to `mediaelement.spec` and `webaudio-media.spec`?

src/webaudio-media.js
thijstriemstra: only the version number is ok, don't include `Version`

marizuccara:
> can you also add an example that uses the backend? Also take a look at adding it to the changelog

Added examples and updated CHANGES.md. 
Tell me what you think about the solution of shared tests.

thijstriemstra:
I'm not sure about the name `WebAudioMedia`, it's a term we came up and I find it not clear. What about `MediaElementSource` or something related to the method we're using?

marizuccara:
> I'm not sure about the name `WebAudioMedia`, it's a term we came up and I find it not clear. What about `MediaElementSource` or something related to the method we're using?
> 
> > updating changelog caused conflicts
> 
> Fixed this for you.
> 
> > Tell me also what do you think about the solution of shared tests.
> 
> Looks good.

Thanks! I called it in this way for using mediaelement with Api Web Audio. With MediaElementSource is it clear that i can use it with API Web Audio, since the creation of the source is an internal stuff? But anyway, surely you know better what is the best way to call this backend, so if you think that is clearer calling it MediaElementSource, i can change without any problems the name!

thijstriemstra:
> With MediaElementSource is it clear that i can use it with API Web Audio,

No, not really. Maybe `MediaElementWebAudio` then? It's long but it's clear (to me).

marizuccara:
> > With MediaElementSource is it clear that i can use it with API Web Audio,
> 
> No, not really. Maybe `MediaElementWebAudio` then? It's long but it's clear (to me).

Yes, I also think that this new name is more explicit and sounds better! So I'm changing it, okay?

thijstriemstra:
> So I'm changing it, okay?

Yep!

src/wavesurfer.js
```diff
@@ -30,12 +31,19 @@ import PeakCache from './peakcache';
...
+ * playback, giving also peaks, so the audio data are not decoded. In this way you can use WebAudio features, like filters,
+ * also with audio with big duration.For example:
+ * ` wavesurfer.load(url, peaks, preload, duration);
+ *   wavesurfer.play();
+ *   wavesurfer.setFilter(customFilter);
+ * `
  * @property {string} backgroundColor=null Change background color of the
  * waveform container.
- * @property {number} barHeight=1 The height of the wave bars.
+ * @property {number} barHeight=1 The height of the wave bars.F
```
thijstriemstra: typo

src/wavesurfer.js
```diff
@@ -30,12 +31,19 @@ import PeakCache from './peakcache';
...
  * center waveform on current progress
- * @property {string} backend='WebAudio' `'WebAudio'|'MediaElement'` In most cases
- * you don't have to set this manually. MediaElement is a fallback for
- * unsupported browsers.
+ * @property {string} backend='WebAudio' `'WebAudio'|'MediaElement'|'MediaElementWebAudio'` In most cases
+ * you don't have to set this manually. MediaElement is a fallback for unsupported browsers.
+ * MediaElementWebAudio allows to use API Web Audio also with big audio files, loading audio like with
+ * MediaElement backend (HTML5 audio tag). You have to use the same methods of MediaElement backend for loading and
+ * playback, giving also peaks, so the audio data are not decoded. In this way you can use WebAudio features, like filters,
+ * also with audio with big duration.For example:
```
thijstriemstra: Missing space before `For`

src/wavesurfer.js
```diff
@@ -67,10 +75,10 @@ import PeakCache from './peakcache';
  * even integer). If the waveform is longer than this value, additional canvases
  * will be used to render the waveform, which is useful for very large waveforms
  * that may be too wide for browsers to draw on a single canvas.
- * @property {boolean} mediaControls=false (Use with backend `MediaElement`)
+ * @property {boolean} mediaControls=false (Use with backend `MediaElement` or `MediaElementWebAudio`)
  * this enables the native controls for the media element
- * @property {string} mediaType='audio' (Use with backend `MediaElement`)
- * `'audio'|'video'`
+ * @property {string} mediaType='audio' (Use with backend `MediaElement` or `MediaElementWebAudio`)
+ * `'audio'|'video'` ('video' only for MediaElement)
```
thijstriemstra: missing backticks

src/webaudio.js
```diff
@@ -11,9 +11,9 @@ const FINISHED = 'finished';
  * @extends {Observer}
  */
 export default class WebAudio extends util.Observer {
-    /** @private */
+    /** scriptBufferSize: size of the processing buffer */
     static scriptBufferSize = 256;
-    /** @private */
+    /** audioContext: allows to process audio with API Web Audio */
```
thijstriemstra: "WebAudio API"

spec/mediaelement-mediaelement_webaudio-shared.js
```diff
@@ -0,0 +1,275 @@
+/* eslint-env jasmine */
```
thijstriemstra: why not use a filename like `media-element-shared.js`?

marizuccara:
The new tests that you write for MediaElement backend (error tests), maybe are good also for the new backend since we can pass to load method the HTML5 tag?

marizuccara:
> > Added examples and updated CHANGES.md (updating changelog caused conflicts).
> 
> I don't see the example?

CHANGES.md
```diff
@@ -5,6 +5,16 @@ wavesurfer.js changelog
...
+- New `WebAudioMedia` backend (#1767):
+  - It allows to use API Web Audio also with big audio files, loading audio like with MediaElement backend (HTML5 audio tag), 
+    so you have to use the same methods of MediaElement backend for loading and playback. In this way, the audio
+    resource is not loaded entirely from server, but in ranges, allowing you to use WebAudio features, like filters, 
+    also with audio with big duration. You have to give also peaks, so the audio data are not decoded.
+    For example:
+    ` wavesurfer.load(url, peaks, preload, duration);
+      wavesurfer.play();
+      wavesurfer.setFilter(customFilter);
+    `
```
marizuccara: This example is not good? The same in changelog

thijstriemstra: I was thinking of an example in the `example` directory that uses the new backend (and adds some filters for example).

marizuccara: I've pushed the example.
- The filters provided are good? In the explanation i've commented only one because took me a lot of space.
- I've uploaded a file audio taken from `split-channels` example. I've initially wanted to load a 3 hours audio (because with an audio of this dimensions, with `WebAudio` backend the browser crashes, but with `MediaElementWebAudio` not, but GitHub tell me that the file is too large).
- I've used `audiowaveform` with `split-channels` command, but the data array that returned is only of one dimension, so i drawn the entire waveform, without splitting the channels (if you know how i can solve the problem, tell me)

example/mediaelement-webaudio/app.js
```diff
@@ -0,0 +1,91 @@
...
+        ]
+    });
+
+    // get audio peaks
+    fetch('peaks.json')
+        .then(response => {
+            return response.json();
+        })
+        .then(peaks => {
+            let max = peaks.data.reduce((max, el) => (el > max ? el : max));
```
thijstriemstra: can you add a comment for the user what's happening here?

example/mediaelement-webaudio/app.js
```diff
@@ -0,0 +1,91 @@
...
+
+            return peaks.data.map(el => {
+                return el / max;
+            });
+        })
+        .then(normalizedPeaks => {
+            // Load audio from URL
+            wavesurfer.load('stereo.mp3', normalizedPeaks, 51);
```
thijstriemstra: you should load the file from `../media/`. If it's a copy, adjust both examples to use the same file.

example/mediaelement-webaudio/peaks.json
```diff
@@ -0,0 +1 @@
...3,-1,5,-5,3,-10,17,-15,13,-11,20,-14,17,-8,17,-9,11,-6,15,-11,12,-5,13,-9,9,-5,17,-10,8,-3,12,-8,7,-2,13,-9,8,-3,12,-7,8,-3,13,-8,6,-2,10,-8,4,0,12,-8,6,0,6,-5,3,0,6,-4,1,1,5,-3,1,0,6,-3,2,0,6,-3,1,1,5,-2,1,0,12,-10,5,-9,20,-14,16,-12,18,-11,11,-8,15,-11,11,-4,16,-10,7,-5,13,-8,8,-3,12,-8,7,-3,14,-8,7,-3,13,-7,5,-1,11,-7,6,-2,11,-7,6,-1,11,-7,6,-1,6,-4,3,0,5,-4,3,0,5,-3,2,0,4,-1,0,0,4,-1,1,1,4,-2,1,0,3,-2,1,-11,21,-14,15,-10,16,-12,11,-7,15,-11,10,-6,16,-10,10,-5,13,-9,8,-4,11,-7,7,-3,11,-7,7,-3,11,-7,7,-3,11,-7,6,-3,10,-7,6,-3,12,-6,7,-1,5,-4,2,0,6,-3,2,0,4,-3,2,0,4,-2,1,1,4,-2,1,1,4,-2,1,0,4,-2,1,-4,15,-11,11,-12,25,-15,17,-7,14,-9,9,-6,17,-10,11,-6,15,-9,9,-5,14,-8,7,-6,13,-8,7,-3,10,-9,7,-3,10,-9,7,-4,12,-7,6,0,6,-4,3,-1,5,-4,3,0,4,-3,1,0,4,-3,1,0,4,-2,1,0,4,-2,1,0,4,-2,1,0,4,-2,0,0,5,-2,1,-11,24,-15,18,-8,16,-13,9,-6,14,-9,9,-5,15,-9,9,-4,13,-8,9,-6,14,-7,6,-5,12,-8,8,-4,11,-9,8,-3,10,-7,6,-3,11,-6,6,-2,5,-4,3,-1,5,-4,3,-1,5,-3,2,0,4,-3,2,0,4,-2,1,0,3,-1,1,0,3,-2,1,0,3,-2,0,-7,13,-13,11,-10,20,-13,16,-10,16,-9,12,-8,13,-8,9,-7,14,-8,9,-5,11,-7,9,-6,12,-10,10,-4,10,-9,9,-5,12,-9,8,-4,10,-8,8,-3,10,-8,7,-1,5,-4,3,-1,4,-4,2,0,5,-3,2,-1,4,-2,1,0,3,-2,1,0,3,-2,1,0,2,-2,0,0,3,-2,0,-
```
thijstriemstra: i'd call it 'filename-peaks' so `stereo-peaks.json`.

kavishatalsania:
Hi @marizuccara / @thijstriemstra / @katspaugh ,
Thanks for this great feature.

I used 'MediaElementWebAudio' backend. It does not load whole audio and I am also able to use filters to the mute channel.

One issue I found is when I load the audio file from s3 it is giving error **'MediaElementAudioSource outputs zeroes due to CORS access restrictions**'

I am able to resolve this issue by adding line media.crossOrigin = "anonymous" in the load method of wavesurfer at line 1955.

```
key: "load",
value: function load(url, container, peaks, preload) {
      var media = document.createElement(this.mediaType);
      media.controls = this.params.mediaControls;
      media.autoplay = this.params.autoplay || false;
      media.preload = preload == null ? 'auto' : preload;
      media.src = url;
      media.style.width = '100%';
      media.crossOrigin = "anonymous";
      var prevMedia = container.querySelector(this.mediaType);
      if (prevMedia) {
        container.removeChild(prevMedia);
      }
      container.appendChild(media);
      this._load(media, peaks);
}
```
Is it right resolution for this problem?

marizuccara:
You should be able also to create yourself an audio tag.
For example, `<audio src="http://domain/audio" crossOrigin="anonymous">`,
or by JavaScript, 
` let audio = document.createElement('audio');`
`audio.src = "http://domain/audio";`
`audio.crossOrigin = "anonymous";`
Then, pass the audio tag to `load` method:
`wavesurfer.load(audio, peaks)`.

kavishatalsania:
Hi @marizuccara 
Thanks for your suggestion. It's working.

I got impression from the documentation and code that I could only pass the audio URL in the load method not audio object.

CHANGES.md
```diff
@@ -18,6 +17,7 @@ wavesurfer.js changelog
       wavesurfer.play();
```
marizuccara: Maybe in this way is clearer?
`wavesurfer.load(url | HTMLMediaElement, peaks, preload, duration);`

kavishatalsania: True

thijstriemstra:
> audio.crossOrigin = "anonymous";

can we add this line to the example?

thijstriemstra:
merged. thanks @marizuccara!

katspaugh:
Thanks a lot, @marizuccara! A great contribution!

marizuccara:
Thanks so much to you, @katspaugh and @thijstriemstra, for this fantastic library!

Leaybc:
Looking forward to releasing new version as soon as possible

kavishatalsania:
Hi @katspaugh / @thijstriemstra 

When are you planning to release new version with these changes?

PR #1774, opened 2019-10-11
throw error when url parameter is not supplied to load()

### LINKED ISSUE #1773: Calling load() without arguments throws error
calling `wavesurfer.load()` without params should not throw an error:

```javascript
var surfer = new WaveSurfer({
        backend: 'MediaElement',
        container: '#waveform'
      });

surfer.load();
```

Error
```
TypeError: this.backend is null
```

thijstriemstra:
fixes #1773

### DIFF
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -1607,8 +1607,12 @@ export default class WaveSurfer extends util.Observer {
                 true
             );
         }
-        this.backend.destroy();
-        this.drawer.destroy();
+        if (this.backend) {
+            this.backend.destroy();
+        }
+        if (this.drawer) {
+            this.drawer.destroy();
+        }
         this.isDestroyed = true;
         this.isReady = false;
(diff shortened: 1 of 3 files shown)


src/wavesurfer.js
```diff
@@ -1290,6 +1290,9 @@ export default class WaveSurfer extends util.Observer {
      * );
      */
     load(url, peaks, preload, duration) {
+        if (url === undefined) {
+            throw new Error('url parameter is invalid');
```
katspaugh: Maybe `if (!url)` to also throw if it's an empty string or null?
The error message could then be `The URL parameter cannot be empty` or similar.

thijstriemstra: sounds good, was struggling to find a good description for the error but that's a better check.

thijstriemstra:
Also fixed error for calling `wavesurfer.destroy()`.

PR #1775, opened 2019-10-11
throw error when invalid url is passed to MediaElement backend

### LINKED ISSUE #1773: Calling load() without arguments throws error

thijstriemstra:
follow-up to #1774 

fixes #1773

### DIFF
--- a/src/mediaelement.js
+++ b/src/mediaelement.js
@@ -133,7 +133,7 @@ export default class MediaElement extends WebAudio {
         // verify media element is valid
         if (
             !(media instanceof HTMLMediaElement) ||
-            typeof media.addEventListener === "undefined"
+            typeof media.addEventListener === 'undefined'
         ) {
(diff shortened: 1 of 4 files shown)


src/mediaelement.js
```diff
@@ -135,6 +139,14 @@ export default class MediaElement extends WebAudio {
             media.load();
         }
 
+        // verify media element
+        if (
+            !(media instanceof HTMLMediaElement) ||
+            media.addEventListener === undefined
```
aburai: i prefer typeof check for undefined
and should the verify be on top of the _load method? i am not sure

thijstriemstra: done

PR #1776, opened 2019-10-16
Cursor Plugin: Flip time text to left of cursor on right edge for readability

mdtanrikulu:
Hey there!

I did a small change on cursor plugin, which improves the readability of duration text next to the cursor. When user point the very right edge, the duration will be placed on left side of the cursor for better UX. See the gif; 

**Before:**
duration_without_flip

**After:**
duration_flip

### DIFF
--- a/src/plugin/cursor.js
+++ b/src/plugin/cursor.js
@@ -241,6 +241,7 @@ export default class CursorPlugin {
      *
      * @param {number} xpos The x offset of the cursor in pixels
      * @param {number} ypos The y offset of the cursor in pixels
+     * @param {boolean} flip Flag to flip duration text from right to left
      */
     updateCursorPosition(xpos, ypos, flip = false) {
         this.style(this.cursor, {
@@ -324,7 +325,7 @@ export default class CursorPlugin {
     /**
      * Get outer width of given element.
      *
-     * @param {e} DOM Element
+     * @param {DOM} e DOM Element
      * @returns {number} outer width
      */
     outerWidth(e) {
--- a/src/plugin/cursor.js
+++ b/src/plugin/cursor.js
@@ -90,13 +90,14 @@ export default class CursorPlugin {
         const bbox = this.wavesurfer.container.getBoundingClientRect();
         let y = 0;
         let x = e.clientX - bbox.left;
+        let flip = bbox.right < e.clientX + this.outerWidth(this.displayTime);
 
         if (this.params.showTime && this.params.followCursorY) {
             // follow y-position of the mouse
             y = e.clientY - (bbox.top + bbox.height / 2);
         }
 
-        this.updateCursorPosition(x, y);
+        this.updateCursorPosition(x, y, flip);
     };
 
     /**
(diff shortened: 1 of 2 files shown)


thijstriemstra:
awesome, thanks!

src/plugin/cursor.js
```diff
@@ -311,4 +321,20 @@ export default class CursorPlugin {
...
         );
     }
+
+    /**
+     * Get outer width of given element.
+     *
+     * @param {DOM} e DOM Element
+     * @returns {number} outer width
+     */
+    outerWidth(e) {
```
thijstriemstra: can you use a more descriptive var name then `e`?

src/plugin/cursor.js
```diff
@@ -311,4 +321,20 @@ export default class CursorPlugin {
...
+        if (!e) return 0;
+
+        var width = e.offsetWidth;
```
thijstriemstra: use let instead of var

src/plugin/cursor.js
```diff
@@ -200,12 +201,15 @@ export default class CursorPlugin {
                         {
                             display: 'inline',
                             pointerEvents: 'none',
-                            margin: 'auto'
+                            margin: 'auto',
+                            visibility: 'hidden' //initial value will be hidden just for measuring purpose
```
thijstriemstra: space between // and initial

src/plugin/cursor.js
```diff
@@ -200,12 +201,15 @@ export default class CursorPlugin {
...
                         },
                         this.params.customShowTimeStyle
                     )
                 )
             );
+            //initial value to measure display width
```
thijstriemstra: space between // and initial

src/plugin/cursor.js
```diff
@@ -200,12 +201,15 @@ export default class CursorPlugin {
...
+            this.displayTime.innerHTML = '0:00:000';
```
thijstriemstra: instead of hardcoding the time value, use the (user-supplied or default) format time method instead.

mdtanrikulu: Oh, I didn't see there is custom format method. Thanks for pointing it, I've updated.

src/plugin/cursor.js
```diff
@@ -255,9 +259,14 @@ export default class CursorPlugin {
             const timeValue =
                 Math.max(0, (xpos / elementWidth) * duration) + scrollTime;
             const formatValue = this.formatTime(timeValue);
+            if (flip) {
+                const textOffset = this.outerWidth(this.displayTime);
+                xpos -= textOffset;
+            }
```
aburai: can be shorten to `if (flip) xpos -= this.outerWidth(this.displayTime);`

src/plugin/cursor.js
```diff
@@ -90,13 +90,14 @@ export default class CursorPlugin {
         const bbox = this.wavesurfer.container.getBoundingClientRect();
         let y = 0;
         let x = e.clientX - bbox.left;
+        let flip = bbox.right < e.clientX + this.outerWidth(this.displayTime);
```
aburai: we should use `const` where possible

aburai:
ups, already merged, i oversee that ;)

PR #1778, opened 2019-10-20
fix destroy in MediaElement backend

thijstriemstra:
one of my projects that uses wavesurfer.js runs some unit tests that create/destroy a MediaElement backend within a few 100 ms. This situation throws some errors that this PR aims to fix.

### DIFF
--- a/src/mediaelement.js
+++ b/src/mediaelement.js
@@ -15,8 +15,11 @@ export default class MediaElement extends WebAudio {
         /** @private */
         this.params = params;
 
-        // Dummy media to catch errors
-        /** @private */
+        /**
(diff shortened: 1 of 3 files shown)


src/mediaelement.js
```diff
@@ -391,6 +403,13 @@ export default class MediaElement extends WebAudio {
     destroy() {
         this.pause();
         this.unAll();
+        this.destroyed = true;
```
aburai: should we put this at start of the block?

thijstriemstra: no, it should be set after `unAll` is called.

PR #1781, opened 2019-10-23
style right and left handles of region plugin independently

osheroff:

In 804334d63e36f88b5340aa8faac2d0db5adf5acf, the css style of the right region resize handle was changed to `right: 0`, but the element in question already had a `left: 0` style.  This means that the drag handles for the region were basically in the same place.


thijstriemstra:
thanks!

PR #1765, opened 2019-10-05
add dblclick event listener to wavesurfer wrapper

aburai:
### Short description of changes:
support dblclick event listener

### TODO
add `dblclick` (and `click`) to docs wavesurfer events

### Related Issues and other PRs:
fixes #1764

PR #1790, opened 2019-10-28
Expose progress to the dblclick event

Reached:
This is in line with how the single click event handler does it.

#1765

Thanks to @thijstriemstra for pointing me in the right direction :)


thijstriemstra:
thanks!

PR #1793, opened 2019-10-31
regions plugin: add maxRegions option to limit max nr of created regions

osheroff:
I've been heavily using the regions plugin in a project i'm playing with.  One feature I definitely needed (for my UI's sanity) was to limit the number of regions a user could create (to 1, in my case).

### DIFF
--- a/src/plugin/regions.js
+++ b/src/plugin/regions.js
@@ -643,6 +643,7 @@ export default class RegionsPlugin {
         this.params = params;
         this.wavesurfer = ws;
         this.util = ws.util;
+        this.maxRegions = params.maxRegions;
         this.util.getRegionSnapToGridValue = value => {
             return this.getRegionSnapToGridValue(value, params);
         };
@@ -836,7 +837,11 @@ export default class RegionsPlugin {
                 return;
             }
 
+
             if (!region) {
+                if (this.maxRegions && Object.keys(this.list).length >= this.maxRegions)
+                    return;
+
                 region = this.add(params || {});
             }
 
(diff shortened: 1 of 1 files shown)


src/plugin/regions.js
```diff
@@ -643,6 +643,7 @@ export default class RegionsPlugin {
         this.params = params;
         this.wavesurfer = ws;
         this.util = ws.util;
+        this.maxRegions = params.maxRegions;
```
thijstriemstra: can you also document this at the top of this class, where the other params are documented?

src/plugin/regions.js
```diff
@@ -836,7 +837,11 @@ export default class RegionsPlugin {
                 return;
             }
 
+
             if (!region) {
+                if (this.maxRegions && Object.keys(this.list).length >= this.maxRegions)
```
thijstriemstra: add a comment describing what this if statement does.

src/plugin/regions.js
osheroff: as I'm documenting this I realize the implementation I wrote only honors maxRegions when a user tries to create a region.  Wondering if it should be honored when a region is programmatically created....

osheroff:
documented

src/plugin/regions.js
thijstriemstra: >  Wondering if it should be honored when a region is programmatically created....

I would assume it does..

osheroff: ok, maxRegions is now honored from the api too..

PR #1794, opened 2019-10-31
regions plugin: add start argument to play and playLoop methods

osheroff:
I've had the need of late to start playing a loop defined in the region plugin in the middle of the loop.

(fyi if you're interested, https://mixmastermiles.com/)

### DIFF
--- a/src/plugin/regions.js
+++ b/src/plugin/regions.js
@@ -98,15 +98,16 @@ class Region {
     }
 
     /* Play the audio region. */
-    play() {
-        this.wavesurfer.play(this.start, this.end);
+    play(start) {
+        const s = start || this.start;
+        this.wavesurfer.play(s, this.end);
         this.fireEvent('play');
         this.wavesurfer.fireEvent('region-play', this);
     }
 
     /* Play the region in loop. */
-    playLoop() {
-        this.play();
+    playLoop(start) {
+        this.play(start);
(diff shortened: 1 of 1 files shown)


src/plugin/regions.js
```diff
@@ -98,15 +98,16 @@ class Region {
...
-    play() {
-        this.wavesurfer.play(this.start, this.end);
+    play(start) {
+        const s = start || this.start;
+        this.wavesurfer.play(s, this.end);
         this.fireEvent('play');
         this.wavesurfer.fireEvent('region-play', this);
     }
 
     /* Play the region in loop. */
```
thijstriemstra: can you properly document this method (with @param etc)?

src/plugin/regions.js
```diff
@@ -98,15 +98,16 @@ class Region {
     }
 
     /* Play the audio region. */
-    play() {
-        this.wavesurfer.play(this.start, this.end);
+    play(start) {
```
thijstriemstra: can you properly document this method (with @param etc)?

osheroff: I documented these, but I couldn't figure out how to make `esdoc` actually generate anything about the methods of the`Region` class -- it seems it only bothers documenting the exported classes in a module?  Or at least that's my best guess...

PR #1795, opened 2019-11-03
allow changing the stop-point of wavesurfer during playback

osheroff:
When someone is playing back a loop or region, this allows for moving the loop end-point while playback is happening.

If you like I can also code that behavior into regions.js.  currently I'm moving the loop end point in external code.

### DIFF
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -768,6 +768,15 @@ export default class WaveSurfer extends util.Observer {
     }
 
     /**
+     * Set a point in seconds for playback to stop at.
+     *
+     * @param {number} position Position (in seconds) to stop at
+     */
+    setPlayEnd(position) {
+        this.backend.setPlayEnd(position);
+    }
+
+    /**
      * Stops and pauses playback
      *
      * @example wavesurfer.pause();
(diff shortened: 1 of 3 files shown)


src/webaudio.js
```diff
@@ -656,7 +656,7 @@ export default class WebAudio extends util.Observer {
 
         this.scheduledPause = end;
 
-        this.source.start(0, start, end - start);
+        this.source.start(0, start);
```
thijstriemstra: this doesn't seem right and break existing stuff..

osheroff: this doesn't break anything; the playback is still stopped at `end` here: https://github.com/katspaugh/wavesurfer.js/blob/2eb72e06b60b2e8ee05f97c1674f187dcaa4fea8/src/webaudio.js#L254

src/wavesurfer.js
```diff
@@ -767,6 +767,15 @@ export default class WaveSurfer extends util.Observer {
         return this.backend.play(start, end);
     }
 
+    /**
+     * Set a point in seconds for playback to stop at.
+     *
+     * @param {number} position Position (in seconds) to stop at
```
thijstriemstra: add a jsdoc version marker for 3.3.0 (also in webaudio backend)

osheroff: done

osheroff:
good call, I hadn't tested with the `MediaElement` backend.  tested, found a bug, fixed.

PR #1798, opened 2019-11-05
regions plugin: allow setting the handleColor inside addRegion

Reached:
Currently we can set the handleColor for a region inside CSS, while this is great, it does not allow the user to set the handleColor dynamically. 

- I currently moved out the creation of `handleLeft` and `handleRight` to the render function instead of inside the resize conditional, just so that we don't have to create them again, I am not sure whether or not this has a performance impact.

**Notes**
- Added a new handleColor attribute to the Region constructor
- The regular CSS solution still works as you would expect

### DIFF
--- a/src/plugin/regions.js
+++ b/src/plugin/regions.js
@@ -29,7 +29,14 @@ class Region {
         this.isDragging = false;
         this.loop = Boolean(params.loop);
         this.color = params.color || 'rgba(0, 0, 0, 0.1)';
-        this.handleColor = params.handleColor || 'rgba(0, 0, 0, 1)';
+        this.handleStyle = params.handleStyle || {
+            left: {
+                backgroundColor: 'rgba(0, 0, 0, 1)'
+            },
+            right: {
+                backgroundColor: 'rgba(0, 0, 0, 1)'
+            }
+        };
         this.data = params.data || {};
         this.attributes = params.attributes || {};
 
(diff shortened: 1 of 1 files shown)


Reached:
Great suggestion @thijstriemstra !

thijstriemstra:
any updates @Reached?

Reached:
Sorry i've been really busy with work stuff, will try to have a look at it this week :)

thijstriemstra:
any updates @Reached, would like to get this into next release.

Reached:
Sorry for the delay, I just added the necessary changes, so handles can now be styled independently from each other :)

thijstriemstra:
thanks. there are some lint errors left, see https://travis-ci.org/katspaugh/wavesurfer.js/builds/620301821#L461

Frank3K:
>  I currently moved out the creation of `handleLeft` and `handleRight` to the render function instead of inside the resize conditional, just so that we don't have to create them again, I am not sure whether or not this has a performance impact.

I just upgraded to wavesurfer 3.3.0. For non-resizable regions, this change has the result that a `<region>` element is created containing two "empty" `<handle>`-elements. These handles just have a single CSS property (`background-color: rgba(0, 0, 0, 1);`) and are not visible to the end-user.

Can you explain why this change was needed? Couldn't the following lines:

```js
        if (!this.handleStyle.left) {
            handleLeft.style.backgroundColor = 'rgba(0, 0, 0, 1)';
        } else {
            handleLeft.style.backgroundColor = this.handleStyle.left.backgroundColor;
        }
        ...
```

be included inside the `if (this.resize)`?

PR #1807, opened 2019-11-21
add barMinHeight option

### LINKED ISSUE #1693: Set minimum bar height
Sometimes, particularly with larger audio files (like a podcast), Wavesurfer generates a waveform with gaps.

This by itself is a surprising result. I followed the method [in the FAQ](https://wavesurfer-js.org/faq/) to generate peak waveform data, and there are no gaps in my audio long enough that they should be visible. The position of the gaps also differs depending on the width of the player, which would suggest a bug in the way that Wavesurfer handles a large amount of peak data.

Regardless of the cause, however, I'd like to ensure that the waveform is never drawn with bars with a height of 0. Is there a way to set the minimum bar height to avoid this issue?

osheroff:
I too wanted to scrub out the gaps in my bar waveform.  I'd add html documentation but I'm really not sure how that stuff works.

fixes #1693

### DIFF
--- a/src/drawer.multicanvas.js
+++ b/src/drawer.multicanvas.js
@@ -302,6 +302,8 @@ export default class MultiCanvas extends Drawer {
                         peaks[Math.floor(i * scale * peakIndexScale)] || 0;
                     let h = Math.round((peak / absmax) * halfH);
 
+                    /* in case of silences, allow the user to specify that we
+                     * always draw *something* (normally a 1px high bar) */
                     if (h == 0 && this.params.barMinHeight)
                         h = this.params.barMinHeight;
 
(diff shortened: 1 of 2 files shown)


src/drawer.multicanvas.js
```diff
@@ -300,7 +300,11 @@ export default class MultiCanvas extends Drawer {
                 for (i; i < last; i += step) {
                     const peak =
                         peaks[Math.floor(i * scale * peakIndexScale)] || 0;
-                    const h = Math.round((peak / absmax) * halfH);
+                    let h = Math.round((peak / absmax) * halfH);
+
+                    if (h == 0 && this.params.barMinHeight)
```
thijstriemstra: can you add a comment describing reason for this if statement

osheroff: done

thijstriemstra:
> . I'd add html documentation but I'm really not sure how that stuff works.

You need to use the `gh-pages` branch (instead of `master`) for pull requests or edit it online: https://github.com/katspaugh/wavesurfer.js/tree/gh-pages

PR #1811, opened 2019-11-24
return Promise from wavesurfer.exportPCM

### LINKED ISSUE #1728: exportPCM has timing issue
Hi @katspaugh 

In the latest version 3.0.0 does the "exportPCM" function work?

    wavesurfer.on('ready', function() {
        var pcmData = wavesurfer.exportPCM(1024, 1024, true);
        console.log("PCM: "+pcmData);
    });

All that shows in the console is:

    PCM: []

thijstriemstra:
Sometimes it can take a while before `JSON.stringify`, used in `exportPCM`, finishes, so return a Promise that resolves with data.

Also added missing `end` parameter to `exportPCM`.

I'm aware this is not backwards-compatible but not sure how else to fix/improve this issue.

fixes #1728

### DIFF
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -1530,15 +1530,15 @@ export default class WaveSurfer extends util.Observer {
      * @param {?boolean} noWindow Set to true to disable opening a new
      * window with the JSON
      * @param {number} start Start index
-     * @todo Update exportPCM to work with new getPeaks signature
+     * @param {number} end End index
      * @return {string} JSON of peaks
      */
-    exportPCM(length, accuracy, noWindow, start) {
+    exportPCM(length, accuracy, noWindow, start, end) {
         length = length || 1024;
         start = start || 0;
         accuracy = accuracy || 10000;
         noWindow = noWindow || false;
-        const peaks = this.backend.getPeaks(length, start);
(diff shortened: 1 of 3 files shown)


src/wavesurfer.js
```diff
@@ -1525,32 +1525,35 @@ export default class WaveSurfer extends util.Observer {
...
+        return new Promise((resolve, reject) => {
+            const json = JSON.stringify(arr);
+
+            if (!noWindow) {
+                window.open(
+                    'data:application/json;charset=utf-8,' +
+                        encodeURIComponent(json)
+                );
+            }
+            resolve(json);
```
katspaugh: Not sure about wrapping a synchronous call in a promise. Is the idea to schedule the execution after all other calls? To not freeze the UI?

thijstriemstra: @katspaugh take a look at #1728 for reasoning

katspaugh: I mean, if it helps, sure.

thijstriemstra: i'd prefer this than have users use that setTimeout hack. JSON.stringify (and backend.getPeaks) could take a while, and returning a Promise allows us to return as soon as it's done.

PR #1823, opened 2019-12-07
regions plugin: don't assign to module object

mustafa0x:
As it's invalid (I noticed when bundling with rollup).

Reproduce in a browser with something like the following:
```html
<script type=module>
import * as util from './util.js';
util.a = 1;
</script>
```


thijstriemstra:
thanks!

PR #1825, opened 2019-12-09
replace util.extend with Object.assign

mustafa0x:
[Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign)'s support is really good.


mustafa0x:
[request-animation-frame.js](https://github.com/katspaugh/wavesurfer.js/blob/master/src/util/request-animation-frame.js) is also obsolete at this point in time.

thijstriemstra:
@mustafa0x why replace this?

mustafa0x:
Why reimplement a native feature? Doing so makes the code more verbose, makes it harder for programmers to understand the code base, and could have performance impacts.

katspaugh:
Thanks, @mustafa0x! It’s about time! 🙂

Good point about keeping the old method in place (just calling Object.assign and logging a depreciation warning).

mustafa0x:
@katspaugh np! 🎉

thijstriemstra:
can you update the pull request @mustafa0x ?

mustafa0x:
I've updated it, unless you mean adding the tests back?

If so, I didn't feel like that made much sense, since it'll just test `Object.assign`.

mustafa0x:
It's also not used anywhere in the codebase, so testing it has little benefit.

src/util/extend.js
```diff
@@ -7,10 +7,6 @@
  * @return {Object} Merged object
  */
 export default function extend(dest, ...sources) {
-    sources.forEach(source => {
-        Object.keys(source).forEach(key => {
-            dest[key] = source[key];
-        });
-    });
-    return dest;
+    console.warn('util.extend is depreciated; use Object.assign');
```
thijstriemstra: typo: it's `deprecated`

thijstriemstra:
> If so, I didn't feel like that made much sense, since it'll just test Object.assign.

it tests existing `util.extend`, and since you're not removing that method (it's deprecated), you should also not remove the test either.

src/util/extend.js
thijstriemstra: don't remove the method content, it's deprecated, not removed.

mustafa0x: Typo fixed; thanks!

> don't remove the method content, it's deprecated, not removed.

I'm only changing the internal implementation, as @katspaugh suggested.

mustafa0x:
> > If so, I didn't feel like that made much sense, since it'll just test Object.assign.
> 
> it tests existing `util.extend`, and since you're not removing that method (it's deprecated), you should also not remove the test either.

I did remove all calls to that method, so testing neither benefits wavesurfer nor applications that use wavesurfer. And even if it did; we're just testing Object.assign (unless I change the test to `'extend' in util`).

Let me know if you want to add it back anyways.

thijstriemstra:
> I did remove all calls to that method, so testing neither benefits wavesurfer nor applications that use wavesurfer. 

This is simply not true. Existing applications might still use util.extend. Until they migrated this function still needs to work otherwise it'll break existing projects (inc my own).

> Let me know if you want to add it back anyways.

Yes please.

mustafa0x:
I've added the test back.

> This is simply not true. Existing applications might still use util.extend. Until they migrated this function still needs to work otherwise it'll break existing projects (inc my own).

I understand this. I was talking about the testing however. =)

src/util/extend.js
```diff
@@ -7,10 +7,6 @@
...
  */
 export default function extend(dest, ...sources) {
-    sources.forEach(source => {
-        Object.keys(source).forEach(key => {
-            dest[key] = source[key];
-        });
-    });
-    return dest;
+    console.warn('util.extend is deprecated; use Object.assign');
+    return Object.assign(dest, ...sources);
```
thijstriemstra: leave the existing implementation as it is.

katspaugh: @thijstriemstra I think we should discuss solutions rather than stating one's opinion in an imperative way. Any reasons to keep the old implementation?

thijstriemstra: Yes, we do not remove existing implementations before deprecating it for a couple releases, no matter what the code is. Don't you agree? Is there a huge problem with that? Or does it **have** to go? And I don't know what you mean with "stating one's opinion in an imperative way."

mustafa0x: Absolute adherence to a _certain way_ without regard to reasons for and against seems a bit bureaucratic =)

In programming, implementation is rarely significant — what matters is the interface. https://google.com/search?q=programming+interface+vs+implementation

thijstriemstra: I do not agree, but I'll leave the decision up to @katspaugh.

katspaugh: I have no strong preference. It’s gonna be removed anyway.

thijstriemstra: > it would be less work to keep it at this point.

I have no problem doing the Right Thing and do more work.

>  agree it would be better not to touch it until removal.

Well, let's stick to that then. @mustafa0x please make the changes or @katspaugh please merge it. I am staying away from this pull request.

katspaugh: Cool 👍 
Thanks, @thijstriemstra. I think everyone appreciates your ideal.
And thanks @mustafa0x for the PR.

mustafa0x: Thanks @katspaugh and @thijstriemstra!

Is there anything I need to do here?

katspaugh: If you could restore the old implementation...

mustafa0x: Done...

katspaugh: Thanks... 😁

PR #1829, opened 2019-12-13
regions plugin: add option to select specific channel to draw on

isabyr:
**Now there can be specified `channelIdx` param to draw region on specific channel**

```
plugins: [
            WaveSurfer.regions.create({
                regions: [
                    {
                        start: 1,
                        end: 5,
                        loop: false,
                        color: 'hsla(400, 100%, 30%, 0.5)'
                    },
                    {
                        start: 7,
                        end: 9,
                        loop: false,
                        color: 'hsla(200, 50%, 70%, 0.4)',
                        channelIdx: 1
                    }
                ]
            })
        ]
```
Height of region is controlled by 2 parameters, `regionHeight` and `marginTop`:
```
this.regionHeight = '100%';
this.marginTop = '0px';
let channelCount = this.wavesurfer.backend.ac.destination.channelCount;
if (channelIdx >= 0 && channelIdx < channelCount) {
    this.regionHeight = Math.floor((1 / channelCount) * 100) + '%';
    this.marginTop = this.wavesurfer.getHeight() * channelIdx + 'px';
}
```
Default behavior(if parameter is omitted) draws whole height.
### Screenshot
image

### DIFF
--- a/src/plugin/regions.js
+++ b/src/plugin/regions.js
@@ -48,6 +48,17 @@ class Region {
         this.scrollSpeed = params.scrollSpeed || 1;
         this.scrollThreshold = params.scrollThreshold || 10;
 
+        // select channel ID to set region
+        let channelIdx =
+            params.channelIdx == null ? -1 : parseInt(params.channelIdx);
+        this.regionHeight = '100%';
+        this.marginTop = '0px';
+        let channelCount = this.wavesurfer.backend.ac.destination.channelCount;
+        if (channelIdx >= 0 && channelIdx < channelCount) {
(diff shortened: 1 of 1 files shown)


src/plugin/regions.js
```diff
@@ -48,6 +48,17 @@ class Region {
         this.scrollSpeed = params.scrollSpeed || 1;
         this.scrollThreshold = params.scrollThreshold || 10;
 
+        // select channel ID to set region
+        let channelIdx =
+            params.channelIdx == null ? -1 : parseInt(params.channelIdx);
```
thijstriemstra: can you also document this new parameter?

PR #1833, opened 2019-12-22
deprecate util.frame

mustafa0x:
Funny that this is even in. The README says:

> wavesurfer.js works only in modern browsers supporting Web Audio.

Anything that supports Web Audio supports rAF.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,6 +20,7 @@ wavesurfer.js changelog
   - Disable drag selection before enabling it (#1698)
   - Add `channelIdx` option to select specific channel to draw on (#1829)
 - Cursor plugin: fix time visibility (#1802)
+- Deprecate `util.frame` (i.e. rAF shim) (#1833)
 
 3.2.0 (24.10.2019)
 ------------------
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,6 +20,7 @@ wavesurfer.js changelog
   - Disable drag selection before enabling it (#1698)
   - Add `channelIdx` option to select specific channel to draw on (#1829)
 - Cursor plugin: fix time visibility (#1802)
+- Deprecate `util.extend` (#1825)
 - Deprecate `util.frame` (i.e. rAF shim) (#1833)
 
 3.2.0 (24.10.2019)
--- a/src/util/frame.js
+++ b/src/util/frame.js
@@ -1,5 +1,3 @@
-import reqAnimationFrame from './request-animation-frame';
-
 /**
  * Create a function which will be called at the next requestAnimationFrame
  * cycle
@@ -9,5 +7,8 @@ import reqAnimationFrame from './request-animation-frame';
  * @return {func} The function wrapped within a requestAnimationFrame
  */
 export default function frame(func) {
-    return (...args) => reqAnimationFrame(() => func(...args));
+    console.warn(
+        'util.frame is deprecated; use window.requestAnimationFrame instead'
+    );
+    return (...args) => window.requestAnimationFrame(() => func(...args));
 }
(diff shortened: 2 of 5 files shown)


thijstriemstra:
thanks @mustafa0x, can you also add a changelog entry?

thijstriemstra:
Forgot to mention; the other deprecation you added can also be added to the changelog.

CHANGES.md
```diff
@@ -20,6 +20,7 @@ wavesurfer.js changelog
   - Disable drag selection before enabling it (#1698)
   - Add `channelIdx` option to select specific channel to draw on (#1829)
 - Cursor plugin: fix time visibility (#1802)
+- Deprecate `util.frame` (i.e. rAF shim) (#1833)
```
thijstriemstra: i would use full name instead of rAF

src/util/request-animation-frame.js
```diff
@@ -11,5 +11,5 @@ export default (
     window.mozRequestAnimationFrame ||
     window.oRequestAnimationFrame ||
     window.msRequestAnimationFrame ||
-    ((callback, element) => setTimeout(callback, 1000 / 60))
```
thijstriemstra: no need to change the existing implementation.

src/util/frame.js
```diff
@@ -9,5 +7,8 @@ import reqAnimationFrame from './request-animation-frame';
  * @return {func} The function wrapped within a requestAnimationFrame
  */
 export default function frame(func) {
-    return (...args) => reqAnimationFrame(() => func(...args));
+    console.warn(
+        'util.frame is deprecated; use window.requestAnimationFrame instead'
+    );
+    return (...args) => window.requestAnimationFrame(() => func(...args));
```
thijstriemstra: no need to change the existing implementation.

src/util/request-animation-frame.js
mustafa0x: Don't be anal man 😂 

It was an unused arg.

thijstriemstra: Love it or hate it, I don't care.

mustafa0x: Mostly bemusement at the insistence for keeping an unused arg around.

mustafa0x: This isn't changing implementation, it's removing an unused arg. 🤷‍♂ 

Nothing else changed.

src/util/frame.js
mustafa0x: 😂 we should keep the dep on a deprecated module? 😂

src/util/request-animation-frame.js
thijstriemstra: There's a lot of quirks in old code, who knows, maybe it's used in some weird way somewhere, by someone. If we keep it, that will definitely **not** cause issues. No need to introduce a change to this function.

src/util/frame.js
thijstriemstra: What's up with the 😂 😂 ? I am trying to give you guidance and you think it's funny? Why the disrespect?

mustafa0x: Disrespect?! Not at all! =)

mustafa0x:
Well, this was fun! =)

I'll excuse myself =)

src/util/frame.js
thijstriemstra: And what does that even mean 'the dep on a deprecated module'? You can replace usage of util.frame in the code and simply leave util.frame alone, besides adding a deprecation warning.

src/util/request-animation-frame.js
mustafa0x: In a language like JS that's literally impossible. =)

Issue #1838, opened 2020-01-01
Regions plugin: improve stylability of region handles


Frank3K:
In PR  #1798 the ability to customize the color of both (left and right) region handles has been added. This is a welcome change, which give us more tools to improve the UX of the handles.

Another improvement for the region handles would be to have more control over the width of the handles. Currently, the width is defined [as follows](https://github.com/katspaugh/wavesurfer.js/blob/17d6731e01933d85a786df3f30725b80a3fc7b62/src/plugin/regions.js#L191-L192):

```js
  width: '1%',
  maxWidth: '4px',
```

With regions of different lengths, this results in handles of different widths. Since version 3.3.0 the region handles are now more visible than before, and (although it's a personal preference) they do not look "that good" with all the differing widths. I think it would look better if all handles would be the same width. Furthermore, very small handles (<1px) can't be properly selected (at least in Firefox 71).

Since PR #1798 has made a start with `handleStyle`, it could be extended to set any given CSS property. This could be done using `Object.assign`:

```js
            const css = {
                cursor: 'col-resize',
                position: 'absolute',
                top: this.marginTop,
                width: '1%',
                maxWidth: '4px',
                height: this.regionHeight
            };
```

Could be extended like:

``` js
  const cssLeft = Object.assign({}, css, this.handleStyle.left);
  const cssRight = Object.assign({}, css, this.handleStyle.right);
```

@Reached: any thoughts on this?

Frank3K:
Furthermore, say one would like to just use CSS classes for styling, it might be beneficial to disable the JS styling (since this has precedence over the CSS styling).

Proposals:

1.  When `this.handleStyle.left` (or `.right`) is set to the string `none`, no CSS is applied to the handles. When `this.handleStyle.left` is an object, then the above `Object.assign` code can be used.
2. Make it possible to do something similar for a region (`this.regionStyle`?).
3. Add properties `this.handleClass.left` (or `handleClasses`?) and `this.handleClass.right` which can be used to dynamically set additional classes on the handles. 
4. Add a property `this.regionClass` (or `regionClasses`?) which can be used to dynamically set additional classes on a region.

@thijstriemstra What do you think? I can do a PR if you're OK with the proposal. Since I'm a new user to this library, I'm not sure if all of the above fits the current philosophy and style.

Frank3K:
The suggestion of the first comment and the first point of the second comment are included in PR #1839.

thijstriemstra:
[...]

PR #1839, opened 2020-01-02
regions plugin: improve handle style support

Frank3K:

This PR improves the stylability of the region handles. Previously (since version 3.3.0) it was only possible to set a background-color. This feature was made possible by #1798. Other styling properties could not be set on the region handles.

This PR makes it possible to:

1. Set any styling property on a handle. As discussed [here](https://github.com/katspaugh/wavesurfer.js/issues/1838#issue-544401172).
2. Disable all JS-styling of a handle. This way styling can be done in CSS. As discussed [in point 1 here](https://github.com/katspaugh/wavesurfer.js/issues/1838#issuecomment-570312988).

None, as far as I can see. It should be compatible with v3.3.0 / PR #1798, but please check my comment [there](https://github.com/katspaugh/wavesurfer.js/pull/1798#issuecomment-570084267).

None, as far as I can see. It should be compatible with v3.3.0 / PR #1798, but please check my comment [there](https://github.com/katspaugh/wavesurfer.js/pull/1798#issuecomment-570084267).

### Todos/Notes:

Issue #1838 contains more suggestions for further improvements.

- Issue #1838.

### DIFF
--- a/src/plugin/regions.js
+++ b/src/plugin/regions.js
@@ -30,12 +30,8 @@ class Region {
         this.loop = Boolean(params.loop);
         this.color = params.color || 'rgba(0, 0, 0, 0.1)';
         this.handleStyle = params.handleStyle || {
-            left: {
-                backgroundColor: 'rgba(0, 0, 0, 1)'
-            },
-            right: {
-                backgroundColor: 'rgba(0, 0, 0, 1)'
-            }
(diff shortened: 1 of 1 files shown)


katspaugh:
Thanks, @Frank3K and @thijstriemstra! 🎉

PR #1844, opened 2020-01-08
regions plugin: add support for a context menu event on a region

### LINKED ISSUE #1842: regions plugin: add support for contextmenu event on regions
In the current version of waversurfer.js (3.3.0) there are several handlers for events on regions: `mouseenter`, `mouseleave`, `click` and `dblclick` ([source](https://github.com/katspaugh/wavesurfer.js/blob/17d6731e01933d85a786df3f30725b80a3fc7b62/src/plugin/regions.js#L318-L338)).

I would like to show a menu when right-clicking on a region. This could be made possible using the following lines:

```js
        this.element.addEventListener('contextmenu', e => {
            e.preventDefault();
            this.fireEvent('click', e);
            this.wavesurfer.fireEvent('region-contextmenu', this, e);
        });
```

If wanted, I can do a PR for this.

Frank3K:
This PR adds support for a context menu event on a region in the regions plugin. It contains an option to disable the native context menu from being opened (`preventContextMenu`, defaults to false).

Fixes #1842.

The functionality was approved by @thijstriemstra [here](https://github.com/katspaugh/wavesurfer.js/issues/1842#issuecomment-570937396).

### DIFF
--- a/src/plugin/regions.js
+++ b/src/plugin/regions.js
@@ -618,6 +618,8 @@ class Region {
  * @property {boolean} resize=true Allow/disallow resizing the region.
  * @property {string} [color='rgba(0, 0, 0, 0.1)'] HTML color code.
  * @property {?number} channelIdx Select channel to draw the region on (if there are multiple channel waveforms).
+ * @property {?object} handleStyle A set of CSS properties used to style the left and right handle.
+ * @property {?boolean} preventContextMenu=false Determines whether the context menu is prevented from being opened.
  */
 
 /**
--- a/src/plugin/regions.js
+++ b/src/plugin/regions.js
@@ -45,6 +45,11 @@ class Region {
         this.scroll = params.scroll !== false && ws.params.scrollParent;
         this.scrollSpeed = params.scrollSpeed || 1;
         this.scrollThreshold = params.scrollThreshold || 10;
+        // Determines whether the context menu is prevented from being opened.
+        this.preventContextMenu =
+            params.preventContextMenu === undefined
+                ? false
+                : Boolean(params.preventContextMenu);
 
         // select channel ID to set region
         let channelIdx =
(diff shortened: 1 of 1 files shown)


src/plugin/regions.js
```diff
@@ -45,6 +45,11 @@ class Region {
         this.scroll = params.scroll !== false && ws.params.scrollParent;
         this.scrollSpeed = params.scrollSpeed || 1;
         this.scrollThreshold = params.scrollThreshold || 10;
+        // Determines whether the context menu is prevented from being opened.
+        this.preventContextMenu =
```
thijstriemstra: the `preventContextMenu` needs to be documented (see other params).

src/plugin/regions.js
```diff
@@ -341,6 +348,14 @@ class Region {
             this.wavesurfer.fireEvent('region-dblclick', this, e);
         });
 
+        this.element.addEventListener('contextmenu', e => {
+            if (preventContextMenu) {
+                e.preventDefault();
+            }
+            this.fireEvent('contextmenu', e);
+            this.wavesurfer.fireEvent('region-contextmenu', this, e);
```
thijstriemstra: why fire 2 events here?

Frank3K: This is in line with the other events (e.g. `dblclick`): one event is on the region itself and one is proxied by the main waveSurfer object.

src/plugin/regions.js
Frank3K: I'm not sure what you mean. The "documentation" is in the comment on line 48. Do I need to document it (also?) somewhere or do you mean that the comment needs to be improved?

thijstriemstra: see params documentation at https://github.com/katspaugh/wavesurfer.js/blob/2943113705caaae1ea9c23fd9516ac1c1d0c9b61/src/plugin/regions.js#L595

Frank3K: Done. Also included the parameter `handleStyle` (PR #1798), which was also missing from the docs.

PR #1845, opened 2020-01-09
regions plugin: handle position fix while using with channelIdx param

yesseyev:
There was issue of region's handle positioning in render. Handle style was previously changed due to channelIdx feature implementation. Region handle styles (top, height) where changed to original values.

### DIFF
--- a/src/plugin/regions.js
+++ b/src/plugin/regions.js
@@ -175,10 +175,10 @@ class Region {
             const css = {
                 cursor: 'col-resize',
                 position: 'absolute',
-                top: this.marginTop,
+                top: '0px',
                 width: '1%',
                 maxWidth: '4px',
-                height: this.regionHeight,
+                height: '100%',
                 backgroundColor: 'rgba(0, 0, 0, 1)'
             };
 


thijstriemstra:
what do you think about this PR @Frank3K?

Frank3K:
@thijstriemstra I'm not sure what the issue is that is reported by @yesseyev. This PR reverts changes that (I think) are needed for d8b768c1ea3902a44a5c29ef2bc9d8b398ec3f0e / #1829.

If `channelIdx` is not used, the `top` and `height` are just as expected. See [lines 52-53](https://github.com/katspaugh/wavesurfer.js/blob/cf295db17d5726ad2522cb133b711184779c5120/src/plugin/regions.js#L52-L53) and [lines 178-181](https://github.com/katspaugh/wavesurfer.js/blob/cf295db17d5726ad2522cb133b711184779c5120/src/plugin/regions.js#L178-L181).

yesseyev:
@Frank3K 
Problem is when you are using **channelIdx** param in _Region_, it draws regions on separate channels, however their handlers located somewhere else (not on the sides of region) and it makes impossible to resize the region. My PR solves this issue.

Frank3K:
@yesseyev Thanks for the explanation, now I understand the issue. Since the handles are children of the region element, they indeed do not need to have the calculated top / height. The 0 px / 100% values ensure they are the same height as the region.

Good fix!

PR #1858, opened 2020-02-07
Regions plugin: fix channelCount

yesseyev:
There is issue with assignment of channelCount in regions plugin:

`let channelCount = this.wavesurfer.backend.ac.destination.channelCount;`

which is always equals to 2, even though audio is mono.
That leads to incorrect region height assignment and positioning. 

Right way to acquire channelCount is to use WebAudio backend's buffer

`let channelCount = this.wavesurfer.backend.buffer.numberOfChannels;`

However, it requires to use add regions after wavesurfer's **'ready'** event. So I changed assignment to:
`let channelCount = this.wavesurfer.backend.buffer != null ? this.wavesurfer.backend.buffer.numberOfChannels : -1;`
and left posibility to add regions before wavesurfer gets ready if it channelIdx param is not specified.


thijstriemstra:
thanks @yesseyev Did you test it with the other backends? can you also add a changelog entry?

thijstriemstra:
also, there are some eslint errors: https://travis-ci.org/katspaugh/wavesurfer.js/builds/647249806#L886

thijstriemstra:
ping @yesseyev

PR #1889, opened 2020-03-25
Add support for secondary colors for split channels

riceboyler:
For my company's use of wavesurfer, we need to specify different colors for each channel in Split Channels (to help use as a key along with transcripts). This adds a second set of canvases and waves (because I couldn't figure out how to do it without creating new canvases and waves) within the canvasentry. Then if the colors are specified, it will apply the second channel colors via new params: `waveColor2` and `progressColor2`.

There should be no breaking changes, only addition of new params.

I tested it as much as I could. As long as `splitChannels` is false, these changes do absolutely nothing. If `splitChannels` is true and colors aren't specified, it picks up the default colors set in wavesurfer.js

### Todos/Notes:
Add tests, and if needed, add getters and setters.

N/A


riceboyler:
A colorful (if gaudy) preview:
image

riceboyler:
Looks like I've got some work to do. I'll be working on this to clean it up and get that test coverage up.

katspaugh:
Hi @riceboyler, thanks for the contribution. That's a nice idea!

Do you think we could avoid repeating the same code for each channel? Maybe by using a loop?
The colours could be specified as an array. E.g. `waveColor: [ 'red', 'blue' ]`. Or even in a new param `channelColors` to not overload the `waveColor` type.

riceboyler:
I'm ashamed that I hadn't even thought of going that direction. Let me take a stab at it today.

riceboyler:
I'm gonna do:
```javascript
channelColors = {
  waveColors: [],
  progressColors: []
}
```
I think that's a clearer expectation in the code. Thoughts?

katspaugh:
Nothing to be ashamed of! Coding is an iterative process ;)

Yeah, of course, it has to be an object, I forgot about the progress colours.

riceboyler:
OK, I switched to the channelColors as above and while I was digging into it, realized that it wasn't actually rendering the second channel at all. Well, it was, it was just rendering it off of the canvas (thanks to the offsetY still being set, even though I was hard setting the height of the second channel canvas in its definition).

So, that is fixed as well. Note that for right now, this only works with waves (not bars). I'll attempt to work over the weekend on getting this working on bars as well.

katspaugh:
Cool, thanks!

There's still significant code duplication though. What I meant with a loop was something like:

```
  for (let i = 0; i < channels; i++) {
    drawChannel(this.channelCtx[i], channelColors.waveColors[i]);
    drawProgress(this.progressCtx[i], channelColors.progressColors[i]);
  }
```

And the same kind of loop for the init, clear etc.  So instead of having `channel1Ctx` and `channel2Ctx`, it should be an array.

thijstriemstra:
> There should be no breaking changes, only addition of new params.

This is obviously not the case. I don't like how the API is changed here just to accommodate a small feature..

riceboyler:
> > There should be no breaking changes, only addition of new params.
> 
> This is obviously not the case. I don't like how the API is changed here just to accommodate a small feature. It seems like overkill.
> 
> > This adds a second set of canvases and waves (because I couldn't figure out how to do it without creating new canvases and waves)
> 
> We should figure out a way to do this without adding **even more** canvas elements.
> 
> I would extend the CanvasEntry class, and use that subclass inside the MultiCanvas renderer. Once you have that working we can figure out how to 'activate' this new subclass and make it available as an option.

I actually did extend the CanvasEntry class rather than adding Canvases because I realized that was the better way. 

A little feedback from a potential contributor: @katspaugh has been extremely friendly and encouraging. @thijstriemstra your comments (though probably not intended as such) came across as "why did you even bother, you're messing things up". I've been developing for the web for 25 years now and I can assure you, you always catch more flies with honey (@katspaugh ) than with vinegar (@thijstriemstra ). Just something to think about as your review and discuss with people who want to contribute to the project.

I'll be going through today and working on the looping mechanism, as you're entirely correct, that would be better.

katspaugh:
Thanks @riceboyler. Looking forward to seeing an improved version.

Yeah, @thijstriemstra's comments can come across as a bit condescending. I'm sure you were just being efficient and straightforward. However, let's try to cultivate a friendlier atmosphere in our reviews. I struggle with this myself all the time. For me personally, at work, this once became a real problem as a few people got really upset because of the harsh reviews. On the internet, it's even more difficult as we cannot even smile to each other. 😉

riceboyler:
I certainly don't want to make it seem like his comments were the only ones I've seen like that. Trust me, I've written my share of "terse" reviews that didn't come across well. That's why I like providing feedback. If feedback isn't shared, how can anything get better?

A quick note about the for loops: I'm going to try to make the calls using those as much as possible, but in the case of drawer.canvasentry.js, it inherently is going to need a little bit of change to the init functions, as they'll need to be instantiated as arrays rather than as a pure object. In looking at it, I'm thinking that it might make more sense for me to create a separate drawer.splitEntries.js file that would closely mirror drawer.canvasentry.js in functionality, but be natively array based rather than assume a single object. Then in drawer.multicanvas.js, we can import both and let treeshaking resolve things there. 

I'd love your thoughts about this approach, as opposed to modifying canvasentry more directly.

riceboyler:
Actually, as I dig into it more, I'm realizing that maybe it makes more sense to manage those things in multicanvas instead, and leave the canvasentry alone as much as possible. This is a problem with multiple valid solutions it would appear.

riceboyler:
Sorry about the spam here, kinda thinking out loud. The relative problem with both of the approaches above is that the different "channels" actually need to be provided somewhat different information (i.e. different peaks, different fill colors). 

That being the case, I think it probably makes more sense for me to look at creating a separate module that uses a base array. I'll be working on that.

katspaugh:
Sounds good! 👍

thijstriemstra:
> @thijstriemstra your comments (though probably not intended as such) came across as "why did you even bother, you're messing things up". 

My first language isn't English. I was not trying to sound like that all. I'm not an asshole. I was trying to put it in a way that didn't make it sound like all your work was for nothing, but I guess I failed, sorry. 

> I've been developing for the web for 25 years now and I can assure you, 

Same here.

thijstriemstra:
> This is a problem with multiple valid solutions it would appear.

Yep, and if you have any questions, let me know, we're here to help.

`MultiCanvas` is the default implementation and it uses `CanvasEntry`. I think MultiCanvas should use the subclass of CanvasEntry I was talking about when the `splitChannels` option is enabled.

riceboyler:
@katspaugh @thijstriemstra This has a new `splitentries` drawer. It's not complete yet, though functional. I'm going to go back through and undo any and all changes I made to the drawer.canvasentry.js file. I'd love feedback. Thank you both!

src/drawer.canvasentry.js
```diff
@@ -103,11 +104,11 @@ export default class CanvasEntry {
         // decimal between 0 and 1
         this.start = this.wave.offsetLeft / totalWidth || 0;
         this.end = this.start + elementWidth / totalWidth;
+        const elementSize = { width: `${elementWidth}px` };
```
riceboyler: I'm gonna leave this change in because I prefer template literals to string concatenation.

katspaugh:
Looks much nicer! `drawer.splitchannels` and `drawer.enrty` still share a lot of code, though.

I guess if we break down the components, it should look like this:

 1. A module that draws an individual waveform
 2. A module that renders instances of Module 1 for each channel in an array – or just one Module 1 for a single waveform (an array of 1 element)
 3. A module to handle multicanvas stuff – should just use Module 2 without any `if`s.

I hope I'm making sense.

example/split-channels/index.html
```diff
@@ -70,6 +70,25 @@ <h3>How to Enable Split Channels</h3>
...
+                <p>
+                    To set these colors, use the <code>channelColors</code> property defined with <code>waveColors</code> and <code>progressColors</code>.
+                </p>
+                <p>
+<pre><code>
+    const wavesurfer = WaveSurfer.create({
+        container: document.querySelector('#wave'),
+        splitChannels: true,
+        channelColors: {
+            waveColors: ['red', 'green'],
```
thijstriemstra, line 88: can you also use a hex color value here to demonstrate that works as well?

src/drawer.multicanvas.js
```diff
@@ -90,6 +99,14 @@ export default class MultiCanvas extends Drawer {
          * @type {number}
          */
         this.barRadius = params.barRadius || 0;
+
+        /**
+         * A helper to determine whether this uses split channels
```
thijstriemstra, line 104: can you expand a little more in this docstring and mention it's related to the `splitChannels` option (and `channelColors` maybe as well)?

thijstriemstra: this would also be a good public var IMO.

src/drawer.multicanvas.js
```diff
@@ -212,6 +234,24 @@ export default class MultiCanvas extends Drawer {
         this.canvases.push(entry);
     }
 
+    addCanvases() {
```
thijstriemstra, line 237: can you add a docstring for this new method?

src/drawer.splitchannels.js
```diff
@@ -0,0 +1,460 @@
+/**
+ * @since 3.0.0
```
thijstriemstra, line 2: this version nr needs to be updated (see changelog for nxt version nr).

src/drawer.splitchannels.js
```diff
@@ -0,0 +1,460 @@
...
+ */
+
+import style from './util/style';
+import getId from './util/get-id';
+
+/**
+ * The `SplitEntry` class represents an element consisting of an array of wave `canvas`
+ * and an (optional) array of  progress wave `canvas`.
+ *
+ * The `MultiCanvas` renderer uses one or more `CanvasEntry` instances to
```
thijstriemstra, line 12: I would remove this sentence.

example/split-channels/index.html
riceboyler: Absolutely.

src/drawer.multicanvas.js
riceboyler: Sure. I can make this public and improve the documentation.

src/drawer.multicanvas.js
riceboyler: Absolutely. I meant to go back and update this, and just forgot.

src/drawer.splitchannels.js
riceboyler: Ooops, I'll update.

src/drawer.splitchannels.js
riceboyler: Good point. I'll remove.

thijstriemstra:
> drawer.splitchannels and drawer.enrty still share a lot of code, though.

Agreed. Why not use a subclass of CanvasEntry and add/override any additional code you need?

thijstriemstra:
ping @riceboyler we'd like to include this in the v4.0.0 release if possible..

riceboyler:
I'll get back to working on this tonight. Got moved to a different project so I lost focus on this in my work day.

thijstriemstra:
Awesome, thanks!

thijstriemstra:
Looks like someone else also decided to implement this feature (!), see pr #1947

katspaugh:
Sorry @riceboyler, #1947 offered a preferred implementation. I hope you like it too. Thanks a ton for starting this. I failed to see the solution could be much simpler and gave you somewhat misleading advice.
Please feel welcome to contribute again!

riceboyler:
Yeah, the solution in #1947 is better than this. Glad I had to wait due to moving!

PR #1890, opened 2020-03-26
Use requestAnimationFrame for clearWave

### LINKED ISSUE #1884: Sometimes peaks are drawn twice, because of requestAnimationFrame timing
While testing semi-transparent bars (i.e. `waveColor: '#FFFFFF55'`) in a playlist context, I noticed they were more opaque at random times when loading a new track. Turns out they're rendered twice, one on top of the other, suggesting something isn't right in the clearing process.

Upon further investigation, I saw that for some reason [`clearWave`](https://github.com/katspaugh/wavesurfer.js/blob/master/src/drawer.multicanvas.js#L262) and [`drawBars`](https://github.com/katspaugh/wavesurfer.js/blob/master/src/drawer.multicanvas.js#L279) are called twice. The intended order of execution should be, tracing calls, something like:

* `clearWave` → `drawBars` → `clearWave` → `drawBars`

But `drawBars` is wrapped inside a `requestAnimationFrame` (via [`prepareDraw`](https://github.com/katspaugh/wavesurfer.js/blob/master/src/drawer.multicanvas.js#L460)) and `clearWave` is not. So it may happens that `drawBars` gets delayed, resulting in a double draw:

* `clearWave` → `clearWave` → `drawBars` → `drawBars`

I managed to patch it using `requestAnimationFrame` for `clearWave` too in [`drawer.multicanvas.js:262`](https://github.com/katspaugh/wavesurfer.js/blob/master/src/drawer.multicanvas.js#L262):

```js
clearWave() {
    util.frame(() => {
        this.canvases.forEach(entry => entry.clearWave());
    })();
}
```

Now my bars never get more opaque, however:

1. It's not clear why `clearWave` and `drawBars` are called twice in the first place. ~~May also be something wrong in my specific application.~~ I checked the playlist example here, and the calls are the same. However `requestAnimationFrame` never causes a sufficient delay for the issue to occur. It may be that in my application I use it for other stuff too, while the example is pretty clean instead.

2. I'm not 100% sure wrapping `clearWave` in `requestAnimationFrame` is without any side effects.

**Wavesurfer.js version(s):** Latest from NPM: 3.3.1
**Browser and operating system version(s):** Firefox 74.0 on Ubuntu 19.10

lorenzos:
As discussed [here](https://github.com/katspaugh/wavesurfer.js/issues/1884#issue-586311889), this is a workaround for #1884.

fixes #1884

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 wavesurfer.js changelog
 =======================
 
+3.3.2 (unreleased)
+
+- Use requestAnimationFrame for clearWave (#1884)
+
 3.3.1 (13.01.2020)
 ------------------
 
--- a/src/drawer.multicanvas.js
+++ b/src/drawer.multicanvas.js
@@ -260,7 +260,9 @@ export default class MultiCanvas extends Drawer {
      * Clear the whole multi-canvas
      */
     clearWave() {
-        this.canvases.forEach(entry => entry.clearWave());
+        util.frame(() => {
+            this.canvases.forEach(entry => entry.clearWave());
+        })();
     }
 
     /**


thijstriemstra:
thanks @lorenzos, can you also add this to the changelog?

PR #1893, opened 2020-03-28
Add reloadMediaElement param and fix region loop

aimkiray:
In the example `Annotations`, after shift-click region, and then shift-click other region, you can not jump out of the current loop. Checking current time, continuing the loop if it is in the current region, and stopping the loop if it is not, seems to solve the problem.

When the `url` param for `wavesurfer.load()` is `MediaElement` in `canplay` state, there is no need for empty and load, I suggest adding `reloadMediaElement` parameter.

- `reloadMediaElement=false`: When MediaElement is in the canplay state, there is no need to reload.

- After drag the region, audio will still pause in its previous position.

### DIFF
--- a/src/plugin/regions.js
+++ b/src/plugin/regions.js
@@ -138,8 +138,16 @@ class Region {
      * @param {number} start Optional offset to start playing at
      * */
     playLoop(start) {
-        this.play(start);
-        this.once('out', () => this.playLoop());
+        const s = start || this.start;
+        this.play(s);
+        this.once('out', () => {
+            if (
+                this.wavesurfer.getCurrentTime() >= this.start &&
+                this.wavesurfer.getCurrentTime() <= this.end
+            ) {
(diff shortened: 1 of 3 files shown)


thijstriemstra:
there are multiple changes going on here, can you limit a pull request to a plugin, or core? you're also changing things in the core codebase to accommodate for fixes in the plugins, not sure if this is way to go here.

thijstriemstra:
ping @aimkiray

src/plugin/regions.js
```diff
@@ -138,8 +138,16 @@ class Region {
      * @param {number} start Optional offset to start playing at
      * */
     playLoop(start) {
```
marizuccara, line 140: Hi, reproducing this code, the method is still executed in infinite loop

aimkiray:
Sorry, I've been a bit busy recently.
I will submit a new PR later.

Issue #1910, opened 2020-04-10
Elan throws error if tiers is not provided


bhogan-mitre:
Elan parameter `tiers` is [documented as optional](https://wavesurfer-js.org/api/typedef/index.html#static-typedef-ElanPluginParams), although it throws an error if it is not provided.

The problematic segment is this [loop](https://github.com/katspaugh/wavesurfer.js/blob/master/src/plugin/elan.js#L235-L242) over `indeces` that refers to `this.params.tiers`:
```
        indeces.forEach(index => {
            const tier = tiers[index];
            const th = document.createElement('th');
            th.className = 'wavesurfer-tier-' + tier.id;
            th.textContent = tier.id;
            th.style.width = this.params.tiers[tier.id];
            headRow.appendChild(th);
        });
```

### Wavesurfer.js version(s):
Latest via https://unpkg.com/wavesurfer.js

### Browser and operating system version(s):
Chrome 81.0.4044.92 on macOS 10.15.4

### Code needed to reproduce the issue:

Borrow from [this example](https://wavesurfer-js.org/example/elan/index.html), but instead of:
```
            WaveSurfer.elan.create({
                url: 'transcripts/001z.xml',
                container: '#annotations',
                tiers: {
                    Text: true,
                    Comments: true
                }
            }),
```

Use undefined `tiers`:
```
            WaveSurfer.elan.create({
                url: 'transcripts/001z.xml',
                container: '#annotations',
                tiers: undefined
            }),
```

thijstriemstra:
> Elan parameter tiers is documented as optional, although it throws an error if it is not provided.

If it's not optional, feel free to open a pull request to update the documentation. If you want to make it optional, go ahead and open a pull request. Thanks!

bhogan-mitre:
PR merged, closing this.

PR #1919, opened 2020-04-17
Elan plugin: optional params.tiers (#1910)

bhogan-mitre:

This fixes a bug where undefined Elan `params.tiers` would lead to an error, thus making the parameter truly optional. #1910 

N/A

N/A

### DIFF
--- a/src/plugin/elan.js
+++ b/src/plugin/elan.js
@@ -237,7 +237,6 @@ export default class ElanPlugin {
             const th = document.createElement('th');
             th.className = 'wavesurfer-tier-' + tier.id;
             th.textContent = tier.id;
-            th.style.width = true;
             headRow.appendChild(th);
         });
 
--- a/src/plugin/elan.js
+++ b/src/plugin/elan.js
@@ -237,7 +237,7 @@ export default class ElanPlugin {
             const th = document.createElement('th');
             th.className = 'wavesurfer-tier-' + tier.id;
             th.textContent = tier.id;
-            th.style.width = this.params.tiers[tier.id];
+            th.style.width = true;
             headRow.appendChild(th);
         });
 
--- a/src/plugin/elan.js
+++ b/src/plugin/elan.js
@@ -237,6 +237,7 @@ export default class ElanPlugin {
             const th = document.createElement('th');
             th.className = 'wavesurfer-tier-' + tier.id;
             th.textContent = tier.id;
+            if (this.params.tiers) { th.style.width = this.params.tiers[tier.id]; }
             headRow.appendChild(th);
         });
 


src/plugin/elan.js
```diff
@@ -237,7 +237,7 @@ export default class ElanPlugin {
             const th = document.createElement('th');
             th.className = 'wavesurfer-tier-' + tier.id;
             th.textContent = tier.id;
-            th.style.width = this.params.tiers[tier.id];
+            th.style.width = true;
```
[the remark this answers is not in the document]
bhogan-mitre: Thanks for taking a look.  I agree that does not make a lot of sense, but that is the value that it takes under the existing behavior when `tiers` is specified.

This [snippet from the example](https://github.com/katspaugh/wavesurfer.js/blob/master/example/elan/app.js#L17-L24) sets the value under specific ids to true.
```
            WaveSurfer.elan.create({
                url: 'transcripts/001z.xml',
                container: '#annotations',
                tiers: {
                    Text: true,
                    Comments: true
                }
            }),
```

Then `this.params.tiers[tier.id]` ends up with the value `true`.

I was aiming to avoid a larger refactor of this plugin, and just replicate the existing behavior when `params.tiers` is provided.  Note that `tiers` gets filtered down in line 196 if that parameter exists, `indeces` is only populated for tiers that exist past that filter, and the line 240 in question is called for each value of `indeces`.

Another approach would be to just delete line 240.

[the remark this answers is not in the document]
thijstriemstra: > Another approach would be to just delete line 240.

If it doesn't do anything, for sure, let's delete it.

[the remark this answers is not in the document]
bhogan-mitre: That works.  I just removed that line and confirmed the example Elan app still works (with and without the provided params.tiers).

src/plugin/elan.js
```diff
@@ -237,7 +237,6 @@ export default class ElanPlugin {
             const th = document.createElement('th');
             th.className = 'wavesurfer-tier-' + tier.id;
             th.textContent = tier.id;
-            th.style.width = this.params.tiers[tier.id];
```
katspaugh, line 240: How about
```
if (this.params.tiers) { th.style.width = this.params.tiers[tier.id]; }
```
?

bhogan-mitre: Sure.  Again, the best I can tell `th.style.width` is not used and it doesn't matter what we set it to as long as it doesn't throw an error.

I updated the PR with this suggested edit.

katspaugh: Thanks! I don't remember now why it was there. I assume it sets the width of the columns that way. Otherwise they would depend on the contents of the cells.

bhogan-mitre: Right, but the value it gets set to is `true` and I'm not sure what that means in the CSS context of `th.style.width`.  It doesn't appear to harm anything, but wanted to mention it.  This is fine with me as long as it is not throwing errors.

katspaugh: True, in example/elan it's passed as true. Maybe it was meant to accept a CSS value and is wrong in the example? I don't remember. Either way, both deleting it entirely and making it optional work. Let's just keep it optional. 👍

PR #1921, opened 2020-04-19
Plugin inheritance example in 'plugin-system'

### LINKED ISSUE #1914: Implementing a plugin that extends an existing plugin
Hi, I would like to override some methods of an existing plugin, extending a basic one. It is possible?

marizuccara:
Example of how extends a basic plugin to add custom features, keeping library support, since you won't have to download the library locally to modify some features of an existing plugin. 

fixes #1914


example/plugin-system/custom-plugin/CursorCustomPlugin.js
```diff
@@ -0,0 +1,72 @@
+import CursorPlugin from '../../../src/plugin/cursor.js';
+
+export default class CursorCustomPlugin extends CursorPlugin {
+    /**
+     * Cursor plugin definition factory
```
thijstriemstra, line 5: can you change the description of the plugin and explain it's a custom plugin that extends the cursor plugin, and also describe what it does different from cursor plugin?

example/plugin-system/index.html
```diff
@@ -26,7 +26,7 @@
         <script src="../../dist/plugin/wavesurfer.spectrogram.js"></script>
 
         <!-- App -->
-        <script src="app.js"></script>
+        <script type="module" src="app.js"></script>
```
thijstriemstra, line 29: is this necessary?

marizuccara: Yes, I added a motivation in description

marizuccara:
I thought ... if it were possible to remove private methods of plugins, so it will be possible to override all of them

example/plugin-system/custom-plugin/CursorCustomPlugin.js
thijstriemstra: awesome thanks!

example/plugin-system/index.html
```diff
@@ -26,6 +26,7 @@
         <script src="../../dist/plugin/wavesurfer.spectrogram.js"></script>
 
         <!-- App -->
+        <!-- Defining app.js as a module allows you to use 'import' statement, to import CursorPlugin -->
```
thijstriemstra, line 29: this should say CustomCursorPlugin instead, shouldn't it?

thijstriemstra:
> I thought ... if it were possible to remove private methods of plugins, so it will be possible to override all of them

Not sure what you mean..

example/plugin-system/custom-plugin/CursorCustomPlugin.js
```diff
@@ -0,0 +1,77 @@
...
+    _onMouseenter = () => this.showCursor();
+
+    /**
+     * @private
+     * @returns {void}
+     */
+    _onMouseleave = () => this.hideCursor();
+
+    /**
+     * Construct the plugin class. You probably want to use `CursorPlugin.create`
```
thijstriemstra, line 60: This still mentions `CursorPlugin`.

example/plugin-system/custom-plugin/CursorCustomPlugin.js
```diff
@@ -0,0 +1,77 @@
...
+ * the library to modify the basic plugin class methods. This is positive thing if you want to maintain the constant
+ * support of the library.
+ */
+export default class CursorCustomPlugin extends CursorPlugin {
+    /**
+     * Cursor plugin definition factory
+     *
+     * This function overrides the create static method of the CursorPlugin, so you can assign it a name
+     *
+     * @param  {CursorPluginParams} params parameters use to initialise the
```
thijstriemstra, line 15: `use` should be `used`

marizuccara:
> > I thought ... if it were possible to remove private methods of plugins, so it will be possible to override all of them
> 
> Not sure what you mean..

The possibility to extends a basic plugin would be more powerful if it would be possible to extends all methods, and override only those which you want to add changes. But many methods of the basic plugins are private, so I have to copy them in my extended class, because they are not heritable.

thijstriemstra:
> But many methods of the basic plugins are private, so I have to copy them in my extended class, because they are not heritable.

If you think a method should be public, feel free to suggest it in a new pull request. We should prevent code duplication and establish a public API where possible.

thijstriemstra:
ps. @katspaugh I would add @marizuccara to the team, more people the better.

katspaugh:
Sure, good idea! If Maria doesn't mind, I'll add them to the team.

marizuccara:
Oh thanks! I'd be so pleased!

katspaugh:
Cool! Welcome to the team, Maria, and thanks for the contributions!

thijstriemstra:
@katspaugh i tried requesting a review on a pull request from marizuccara but the name doesn't show up in the list, not even when i type it out. any idea why, does Maria need more rights to show up in the reviewers list?

katspaugh:
The invite is still pending:
Screenshot 2020-04-28 at 14 30 40

marizuccara:
> The invite is still pending:
> Screenshot 2020-04-28 at 14 30 40

Accepted! Sorry, I didn't see it!

PR #1926, opened 2020-04-23
Regions plugin: fix click propagation issue

### LINKED ISSUE #1925: Region click is overriden by interaction if it's enabled
### Wavesurfer.js version(s):
3.3.3

Chrome 81 on Windows 10 / Chrome on Android

Full reproduction here:

https://stackblitz.com/edit/js-paffsu?file=index.html

### User behaviour needed to reproduce the issue:
1. Play the reproduction above
2. Click on the region during playback
3. The region will not be played but the playback will seek to where was clicked

The fix is that we now call `e.stopPropagation()` in the click handler for the region to enable the region click to override the default interaction.

iboshkov:
This PR fixes the issue described in Issue #1925 where a click on the region is ignored by the wavesurfer interact click implementation. This behavior can be undesirable and this PR fixes that.

fixes #1925 

None

None

### Todos/Notes:
Possible improvement is to not stop propagation for when the current behavior is desirable.
The way this could be implemented is by passing the browser event with the region in the handler callback and letting the callback implementation can determine if stopping propagation is needed or not.

Issue #1925 

### User behaviour needed to reproduce the issue:

https://stackblitz.com/edit/js-paffsu?file=index.html

1. Play the reproduction above
2. Click on the region during playback
3. The region will not be played but the playback will seek to where was clicked

### DIFF
--- a/src/plugin/regions.js
+++ b/src/plugin/regions.js
@@ -342,6 +342,7 @@ class Region {
         });
 
         this.element.addEventListener('click', e => {
+            e.stopPropagation();
             e.preventDefault();
             this.fireEvent('click', e);
             this.wavesurfer.fireEvent('region-click', this, e);


thijstriemstra:
thanks!

PR #1927, opened 2020-04-25
Fixed destroy method of MediaElementWebAudio backend

### LINKED ISSUE #1923: MediaElementWebAudio destroy method
Hi, since MediaElementWebAudio backend inherits from MediaElement backend, when destroy method is called, it calls the MediaElement backend destroy method, which destroy the media audio tag. But, since MediaElementWebAudio uses WebAudio features, it should be better if we destroy also web audio related stuffs, as disconnecting audio nodes, close audio context, destroy it, etc (the same methods which are used in the WebAudio backend destroy).

marizuccara:
When destroy method is called, instead of destroying only the media element, the audio nodes are disconnected and the audio context is closed. 

The destroy method of the WebAudio backend was splitted, so it calls the new `destroyWebAudio` method to cancel all stuffs related to WebAudio.
The destroy method of MediaElementWebAudio backend overrides that of the MediaElement backend, calling its destroy method to cancel the media element, calling also the new `destroyWebAudio` method.

fixes #1923


thijstriemstra:
thanks @marizuccara looks good to me. Can you also add a changelog entry?

PR #1928, opened 2020-04-26
Removed private methods of plugins and generalized plugins' access

marizuccara:
Removed all private methods of plugin classes, so the plugins can be extended avoiding duplicated code, because private methods can not be inherited. 

Allow to extend also the `Region` class, introducing the possibility to create your `RegionCustom` class that extends the basic one, overriding its methods if necessary. 

Generalized access to plugins object: in some plugin classes, the access to their own params was made from the outside. For example, to access to the `timeline` plugin `duration` parameter, the access was done in this way:

`this.wavesurfer.timeline.params.duration`
instead of:
`this.params.duration`.

Accessing this parameter from the outside made it impossible to call custom plugins with custom names, because to the `wavesurfer` object is assigned a plugin with the name you give it. So, if you for example call it `timelineCustom`, the access to the `duration` parameter from the outside, should be instead `this.wavesurfer.timelineCustom.params.duration`. But in this way you cannot generalize the plugin class.

- Removed private methods from the `cursor.js` and `timeline.js` files.

- In the `timeline` plugin, access to the `duration` parameter from its `params` object.

- Added `export` clause to the `Region` class, so it can be also extended.

- Access to the `getRegionSnapToGridValue` method of the `regions` plugin from the internal `util` object, instead of from the outside.

- Pass to the `Region` class constructor, the `util` object of the `RegionsPlugin` class, so the `getRegionSnapToGridValue` method have not be accessed from outside. 

Issue https://github.com/katspaugh/wavesurfer.js/issues/1914 
PR https://github.com/katspaugh/wavesurfer.js/pull/1921


src/plugin/cursor.js
```diff
@@ -83,7 +83,6 @@ export default class CursorPlugin {
     };
 
     /**
-     * @private
      * @param {object} e Mouse move event
      */
     _onMousemove = e => {
```
thijstriemstra, line 88: the underscore in front of these methods implies it's private. What do you suggest we do with those (without breaking the api)?

marizuccara: Underscore not make the method truly private, it can still be accessed outside the body of class, so it can be inherited. But yes, it is a convention and a "suggestion" that is treated as a private method.
So we can:
 - or leave the underscore, because it can be inherited
- or follow the formal conventions, and remove all underscores in this plugin. Since it was considered private, these methods were only used in this class, so the modifications are limited only in this class.

marizuccara:
Another thing. The `Region` class can be moved to a separate file? This is because if we export both `Region` and `RegionsPlugin` class, we have errors when try to create the plugin in this way:
`Wavesurfer['regions'].create(options)`
bacause `Wavesurfer['regions']' is this:

src/plugin/cursor.js
thijstriemstra: +1 for removing the underscores. what about you @katspaugh?

thijstriemstra:
> Another thing. The Region class can be moved to a separate file? 

If we do that I would place the plugin files in it's own directory, e.g. `src/plugin/regions/`. This has implications for the build process but shouldn't be too hard to fix.

src/plugin/cursor.js
katspaugh: I think the _ in this case means that they are functions without `this`. Because they are passed directly as event listener callbacks:
`this.wrapper.addEventListener('mousemove', this._onMousemove);`
So I would keep them prefixed.

marizuccara: So, if the underscore in this case don't mean that formally these methods are considered private, they can be leave with this notation, because they can be inherited without problems, correct?

thijstriemstra: >  they can be leave with this notation, 

Agreed.

src/plugin/minimap.js
```diff
@@ -124,7 +124,7 @@ export default class MinimapPlugin {
             if (!document.body.contains(this.params.container)) {
                 ws.container.insertBefore(this.params.container, null);
             }
-
+            // how solve this dependency??
```
marizuccara, line 127: Here, checking if there is `regions` plugin with `this.wavesurfer.regions`, forces to call the plugin "regions". In this way, extending regions plugin, it cannot be called with a custom name. 
Any idea of how to solve this "name" dependency ?

src/plugin/minimap.js
```diff
@@ -124,7 +124,7 @@ export default class MinimapPlugin {
...
             if (this.wavesurfer.regions && this.params.showRegions) {
```
marizuccara, line 128: Here, checking if there is regions plugin with this.wavesurfer.regions, forces to call the plugin "regions". In this way, extending regions plugin, it cannot be called with a custom name.
Any idea of how to solve this "name" dependency ?

thijstriemstra: hi maria, maybe it's better to move the separate regions file to a new pull request and restrict this one to "Removed private methods of plugins and generalized plugins' access".

marizuccara: > Here, checking if there is regions plugin with `this.wavesurfer.regions`, forces to call the plugin "regions". In this way, extending `regions` plugin, it cannot be called with a custom name.
> Any idea of how to solve this "name" dependency ?

Hi, we don't decide how to fix this

PR #1929, opened 2020-04-29
Regions plugin: fix switch loop region

aimkiray:

If a region is in loop, try to loop another region at this time, the `out` event will still occur, causing the jump to fail, and will bind multiple events, browser may crash.

In order to solve this problem, I added a judgment in the `playLoop()` function, only looped when the play time was in the current region.

None

None

## Reproduction and steps:

https://wavesurfer-js.org/example/annotation/index.html

1. Hold down shift and click on a region to start loop.
2. Click any region before the region to reproduce the bug.

### DIFF
--- a/src/plugin/regions.js
+++ b/src/plugin/regions.js
@@ -138,8 +138,14 @@ class Region {
      * @param {number} start Optional offset to start playing at
      * */
     playLoop(start) {
-        this.play(start);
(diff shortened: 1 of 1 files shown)


marizuccara:
Hi! Tried your changes, and you fixed the problem! It is now possible to exit from the loop, great!

PR #1934, opened 2020-05-04
regions plugin: move classes to separate files

marizuccara:
`regions.js` file was splitted in two files, one containing the `RegionsPlugin` class, another the `Region` class.
This approach is useful because separating two conceptually different objects allows to treat them as different modules. In this way creating a Custom Plugin that extends the `RegionsPlugin` is simpler, and allows you to decide whether to leave `Region` class unchanged, or extends it too.

- Splitted `regions.js` in `index.js` and `region.js` and moved both files in `src/plugin/regions/` directory.
- Modified the plugins' build file, so to build the entry point are considered also nested directories

PR https://github.com/katspaugh/wavesurfer.js/pull/1928


src/plugin/regions/regions.js
```diff
@@ -0,0 +1,412 @@
+/**
```
thijstriemstra, line 1: can you add a since tag to this file (I would use version 3.4.0): https://jsdoc.app/tags-since.html

marizuccara: also in `region.js`?

thijstriemstra: oh sorry, I meant region.js. Perhaps in regions.js you can add a note that it was moved etc in 3.4.0. I'm also not sure if it should be called `src/regions/regions.js` or `src/regions/index.js`. I don't like the word `regions` used twice there.

marizuccara: We can call it or `index.js` or `regionsPlugin.js`, what you prefer?

thijstriemstra: not sure, what about you @katspaugh ?

thijstriemstra: there are also merge conflicts in `src/plugin/regions.js`. Maybe you should reset the master branch of your fork @marizuccara, there are a lot of unrelated changes in your pull requests.

katspaugh: `index.js` does sound better. You can import it with `import regions from 'plugin/regions` then.

thijstriemstra: agreed

marizuccara: But naming it index.js, when creating the bundle of regions plugin, it will be `wavesurfer.index.js` and `wavesurfer.region.js`, instead of  `wavesurfer.regions.js` and `wavesurfer.region.js`. Correct?

thijstriemstra: Yea, the way it's setup now, it will. So the build script needs to include a ``src/plugin/*/index.js` as well and use it's parent directory name for the plugin.

We could also refactor all plugins afterwards and move them into a directory and use the `index.js` convention but that's for a different pull request.

marizuccara: So, in this way, only `wavesurfer.regions.js` have to be created, and not also `wavesurfer.region.js`?

CHANGES.md
```diff
@@ -4,13 +4,21 @@ wavesurfer.js changelog
 3.4.0 (unreleased)
 ------------------
 
+- Splitted  `regions.js` file in  `region.js` containing the `Region` class, and `region.js` containing the `RegionsPlugin`
+  class, both moved into `src/plugin/regions` directory. In this way creating both classes can be extended to create
+  Custom Plugins.
+
+
+3.3.4 (unreleased)
```
thijstriemstra, line 12: No need for a 3.3.4 release first.

src/plugin/regions/regions.js
thijstriemstra: Yes.

CHANGES.md
marizuccara: Maybe 4.0.0 because there are many changes?

thijstriemstra: yeah, you can rename it.

src/plugin/regions/regions.js
```diff
@@ -0,0 +1,414 @@
...
+ *  @since 3.4.0 this class was splitted and moved to 'plugin/regions' directory
```
thijstriemstra, line 2: i would use `split` instead of `splitted`, and don't mention the directory move.

src/plugin/minimap.js
```diff
@@ -124,7 +124,7 @@ export default class MinimapPlugin {
             if (!document.body.contains(this.params.container)) {
                 ws.container.insertBefore(this.params.container, null);
             }
-            // how solve this dependency??
+
             if (this.wavesurfer.regions && this.params.showRegions) {
```
marizuccara, line 128: Do you remember my question here? If create a custom region plugin and call it in different way (e.g `customRegions` instead of `regions`), this statement will be always false, so the regions in the minimap will not be showed

thijstriemstra: ah yes. something like a new plugin parameter, e.g. `regionsPlugin`, where `this.wavesurfer.regions` is the default, but you should also be able to pass in your custom regions plugin instance (created earlier I guess)..?

marizuccara: Yes, maybe something like this:
- Add new `minimap` parameter, e.g `regionsPluginName`
- If is not passed from outside, its default is `'regions'`
- In the statement, check for `regions` plugin in this way:
  ```
  if (this.wavesurfer[this.params.regionsPluginName] && this.params.showRegions) {
       this.regions();
  }
What do you think?

thijstriemstra: Yes, sounds good. But I would add a new public var for `this.wavesurfer[this.params.regionsPluginName]` so it can be easily accessed for other usage.

marizuccara: You mean in the `minimap`, correct?

marizuccara: This changes have to be included in this pull request?

thijstriemstra: yea in minimap. The ability to use a customized regions plugin in the minimap plugin should/can be done in another pull request.

PR #1942, opened 2020-05-10
Make waveforms exclude x-axis (restore #478)

dpwe:

Waveform outlines no longer include 0 sample (x-axis). This is particularly important for highly zoomed-in views. Before this change, a highly-zoomed-in waveform might look like:

(where the amplitude=0 point is included at every time). With this change, it looks more like:

(where the actual waveform is drawn).

Thanks to alimatinfar1376@gmail.com for drawing my attention to this problem.

None.  Zoomed-in or offset waveforms appear differently, as in image above.

None.

### Todos/Notes:
None.

I added the ability to follow the detail in zoomed-in waveforms in katspaugh#478. This was regressed by @nathansnider soon afterwards in katspaugh@46e6edd#diff-268e184f68ebebc7a4772c6e3a47c2ab , but I'm not clear if the change was actually breaking something. I believe a display that doesn't arbitrarily render the waveform as connected to zero is preferable, but I don't understand why nathansnider made the change (and he's no longer active on GitHub).

refs #478


katspaugh:
Nice! Thanks a lot!

Does it cause any change in how silence looks like?

dpwe:
No.  If the waveform is already all zero, then the max and min peaks are also all zero, both before and after this change.

I modified the "zoom" example in the dist to allow the pixelsPerSecond to go up to 1000, then zoomed in to get comparison screenshots on a familiar example.  Before:

image

After:

image

This segment begins with near-silence, so you can see that it is little changed.

katspaugh:
Perfect, thanks! 👍

dpwe:
I added an inline comment to describe the intention of the initialization. 
I added a description to CHANGES.md.
I included the modification to the max zoom in examples/zoom to showcase the effect.

thijstriemstra:
thanks. the comment indentation looks a little off though.

dpwe:
Sorry!  Apparently untabify was not running in my editor.

PR #1943, opened 2020-05-11
minimap plugin: added ability to use a customized regions plugin

marizuccara:
To support customized regions plugin, it was added new parameter into `minimap` plugin, `regionsPluginName`, which could be setted to your custom region plugin name. 

New parameter `regionsPluginName` for `minimap` plugin.
- Its default value is 'regions'
- If you setted `showRegions` to `true` and you're using a customized regions plugin calling it with a custom name, set its name to `regionsPluginName` parameter

- Added  `regionsPluginName` in `minimap` parameters
- Added new variable `this.regionsPlugin`, so it can be accessed from outside

PR #1934 #1928


marizuccara:
Hi, I think that the `plugin inheritance example` should be updated, because we removed all private methods, so we can override them, instead of copying. What do you think? If yes open another pull request or reopen the old one?

thijstriemstra:
> What do you think? If yes open another pull request or reopen the old one?

Sounds good; in a new one.

PR #1949, opened 2020-05-15
regions plugin: ability to specify time format using timeformatCallback

### LINKED ISSUE #1948: Custom time format in regions tootip
It would be nice to have a way to provide custom function that will generate text displayed on region hower. 

Currently there is no way to do so and region tooltip always show m:ss format

jscrusader:
Adds timeformatCallback functionality to Regions

fixes #1948


thijstriemstra:
thanks @jscrusader! can you also add a note to the changelog?

jscrusader:
I have no idea why AppVeyor build failed :(

PR #1953, opened 2020-05-24
minimap plugin: added check for drawer presence

marizuccara:
Since `minimap` and `regions` plugin are related when the `showRegions` option is setted to `true`, there may be errors when one plugin is destroyed before the other. In particular, when the `minimap` plugin cannot find the `drawer` of regions. So the error which occurs is this:

image

Check il the `drawer.wrapper` exists before rendering regions.


thijstriemstra:
tests are failing but seems unrelated to this PR.

PR #1956, opened 2020-06-07
spectrogram plugin: perform null check when removing scroll listener

zhoutwo:
It turns out, if the parent WaveSurfer instance is not ready by the time `destroy()` is called, this line will throw an error.

Reason being https://github.com/katspaugh/wavesurfer.js/blob/master/src/plugin/spectrogram.js#L306 shows `this.drawer` is only assigned when `this._onReady` is called.

None

None


jagritkhetarpal:
Great !!
When will the changes will reflect in the package? @thijstriemstra

PR #1961, opened 2020-06-08
ci: use latest node on appveyor

thijstriemstra:


thijstriemstra:
Chrome breaks on travis and appveyor, same issue is described here: https://stackoverflow.com/questions/61957878/unit-tests-not-running-after-chrome-83-update

No solution yet though.

Issue #1531, opened 2018-12-13
Resume context in Chrome >= 71


thijstriemstra:
The following warning appears in Chrome >= 71 on wavesurfer-js.org:

```
webaudio.js:76 The AudioContext was not allowed to start.
It must be resumed (or created) after a user gesture on the page.
```

https://goo.gl/7K7WLu

nsomaru:
Is there any solution or workaround to this? Works fine in Firefox but refuses to play in Chrome.

thijstriemstra:
Yeah, see the link; https://goo.gl/7K7WLu Basically create an AudioContext on a user gesture, like button press, and pass that wavesurfer using `audioContext` parameter.

sundayz:
Any idea why Chrome emits this warning when creating the web audio audiocontext?
![image](https://user-images.githubusercontent.com/11685806/88923070-0675bd80-d269-11ea-95a1-49944d794bba.png)

Wavesurfer isn't set to autoplay... there's no audio playing on my page at startup. It always waits for the user to click buttons. But now even if you try to call wavesurfer.play() after a button press it won't start anymore.

Is it enough to call .resume() on the audiocontext before wavesurfer.play()? Or do I have to re-create the AudioContext object?

To make matters worse, the warning doesn't show on localhost, but it shows on the live website.

marizuccara:
Hi, I do this and I have not any warning on console.

![image](https://user-images.githubusercontent.com/44341274/88925569-fa8ff880-d274-11ea-9f5a-f6721eb9cd8b.png)

thijstriemstra:
@sundayz did you check https://goo.gl/7K7WLu?

sundayz:
Thanks guys. Calling wavesurfer.backend.ac.resume() myself works. I'm not sure why it doesn't work without me doing it, since if I look at the code inside WebAudio.play(), WaveSurfer should be calling resume() for me.

marizuccara:
Maybe is better to do it from the outside, because when the Promise from the resuming of the AudioContext is returned, you could want do do more stuffs, instead of only call play.

sundayz:
https://github.com/katspaugh/wavesurfer.js/blob/master/src/webaudio.js#L684

It is resumed here but doesn’t seem to work 🤔 maybe because it’s inside a Promise...?

marizuccara:
It should play only when audioContext is resumed, that is, when the Promise returned from resume method  is solved

[...]

PR #1964, opened 2020-06-13
Mediaelement safari fixes

osheroff:
This fixes a condition where Safari was not playing with mediaelement/webaudio.  It seems to implement the webaudio spec slightly differently than chrome and requires that you call AudioContext#resume().  

I also rewrote the header comment into more natural english.

refs #1531

### DIFF
--- a/src/mediaelement-webaudio.js
+++ b/src/mediaelement-webaudio.js
@@ -59,6 +59,13 @@ export default class MediaElementWebAudio extends MediaElement {
         this.sourceMediaElement.connect(this.analyser);
     }
 
+    play(start, end) {
+        if (this.ac.state == 'suspended') {
+            this.ac.resume && this.ac.resume();
+        }
+        return super.play(start, end);
+    }
+
     /**
      * This is called when wavesurfer is destroyed
      *
--- a/src/mediaelement-webaudio.js
+++ b/src/mediaelement-webaudio.js
@@ -1,12 +1,11 @@
 import MediaElement from './mediaelement';
 
 /**
- * MediaElementWebAudio backend: allows to load audio as HTML5 audio tag and use it with WebAudio API.
- * Setting the MediaElementWebAudio backend, there is the possibility to load audio of big dimensions, using the WebAudio API features.
- * The audio to load is an HTML5 audio tag, so you have to use the same methods of MediaElement backend for loading and playback.
- * In this way, the audio resource is not loaded entirely from server, but in ranges, since you load an HTML5 audio tag.
- * In this way, filters and other functionalities can be performed like with WebAudio backend, but without decoding
- * internally audio data, that caused crashing of the browser. You have to give also peaks, so the audio data are not decoded.
+ * MediaElementWebAudio backend: load audio via an HTML5 audio tag, but playback with the WebAudio API.
+ * The advantage here is that the html5 <audio> tag can perform range requests on the server and not
+ * buffer the entire file in one request, and you still get the filtering and scripting functionality
+ * of the webaudio API.
+ * Note that in order to use range requests and prevent buffering, you must provide peak data.
  *
  * @since 3.2.0
  */
--- a/src/mediaelement-webaudio.js
+++ b/src/mediaelement-webaudio.js
@@ -60,9 +60,7 @@ export default class MediaElementWebAudio extends MediaElement {
     }
 
     play(start, end) {
-        if (this.ac.state == 'suspended') {
-            this.ac.resume && this.ac.resume();
-        }
+        this.resumeAudioContext();
         return super.play(start, end);
     }
 
--- a/src/webaudio.js
+++ b/src/webaudio.js
@@ -576,6 +576,17 @@ export default class WebAudio extends util.Observer {
     }
 
     /**
+     * @private
+     *
+     * Safari requires an explicit call to #resume before it will play back audio
+     */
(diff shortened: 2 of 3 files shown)


src/mediaelement-webaudio.js
```diff
@@ -59,6 +59,13 @@ export default class MediaElementWebAudio extends MediaElement {
         this.sourceMediaElement.connect(this.analyser);
     }
 
+    play(start, end) {
+        if (this.ac.state == 'suspended') {
+            this.ac.resume && this.ac.resume();
```
marizuccara, line 64: Hi, because `resume` method of AudioContext return a `Promise`, maybe is better to call `play` when Promise is solved? @thijstriemstra what do you thing?

thijstriemstra: Makes sense.

osheroff:
> Why rewrite the description?

I felt the language used could be clearer and more accurate.

src/mediaelement-webaudio.js
osheroff: Hi folks, 
I tried this approach but Safari is throwing me:

```
Unhandled Promise Rejection: NotAllowedError: The request is not allowed by the user agent or the platform in the current context, possibly because the user denied permission.
```

(this appears to get thrown from the actual call to `play`).  I think it's because when we try to call `super.play()` from the `ac.resume` promise, we leave the context of the `click` handler and Safari thinks we're trying to auto-play audio.

So I think the code is probably correct as-is.

thijstriemstra:
@katspaugh can you add @osheroff to the team? I think he made some valuable contributions: https://github.com/katspaugh/wavesurfer.js/commits?author=osheroff

src/mediaelement-webaudio.js
thijstriemstra: > we leave the context of the click handler and Safari thinks we're trying to auto-play audio.

Did you try `resume`-ing in the error handler of the promise? I guess the audiocontext/state is destroyed/polluted by then and `resume` doesn't make a difference anymore.

thijstriemstra: also, this check/statement is also present in webaudio.js, perhaps we can move this check to some utility function and re-use it?

osheroff: > Did you try resume-ing in the error handler of the promise? I expect the audiocontext/state is destroyed/polluted by then and resume doesn't make a difference anymore. 

I was trying like:

```
  ac.resume().then(() => super.play(...)
```

which seems like the most natural code to write.  The call to `resume` succeeds, but the call here: 

https://github.com/katspaugh/wavesurfer.js/blob/b7d1a25965869acb1843323232831ad48f5780af/src/mediaelement.js#L291

throws `NotAllowedError`.  If you have another approach you'd like me to try let me know.

katspaugh:
> @katspaugh can you add @osheroff to the team? I think he made some valuable contributions: https://github.com/katspaugh/wavesurfer.js/commits?author=osheroff

Sure, good idea. Ben, I've sent you an invite. Many thanks for your work!

src/webaudio.js
```diff
@@ -575,6 +575,17 @@ export default class WebAudio extends util.Observer {
         this.source.connect(this.analyser);
     }
 
+    /**
+     * @private
+     *
+     * Safari requires an explicit call to #resume before it will play back audio
```
thijstriemstra, line 581: I would use 'some browsers' instead of Safari only.

osheroff: fixed.

osheroff:
off-topic, I was trying to figure out why all the builds were failing.  I got down the rabbit hole so far as "something, somewhere, for some reason is adding a `--no-eslintrc` flag to the `eslint` command line", but I got quite stuck after that....

thijstriemstra:
> "something, somewhere, for some reason is adding a --no-eslintrc flag to the eslint command line"

I added this yesterday, but local build was fine with it. See `lint` command in package.json. Feel free to remove/fix of course.

osheroff:
> I added this yesterday, but local build was fine with it. See lint command in package.json. Feel free to remove/fix of course.

ah, mystery solved, this is one of those odd travis "merged with master" builds.  got it.

What's the merge policy here?  2 plus-ones?

thijstriemstra:
> What's the merge policy here? 2 plus-ones?

What do you mean?

osheroff:
> What do you mean?

like, how many approvals do you want before a PR is merged?

thijstriemstra:
> like, how many approvals do you want before a PR is merged?

There is not really such a rule in this project. If you see a PR that is 99.99% correct to you, go ahead and merge it. If you have doubts, pick a couple of people and request a review. Hopefully someone responds soon and you can then merge it. If you open a PR that is trivial (doc/formatting changes), you can even merge it yourself afterwards (or commit direct to master but only if it's, like i said, trivial). Now that we add more and more people to the team it might make sense to have some formal policy at some point but for now it all seems to work out like this.

thijstriemstra:
> Why rewrite the description?
> I felt the language used could be clearer and more accurate.

I think it has improved @marizuccara.

bwhitty:
@thijstriemstra is it possible that this gets into a 3.x release? Or is the version 3 release line finished in favor of everything going towards the version 4 release?

thijstriemstra:
I think 4.0.0 can be released next week.

Issue #1931, opened 2020-05-01
On Iphone Volume control mute is not working


namitgarg:
Hey when playing on iphone safari browser volume control , mute are not working , i have tested the below page https://wavesurfer-js.org/example/mute/index.html

sundayz:
You need to use the `WebAudio` or `MediaElementWebAudio` backends on iOS safari.


iOS Safari doesn't allow setting volume or playback rate with the default `MediaElement` backend

Issue #1703, opened 2019-07-26
toggleMute() is not return audio.


iagosilvamelo:
Good afternoon, how are you?
i'm development a audio player but the toggleMute() is muted audio and is not return audio.

This is my code:

```
$("#playerMute").click( e => {
    e.target.classList.toggle('active');
    wavesurfer.toggleMute();
});
```
Thanks for your attention.

thijstriemstra:
@iagosilvamelo what is the wavesurfer.js configuration you are using?

> and is not return audio.

what do you mean?

iagosilvamelo:
Helo @thijstriemstra.
This is my wavesurfer instance:
```
const wavesurfer = WaveSurfer.create({
    container: document.querySelector('#waveform'),
    waveColor: '#0E8F84',
    progressColor: '#23514C',
    backend: 'MediaElement',
    hideScrollbar: true,
    normalize: true,
    plugins: [
        WaveSurfer.cursor.create({
            showTime: true,
            opacity: 0.6,
            customShowTimeStyle: {
                'background-color': '#000',
                color: '#fff',
                padding: '2px',
                'font-size': '12px'
            }
        })
     ]
});
```

When activating the function toggleMute() to click button, the audio is muted, this is ok. With audio muted when activating the function toggleMute() to click button, mute is not disabled.

I tried with setMute(true) and false and setMute(1) and 0. It didn't work either.

Sorry my english, hehehe.

thijstriemstra:
what version of wavesurfer.js?

iagosilvamelo:
3.0.0

aburai:
i can reproduce this in v3. when i use `toggleMute` the volume was not restored and stays on 0
i will investigate and create a PR when i found a solution

iagosilvamelo:
I ended up doing like this:

```
setVol(e) {
        this.volume = e.target.value;
        this.player.setVolume(e.target.value);
},
mute() {
        this.player.setVolume(0);
        $("#playerVolume").val(0)
},
desmute() {
        this.player.setVolume(this.volume);
        $("#playerVolume").val(this.volume)
},
```
I'm not using mute.

aburai:
i think it depends on this change for backend MediaElement
```
        // only needed for MediaElement backend
        if (this.params.backend === 'MediaElement') {
            this.backend.on('seek', () => {
                this.drawer.progress(this.backend.getPlayedPercents());
            });

            this.backend.on('volume', () => {
[...]

PR #1966, opened 2020-06-15
Fix muting behavior with the MediaElement backend

ThibG:

`wavesurfer.js` currently handles muting the media by setting the backend's volume to 0 and remembering the previous value, **but** it also checks whether the backend maintains muting state and uses it to override its own. This is currently used by the `MediaElement` to make wavesurfer aware of muting/volume changes performed directly on the `Audio` element, but muting through `wavesurfer.js` itself triggers a backend volume change without changing the backend's `isMuted` state, so the end result is the volume switching to 0 and the media being unmuted, leading to inconsistent and erratic behavior.

This PR adds an optional `setMute` method to backends so that the `MediaElement` backend can handle muting by itself.

None.

None (the `setMute` method is optional so that other backends do not need change—but if we are ok with breaking the internal API by requiring the `setMute` method, we can move the `wavesurfer.js` code to the WebAudio backend).

I believe the “immediately unmuting” behavior was introduced by #1635. This PR fixes it while still handling external changes to the `Audio` element state.

refs #1931
refs #1703

### DIFF
--- a/src/mediaelement.js
+++ b/src/mediaelement.js
@@ -398,6 +398,7 @@ export default class MediaElement extends WebAudio {
     /**
      * Enable or disable muted audio
      *
+     * @since 4.0.0
      * @param {boolean} mute Specify `true` to mute audio.
      */
     setMute(muted) {
--- a/src/mediaelement.js
+++ b/src/mediaelement.js
@@ -396,6 +396,17 @@ export default class MediaElement extends WebAudio {
     }
 
     /**
+     * Enable or disable muted audio
+     *
+     * @param {boolean} mute Specify `true` to mute audio.
+     */
+    setMute(muted) {
+      // This causes a volume change to be emitted too through the
+      // volumechange event listener.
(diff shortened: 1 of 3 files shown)


src/mediaelement.js
```diff
@@ -395,6 +395,17 @@ export default class MediaElement extends WebAudio {
         }
     }
 
+    /**
+     * Enable or disable muted audio
+     *
+     * @param {boolean} mute Specify `true` to mute audio.
+     */
+    setMute(muted) {
```
thijstriemstra, line 403: can you add `@since 4.0.0` here?

ThibG:
Done!

osheroff:
hey folks, this PR broke a few tests (that were previously hidden behind some eslint issues).  I'm not sure if the tests are incorrect or if the code is, but maybe someone wants to look?

thijstriemstra:
@osheroff fixed with #1968 that was merged just now.

PR #1968, opened 2020-06-15
Fix setMute setting incorrect isMuted

ThibG:
#1966 incorrectly set wavesurfer object's `isMuted` to `true` even when it's supposed to be `false`.

This is not an issue in practice, as it gets correctly re-set by the `volumechange` handler, but this is wrong anyway, and as a result make the tests fail.


thijstriemstra:
thanks. did you also run the tests locally and see if they pass @ThibG?

Issue #1918, opened 2020-04-16
Long track in mobile


Garnetoff:
Hi.
If track over 1h, then site crashes in mobile phone. On desktop everything is fine.

thijstriemstra:
mobile device does not have enough resources compared to your desktop.

joeytwiddle:
If you use a peaks file, then your mobile can use that to display the waveform, and it won't need to read the entire audio file into memory at one go.

Issue #1866, opened 2020-02-18
Needs two clicks to play in Safari 13.0.5 with preloaded peaks.


octave:
### Code needed to reproduce the issue:

```
var wavesurfer = WaveSurfer.create({
    container: waveformElement,
    barWidth: 1,
    barGap: 2
  });

wavesurfer.load(url, peaks.data);

playPauseButton.click(function(e) {
   wavesurfer.playPause();
});
```

### Use behaviour needed to reproduce the issue:

In Safari, when clicking the play/pause button the first time, it loads the file but doesn't start to play. It works only when pressing a second time.

I tried to use the `backend: 'MediaElement` instead for Safari, but even with `preload='none'` file is being downloaded on page load. I need to have the file loaded on click and not earlier as I have a lot of instances in the page.

Seems to be an AudioContext issue with Safari, but the `wavesurfer.playPause();` on click should take care of this.

It works fine in Chrome and Firefox.

octave:
@thijstriemstra thanks for updating the issue. Is there any workaround in the meantime?

thijstriemstra:
create the `audioContext` when the user presses a btn, and pass that to wavesurfer.js at startup using the audioContext option.

octave:
@thijstriemstra my instances of wavesurfer.js are rendered at page load to display waveforms. So I can't pass an audioContext if the instance is already created correct?

What I'm wondering is when clicking the play/pause button the first time that calls `wavesurfer.playPause();` wavesurfer.js is creating the audioContext and so should make it work.

I did a quick demo that create an audioContext and play a file on click and it works fine in Safari: https://codepen.io/octavez/pen/RwPKyqM

So that looks like it's an issue with wavesurfer.js more than Safari.

matt3224:
I also ran into this today in Safari 13.1 and wavesurfer 3.3.1

Issue #740, opened 2016-06-06
iOS w/ MediaElement (double click and slower)


eduzaka:
Hi, first of all congratulations, this is really good.

We are looking info but we didn't find what's the way to make it working on iOS with MediaElement. We think the problem is the autoplay not accepted on iOS, but other audio players work like a charm (without waveforms) in iOS without autoplay, can we make some code to solve it?

When we click on track, it loads very slowly (not like desktop) and later we have to click again to make it sound.

katspaugh:
What are this other players? We could look at their source code and see how they work around the iOS limitations.

Regarding slowness, it's because it downloads the file via AJAX, decodes the binary into a JavaScript array and then draws a waveform from this array. I wouldn't be surprised it to be slower on mobile devices. You may want to use a pre-decoded waveform array. See https://github.com/katspaugh/wavesurfer.js/blob/master/example/audio-element/main.js#L26-L28

eduzaka:
Hey, thanks for the quick reply.

We are using prerendered peaks with a JSON data, in the desktop it loads perfectly but mobile loads it slowly. BTW, the first click loads peaks and the second click plays audio.

We bought some other players time ago, for example: http://codecanyon.net/item/html5-audio-player-with-playlist/1694831

We need something like this, but with waveform (like wavesurfer)... we are looking to mix both, but maybe there are other solution.

entonbiba:
@eduzaka please test using latest version, performance has improved significantly since this ticket.

reopen is issue still persists.

PR #1969, opened 2020-06-16
avoid call to HTMLMediaElement#load when peaks are given

osheroff:
calling `.load()` explicitly on a mediaelement entirely defeats the "preload" setting and makes it impossible to avoid downloading data before playback, which is often what the user is trying to do by providing peak data. 

When HTMLMediaElement#load is called, all of the major browsers make some request for the file. 
Chrome and FF make a range request for the entire file, but only report downloading them partially.  Safari reports downloading the file in its entirety.


thijstriemstra:
thanks @osheroff, could you check if there are any tickets that get fixed by this change (and add them to the PR description)?

osheroff:
I would have guessed there would have been more.

- maybe affects: https://github.com/katspaugh/wavesurfer.js/issues/740
- will help: https://github.com/katspaugh/wavesurfer.js/issues/1918
- partially: https://github.com/katspaugh/wavesurfer.js/issues/1866 (here there are two issues, the first is a problem with the webaudio backend but he reports being unable to use the mediaelement backend due to the issue I've been seeing)

There's also a whole host of people talking about memory issues etc, but they're likely using the webaudio backend.

thijstriemstra:
can you add a changelog entry?

thijstriemstra:
ps. it seems only Chrome dies during the CI builds, Firefox seems fine. We can disable Chrome on CI for a while and try again when it's back up and stable. Not being able to trust the CI results is annoying.

marizuccara:
Maybe it can be done not when `!peaks` but when `preload `attribute is different from `none` ?

marizuccara:
Hi maybe the check can be done not if `!peaks` but for example if `preload` attribute is of type `none`?

osheroff:
> Hi maybe the check can be done not if !peaks but for example if preload attribute is of type none?

@marizuccara I think it's a good point, but maybe it should be both `!peaks && (preload == 'none')`?  If we're not given server-generated peaks, I think we always have to call `.load` in order to trigger the waveform drawing (at least that's what the comment in the file seemed to indicated).

osheroff:
any further comments here?  @marizuccara I modified it to check if `preload == 'none'`

CHANGES.md
```diff
@@ -4,6 +4,8 @@ wavesurfer.js changelog
 4.0.0 (21.06.2020)
 ------------------
 
+- Don't call HTMLMediaElement#load when given peaks and preload == 'none'.
+  Prevents browsers from pre-fetching audio (#1969)
```
thijstriemstra, line 8: 4.0.0 is already released, you can add a new entry, e.g. 4.1.0.

osheroff: k fixed.

thijstriemstra:
branch has merge conflict.

osheroff:
Ok I fixed the merge conflict.  The system of putting changelog entries into PRs is a bit problematic, as every time another PR is merged into master my branch goes back into conflict (this is like the 3rd time I've had to update CHANGES.md on this PR).  I'm wondering if there's a better solution...

thijstriemstra:
>  The system of putting changelog entries into PRs is a bit problematic, as every time another PR is merged into master my branch goes back into conflict

We can use a section like this at top to reduce merge conflicts:

```
Next (unreleased)
------------------
```

src/mediaelement.js
```diff
@@ -160,7 +160,7 @@ export default class MediaElement extends WebAudio {
         elt.controls = this.params.mediaControls;
         elt.autoplay = this.params.autoplay || false;
 
-        this._load(elt, peaks);
+        this._load(elt, peaks, elt.preload);
```
thijstriemstra, line 163: What if `elt.preload` is undefined?

osheroff: good question. 

It won't be undefined.  If you don't set the HTML tag, in chrome/ff, the `preload` element will default to `""`.  In Safari, it defaults to `"auto"`.  The behavior is also similar: chrome/ff don't preload at all, and Safari first does a range request and then ends up preloading a whole 2mb file. 

I can detect these cases here and change our `.load` behavior accordingly, but I'm really not sure what the user's intention really is.  My guess is that if they're providing peak data they almost *never* want a preload, but that's just a guess...

PR #1970, opened 2020-06-16
guard against null _disabledEventEmissions in Observer#fireEvent

### LINKED ISSUE #1975: Cannot read property 'includes' of undefined
Uncaught TypeError: Cannot read property 'includes' of undefined
    at e.fireEvent (observer.js:125)
    at HTMLUnknownElement.<anonymous> (region.js:349)

### Wavesurfer.js version(s):
https://unpkg.com/wavesurfer.js
https://unpkg.com/wavesurfer.js/dist/plugin/wavesurfer.regions.min.js

Chrome 83, Windows 10

```
<script src="https://unpkg.com/wavesurfer.js/dist/plugin/wavesurfer.regions.min.js"></script>

// example region
wavesurfer.addRegion({
		id: "pad-1",
		start: 1,
		end: 10,
		color: "blue",
});
```
Error is produced on interactions with the wavesurfer regions UI
ex. hovering over a region

*edit - the initialized regions appear to load correctly.. but attempting to clearRegions() and load new produces the error above, and the new regions fail.

osheroff:
due to the odd way the RegionsPlugin pulls in the methods in util.Observer, it needs to setup a private array or it crashes in `fireEvent`.

I could also fix this in `Observer` by checking for the existence of the array.  I dunno.  both seem odd.

fixes #1975

### DIFF
--- a/src/util/observer.js
+++ b/src/util/observer.js
@@ -116,13 +116,22 @@ export default class Observer {
     }
 
     /**
+     * plugins borrow part of this class without calling the constructor,
+     * so we have to be careful about _disabledEventEmissions
+     */
+
+    _isDisabledEventEmission(event) {
+        return this._disabledEventEmissions && this._disabledEventEmissions.includes(event);
+    }
+
+    /**
      * Manually fire an event
      *
      * @param {string} event The event to fire manually
      * @param {...any} args The arguments with which to call the listeners
      */
     fireEvent(event, ...args) {
-        if (!this.handlers || this._disabledEventEmissions.includes(event)) {
+        if (!this.handlers || this._isDisabledEventEmission(event)) {
             return;
         }
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,8 @@ wavesurfer.js changelog
 4.0.0 (unreleased)
 ------------------
 
+- Fixes for event handling with certain plugins (regions, microphone).
+  The crash would have involved '_disabledEventEmissions'
 - Fixed mediaelement-webaudio playback under Safari (#1964)
 - Fixed the `destroy` method of the `MediaElementWebAudio` backend. Instead of
   destroying only the media element, the audio nodes are disconnected and the


osheroff:
hmmm... this doesn't seem to be fixing the crash

osheroff:
nevermind.  I'm doing something dumb.

osheroff:
@katspaugh I was doing something dumb in my own code *and* in this fix.  But it does need to be fixed, regions are totally broken in master.

src/plugin/regions/index.js
```diff
@@ -123,6 +123,9 @@ export default class RegionsPlugin {
         observerPrototypeKeys.forEach(key => {
             Region.prototype[key] = this.util.Observer.prototype[key];
         });
+
+        Region.prototype["_disabledEventEmissions"] = [];
```
katspaugh, line 127: Thanks for the quick fix! Could you change the quotes to single?

cadavre:
The problem is with https://github.com/katspaugh/wavesurfer.js/blob/master/src/util/observer.js#L124

Inside `fireEvent` – `this` is being an argument passed when triggering the event, not Observer itself.

It looks like Observer is somehow broken, not Regions plugin.

src/plugin/regions/index.js
cadavre: Why set `disabledEventEmissions` on region? It should be set on wavesurfer itself only. This is not a fix.

osheroff: no, this is the fix.  I had previously opened a PR where it wasn't the fix.

osheroff:
@cadavre test it yourself before and after this patch.  this does indeed fix it...

osheroff:
@katspaugh updated with single quotes

src/plugin/regions/index.js
osheroff: @cadavre the code is confusing, but my fix is correct.  On line 123 we grab all the methods from `utils.Observer` and stuff them into `Region`.  but `Observer` now requires that `this` contains an array called `_disabledEventEmissions`.   If you read further you'll see that many objects in the wavesurfer codebase inherit from `Observer`, not just wavesurfer.

cadavre: @osheroff I see now – Plugins are Observers as well (`Plugin extends Observer`).

The `_disabledEventEmissions` was missing, because it's initialized in Observer's constructor. Shouldn't we call super() somewhere while initializing plugins @katspaugh ?

cadavre: Thanks for clarification @osheroff !

osheroff: > shouldn't we call super() somewhere while initializing plugins 

The Region plugin doesn't literally inherit from `Observer`, it kinda just borrows all the code.  I'm not sure why this was done, I think probably because they're built into different source files?  I agree it's kind of goofy what the Region plugin is doing here.

thijstriemstra:
All of the plugins that use observer a broken currently, e.g. microphone plugin. I don't like adding this change to every plugin (unless there's really no other way).

osheroff:
well @thijstriemstra here's the other approach...

src/plugin/regions/index.js
```diff
@@ -123,6 +123,7 @@ export default class RegionsPlugin {
         observerPrototypeKeys.forEach(key => {
             Region.prototype[key] = this.util.Observer.prototype[key];
         });
+
```
thijstriemstra, line 126: can you revert this whitespace change?

src/util/observer.js
```diff
@@ -115,14 +115,23 @@ export default class Observer {
         this._disabledEventEmissions = eventNames;
     }
 
+    /**
+     * looks overly paranoid but plugins borrow part of this class
+     * without calling the constructor.
+     */
```
thijstriemstra, line 121: can you rephrase this? :)

osheroff:
@thijstriemstra updated

PR #1976, opened 2020-06-22
Include properties in Regions instance

iansharkey:
Include the _disabledEventEmissions and handlers properties in the Regions class

Notes:
Destroying a wavesurfer instance in version 4.0.0 with the regions plugin active would result in an undefined access to the _disabledEventEmissions property. This is just a small patch to include those properties, but I'm not sure this is the correct fix, feel free to disregard!

Thanks for the great project!


thijstriemstra:
> This is just a small patch to include those properties, but I'm not sure this is the correct fix, feel free to disregard!

Thanks for helping out! I'll close this PR in favor of #1970

PR #1977, opened 2020-06-23
regions plugin: improve loop playback logic

### LINKED ISSUE #1868: Cancel Region Play Loop Properly
Hi, thanks for making this fantastic repo!

I'm trying to build a system with which users can freely navigate the video and loop whatever intervals they want. The region plugin seems to work properly at first glance. However,  it seems there are some bugs with the region plugin, specifically associated with looping.

[Problem]
1. Once I start looping on a region, I would not be able to use the native cursor from the wavesufer to navigate the video (you might still be able to select timestamps on the waveform but when you start playing it will jump back to the looping region). The only way to jump to other parts of the video is to click on other regions. 

2. If I loop on multiple regions, the browser would start being overloaded and becomes extremely slow. This happens when I've done looping a region and try to loop on other regions without stop the video first.

[Reproduce]
You could reproduce all these problems with the annotation and video-annotation examples.

I've tried to solve it by myself but haven't succeeded yet. I doubt it's because there's not any function actually "cancel" the looping. Does anyone have an idea?

Thanks

cadavre:

`regions` plugin has doubled loop playback logic.

**First** can be found here, binding `out` event:
```js
    /* Bind audio events. */
    bindInOut() {
        // ...

        /* Loop playback. */
        this.on('out', () => {
            if (this.loop) {
                this.wavesurfer.play(this.start);
            }
        });
    }
```
https://github.com/katspaugh/wavesurfer.js/blob/master/src/plugin/regions/region.js#L336

**Second** is in `playLoop()` method:
```js
    /**
     * Play the audio region in a loop.
     * @param {number} start Optional offset to start playing at
     * */
    playLoop(start) {
        const s = start || this.start;
        this.wavesurfer.play(s);
        this.once('out', () => {
            const realTime = this.wavesurfer.getCurrentTime();
            if (realTime >= this.start && realTime <= this.end) {
                return this.playLoop();
            }
        });
    }
```
https://github.com/katspaugh/wavesurfer.js/blob/master/src/plugin/regions/region.js#L151

**The problem** is that both "loops" cannot be canceled in any way. So instead of doubling the same logics about looping and to allow disabling the loop – I propose to remove logics from `playLoop()` but keep method itself it for BC, **but** depend on `this.loop: boolean` flag to handle looping via `bindInOut()`. This way we can add `setLoop(boolean)` method to control the looping and remove strange logic from `playLoop()` if.

No breaking changes.

No breaking changes.

### Todos/Notes:

- [ ] Update documentation.

fixes #1868

EDIT: Removed TODO from example.

### DIFF
--- a/src/plugin/regions/region.js
+++ b/src/plugin/regions/region.js
@@ -149,16 +149,18 @@ export class Region {
      * @param {number} start Optional offset to start playing at
      * */
     playLoop(start) {
-        const s = start || this.start;
-        this.wavesurfer.play(s);
-        this.once('out', () => {
-            const realTime = this.wavesurfer.getCurrentTime();
-            if (realTime >= this.start && realTime <= this.end) {
-                return this.playLoop();
-            }
-        });
+        this.loop = true;
+        this.play(start);
     }
 
+    /**
+     * Set looping on/off.
+     * @param {boolean} loop True if should play in loop
+     */
+    setLoop(loop) {
+        this.loop = loop;
+    }
+    
     /* Render a region as a DOM element. */
     render() {
         const regionEl = document.createElement('region');
--- a/src/plugin/regions/region.js
+++ b/src/plugin/regions/region.js
@@ -70,7 +70,7 @@ export class Region {
                 this.marginTop = this.wavesurfer.getHeight() * channelIdx + 'px';
             }
         }
-        
+
         this.formatTimeCallback = params.formatTimeCallback;
 
         this.bindInOut();
@@ -160,7 +160,7 @@ export class Region {
     setLoop(loop) {
         this.loop = loop;
     }
-    
+
     /* Render a region as a DOM element. */
     render() {
         const regionEl = document.createElement('region');
@@ -338,7 +338,10 @@ export class Region {
         /* Loop playback. */
         this.on('out', () => {
             if (this.loop) {
-                this.wavesurfer.play(this.start);
+                const realTime = this.wavesurfer.getCurrentTime();
+                if (realTime >= this.start && realTime <= this.end) {
+                    this.wavesurfer.play(this.start);
+                }
             }
         });
     }
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,11 +1,16 @@
 wavesurfer.js changelog
 =======================
 
-4.0.0 (21.06.2020)
+4.0.1 (In progress)
 ------------------
 
 - Fixes for event handling with certain plugins (regions, microphone).
   The crash would have involved '_disabledEventEmissions'
+- Improved and unified loop playback logic in regions plugin. (#1868)
+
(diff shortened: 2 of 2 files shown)


src/plugin/regions/region.js
```diff
@@ -149,16 +149,18 @@ export class Region {
...
-        const s = start || this.start;
-        this.wavesurfer.play(s);
-        this.once('out', () => {
-            const realTime = this.wavesurfer.getCurrentTime();
-            if (realTime >= this.start && realTime <= this.end) {
-                return this.playLoop();
-            }
-        });
+        this.loop = true;
+        this.play(start);
```
osheroff, line 153: need to keep the `start || this.start` logic that was in the previous version of the function.

cadavre: `this.play()` is already having it.

osheroff:
Overall the logic here looks correct to me.  I went to go test this code, as it turns out, region playback is totally broken in master, needs https://github.com/katspaugh/wavesurfer.js/pull/1970 to fix it. 

I tested with #1970 and this pr together and the looping code appears to work just fine.

src/plugin/regions/region.js
osheroff: got it, sorry.

marizuccara:
Hi, this new logic seems to break again the loop: 
- if a region is looping and you click outside, the cursor seems to go to the clicked position, but after that returns looping on the region. 
- Again, if click inside other region, it goes in loop from one region to another.
I think is that because with the previous controls, the loop was interrupted if the current time was not inside the region.

cadavre:
@marizuccara please provide scenarios of what should happen to your cases. I'd like to improve looping so all cases could be covered.

cadavre:
@thijstriemstra changelog updated. It looks like also previous merged PR had wrong entry – made for 4.0.0 instead of 4.0.1 – so I also fixed this.

@marizuccara I reused code from before committing, to update loop logic based on current playback time vs region start-end times. Could you please test it in your case?

thijstriemstra:
This branch has merge conflict.

> changelog updated. It looks like also previous merged PR had wrong entry – made for 4.0.0 instead of 4.0.1 – so I also fixed this.

4.0.1 was released today: https://github.com/katspaugh/wavesurfer.js/commit/aaf433ecaf2f032a28a0419401ed56958f67a17e

marizuccara:
> @thijstriemstra changelog updated. It looks like also previous merged PR had wrong entry – made for 4.0.0 instead of 4.0.1 – so I also fixed this.
> 
> @marizuccara I reused code from before committing, to update loop logic based on current playback time vs region start-end times. Could you please test it in your case?

Hi @cadavre , tested your code! I think that with this unified logic, and with your latest fix, the solution seems very nice!

cadavre:
@thijstriemstra conflicts resolved. Merged.

PR #1978, opened 2020-06-23
Improve region example

osheroff:
this PR includes #1970, I will merge them separately tho.  Adds some nice buttons to the regions plugin example and gives us a way to test looping / region playback.


osheroff:
updated

PR #1983, opened 2020-06-24
performance: use clip-path css instead of width

osheroff:
constantly changing the width of the multi-canvas `progressWave` element triggers a layout pass in the browser.  This doesn't seem to be a huge problem for Chrome, but in Safari on mac it's worth quite a hunk of performance.

This is Safari pre-patch  (I played the whole file on /example/html-init/ for these tests):

Screen Shot 2020-06-23 at 9 17 18 PM

Safari post-patch:

Screen Shot 2020-06-23 at 9 15 15 PM

Chrome seems to jitter more and benefit less, but note the ~10% of rendering that we seem to save:

pre-patch:
Screen Shot 2020-06-23 at 9 28 19 PM
Screen Shot 2020-06-23 at 9 29 08 PM

post-patch:
Screen Shot 2020-06-23 at 9 30 35 PM
Screen Shot 2020-06-23 at 9 31 30 PM

would love to know if I'm missing anything with the logic here or if other people can run benchmarks and see if this feels like a win.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,7 @@ Next (unreleased)
 ------------------
 
 - Use `clip-path` instead of `width` changes in the multicanvcas drawer
-  to avoid layout thrashing.
+  to avoid layout thrashing. (#1983)
 - Improved and unified loop playback logic in regions plugin. (#1868)
 
 4.0.1 (23.06.2020)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,12 @@
 wavesurfer.js changelog
 =======================
 
+Next (unreleased)
+------------------
+
+- Use `clip-path` instead of `width` changes in the multicanvcas drawer
+  to avoid layout thrashing.
+
 4.0.1 (23.06.2020)
 ------------------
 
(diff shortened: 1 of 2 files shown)


marizuccara:
Tried in Chrome/Windows and it works correctly. I have not Safari, so I cannot view the performance there. But this improvement seems to have sense, so I approve

src/drawer.multicanvas.js
```diff
@@ -555,12 +558,18 @@ export default class MultiCanvas extends Drawer {
         }
     }
 
+    makeInset(rightInset) {
```
thijstriemstra, line 561: can you add a docstring that describes this performance enhancement?

osheroff:
ok, all fixed.  @katspaugh is this part of the officially required style for the project (template strings over concatenation, single quotes")?  could this be in es-lint rather than a round-trip on every PR?

katspaugh:
Thanks, @osheroff!
It would be great to add these rules to the eslint config, yes. There might be a few places where we had used concatenation though.

CHANGES.md
```diff
@@ -1,9 +1,11 @@
 wavesurfer.js changelog
 =======================
 
-4.0.2 (In progress)
--------------------
+Next (unreleased)
+------------------
 
+- Use `clip-path` instead of `width` changes in the multicanvcas drawer
+  to avoid layout thrashing.
```
thijstriemstra, line 8: add PR request number

osheroff: added

PR #1985, opened 2020-06-24
Regions: remove cursor col-resize when disable region resizing

### LINKED ISSUE #1963: regions plugin: couple of bugs report
I'm sorry I closed check list "to-do before open issue" once, and I can't reproduce it... 
`wavesurfer.js version is 3.3.3`

I faced with this strange behavior. 
1) If I init waveform and set regions with { resize: false, drag: false } on initialization, then I can't change those fields by region.update() in the future. For now I solve it by don't touching  {resize, drag} on init, and update them all just after initialization. 
2) if I do region.update( { resize: false } ) - cursor 'resize' still appears upon the region (what is not expected).  But if I do this on init, cursor 'resize' doesn't appear, what is expected (good). 

Proposal:
It will be great to have enable/disable for Drag and Resize - affecting all regions (existing and newly created). Like enable/disable DragSelection. 

I hope it will be helpful for you. If you want to see my app where I use Wavesurfer with Regions, please visit [frazy.me](frazy.me) - platform for creating and hosting language learning materials.

Thank you for your great library.

marizuccara:
Added `updateHandlesResize` method to remove cursor `col-resize` style when updating region with resize setted to false, and to restore `col-resize` when is setted to true

`updateHandlesResize` method 

fixes #1963

### DIFF
--- a/src/plugin/regions/region.js
+++ b/src/plugin/regions/region.js
@@ -38,6 +38,8 @@ export class Region {
             left: {},
             right: {}
         };
+        this.handleLeftEl = null;
+        this.handleRightEl = null;
         this.data = params.data || {};
         this.attributes = params.attributes || {};
 
@@ -102,6 +104,7 @@ export class Region {
         }
         if (params.resize != null) {
             this.resize = Boolean(params.resize);
+            this.updateHandlesResize(this.resize);
         }
         if (params.drag != null) {
             this.drag = Boolean(params.drag);
@@ -185,15 +188,15 @@ export class Region {
 
         /* Resize handles */
         if (this.resize) {
-            const handleLeft = regionEl.appendChild(
+            this.handleLeftEl = regionEl.appendChild(
                 document.createElement('handle')
             );
-            const handleRight = regionEl.appendChild(
+            this.handleRightEl = regionEl.appendChild(
                 document.createElement('handle')
             );
 
-            handleLeft.className = 'wavesurfer-handle wavesurfer-handle-start';
-            handleRight.className = 'wavesurfer-handle wavesurfer-handle-end';
+            this.handleLeftEl.className = 'wavesurfer-handle wavesurfer-handle-start';
+            this.handleRightEl.className = 'wavesurfer-handle wavesurfer-handle-end';
 
             // Default CSS properties for both handles.
             const css = {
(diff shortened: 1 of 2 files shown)


src/plugin/regions/region.js
```diff
@@ -102,6 +102,7 @@ export class Region {
         }
         if (params.resize != null) {
             this.resize = Boolean(params.resize);
+            this.updateHandlesResize(this.resize);
```
osheroff, line 105: wondering if it would be better to use a javascript setter on `this.resize` to trigger the call to `updateHandlesResize`.  (@katspaugh are setters acceptable in this library?)

It seems like a user could very naturally do `region.resize = true` and expect it to work.  but maybe that's out of the scope of the PR?

src/plugin/regions/region.js
```diff
@@ -611,4 +612,15 @@ export class Region {
             });
         }
     }
+
+    updateHandlesResize(resize) {
+        const cursorStyle = resize ? 'col-resize' : 'none';
+        const childNodes = this.element.childNodes;
+
+        childNodes.length > 0 && childNodes.forEach(child => {
```
osheroff, line 620: IMO it'd be better to store the handle elements when we create them  in `render()`.  Checking `nodeName` here feels gross.

src/plugin/regions/region.js
```diff
@@ -611,4 +612,15 @@ export class Region {
...
+            if (child.nodeName === 'HANDLE') {
+                this.style(child, {...child.style, 'cursor': cursorStyle});
```
osheroff, line 622: no need for `...child.style` here.  `this.style` only sets new style key/values, see src/util/style.js

src/plugin/regions/region.js
marizuccara: Hi, but in this case, if we'll use the setter for `resize`,  we should use it for all the region fields (instead of using `update` method). But in this case all the setters should invoke the `updateRender` method, and fire the `update` event. Correct?

src/plugin/regions/region.js
marizuccara: Yes, you're right, thanks, I'm fixing this

src/plugin/regions/region.js
marizuccara: Yea, thanks for suggestion

osheroff:
sorry I din't get a notification that you had pushed new code.  LGTM.

PR #1990, opened 2020-07-02
add a call to drawBuffer in `loadMediaElement`

osheroff:
PR #1969 removed the call to MediaElement#load, but this means that we weren't properly triggering the waveform to render in the case where peaks were provided.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,13 +4,11 @@ wavesurfer.js changelog
 Next (unreleased)
 -----------------
 
-- Always call drawBuffer when for media-elements when peaks are provided, fixes
-  a missed call uncovered by #1969 (#1990)
 - Use `clip-path` instead of `width` changes in the multicanvcas drawer
   to avoid layout thrashing. (#1983)
 - Improved and unified loop playback logic in regions plugin. (#1868)
 - Don't call HTMLMediaElement#load when given peaks and preload == 'none'.
-  Prevents browsers from pre-fetching audio (#1969)
+  Prevents browsers from pre-fetching audio (#1969, #1990)
 
 4.0.1 (23.06.2020)
 ------------------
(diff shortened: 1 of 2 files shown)


marizuccara:
Hi , what do you mean with "not properly triggered"? The `drawBuffer` method is triggered when the media element fires the `canplay` method, no?

osheroff:
Yes, but this event doesn’t fire when preload is ‘none’ and we don’t call .load.

CHANGES.md
```diff
@@ -4,6 +4,8 @@ wavesurfer.js changelog
 Next (unreleased)
 -----------------
 
+- Always call drawBuffer when for media-elements when peaks are provided, fixes
+  a missed call uncovered by #1969 (#1990)
```
thijstriemstra, line 8: typo in sentence ("when for"?) and use backticks for `drawBuffer`.

thijstriemstra: since #1969 hasn't been released yet you can leave out this changelog entry completely (and perhaps add reference to pull request instead to changelog entry of #1969).

src/wavesurfer.js
```diff
@@ -1459,13 +1459,14 @@ export default class WaveSurfer extends util.Observer {
...
-        // If no pre-decoded peaks provided or pre-decoded peaks are
-        // provided with forceDecode flag, attempt to download the
-        // audio file and decode it with Web Audio.
         if (peaks) {
             this.backend.setPeaks(peaks, duration);
+            this.drawBuffer();
         }
 
+        // If no pre-decoded peaks provided or pre-decoded peaks are
+        // provided with forceDecode flag, attempt to download the
```
thijstriemstra, line 1468: "are provided"

osheroff:
changelog, comment updated

PR #2002, opened 2020-07-10
regions plugin: fix region ignoring minLength when being resized

### LINKED ISSUE #2001: Regions Plugin: Region minLength is not respected when resizing
Hi all, I've been using wavesurfer for quite a while and I noticed this bug. When resizing regions using the handle, you can create regions that are of 0 length. To fix this you can use the region's minLength setting (which is not documented, and probably should be). But even when using minLength, you can continue dragging regions which "pushes" them along when you drag past the minLength. I'm making a PR right now.

Check the wavesurfer Regions plugin example page.

Grab the region by the start or end handle, and drag to shrink it. If you keep dragging, it will move the region along. When setting minLength, it will keep pushing the region instead of stopping.

sundayz:

- Fix region onResize code not checking minLength, which allowed you to "push" regions while dragging them past the point where they should stop shrinking. See related issue for better description.
- Fix handles CSS in region example. They were set to width: 1% which made them scale with the size of the region, and they became invisible when the region was small enough. Now they're always 2px.

None

None

### Todos/Notes:
The resizing behaviour in general could do with some improvements. For example I don't think you should be able to drag regions to size 0, because then you can't grab them again and you lose them. Maybe regions could have a default minLength: 0.1 or so.

Other possible future enhancements:
- minLength is not documented in the example page.
- minLength/maxLength don't seem to be used a lot in the code, maybe they were never fully implemented?
- Resizing delta is calculated by checking difference between current and last mouse position. It would be better if you could only resize the region when your mouse was near the handle (maybe calculate delta as handle position change)
- Other functions which affect region size should probably check minLength, not just onResize

fixes #2001

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,6 +10,7 @@ Next (unreleased)
 - Regions plugin:
   - Removed 'col-resize' cursor when resize is disabled (#1985)
   - Improved and unified loop playback logic (#1868)
+  - Check `minLength` before resizing region (#2001)
 - Microphone plugin: move to separate directory (#1997)
 - Minimap plugin: move plugin to separate directory (#1999)
 - Cursor plugin: move plugin to separate directory (#1998)
--- a/example/regions/index.html
+++ b/example/regions/index.html
@@ -83,7 +83,8 @@
                     start: 5,
                     end: 7,
                     loop: false,
-                    color: 'hsla(200, 50%, 70%, 0.4)'
+                    color: 'hsla(200, 50%, 70%, 0.4)',
+                    minLength: 1,
                 }
             ],
             dragSelection: {
--- a/example/regions/index.html
+++ b/example/regions/index.html
@@ -29,6 +29,13 @@
         <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.13.1/styles/default.min.css">
         <script src="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.13.1/highlight.min.js"></script>
         <script>hljs.initHighlightingOnLoad();</script>
+
+        <!-- Fix strange default css that makes handles hard to grab when the region is small -->
+        <style>
+            .wavesurfer-region > handle {
+                width: 2px !important;
+            }
+        </style>
     </head>
 
     <body itemscope itemtype="http://schema.org/WebApplication">
--- a/example/regions/index.html
+++ b/example/regions/index.html
@@ -29,13 +29,6 @@
         <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.13.1/styles/default.min.css">
         <script src="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.13.1/highlight.min.js"></script>
         <script>hljs.initHighlightingOnLoad();</script>
-
-        <!-- Fix strange default css that makes handles hard to grab when the region is small -->
-        <style>
-            .wavesurfer-region > handle {
-                width: 2px !important;
-            }
-        </style>
     </head>
 
     <body itemscope itemtype="http://schema.org/WebApplication">
--- a/src/plugin/regions/region.js
+++ b/src/plugin/regions/region.js
@@ -606,13 +606,33 @@ export class Region {
         });
     }
 
+    /**
+     * @example
+     * onResize(-5, 'start') // Moves the start point 5 seconds back
+     * onResize(0.5, 'end') // Moves the end point 0.5 seconds forward
+     *
+     * @param {number} delta How much to add or subtract, given in seconds
+     * @param {string} direction 'start 'or 'end'
+     */
     onResize(delta, direction) {
         if (direction === 'start') {
+            // Check if changing the start by the given delta would result in the region being smaller than minLength
+            // Ignore cases where we are making the region wider rather than shrinking it
+            if (delta > 0 && Math.abs(this.end - (this.start + delta) < this.minLength)) {
+                return;
+            }
+
             this.update({
                 start: Math.min(this.start + delta, this.end),
                 end: Math.max(this.start + delta, this.end)
             });
         } else {
+            // Check if changing the end by the given delta would result in the region being smaller than minLength
+            // Ignore cases where we are making the region wider rather than shrinking it
+            if (delta < 0 && Math.abs((this.end + delta) - this.start) < this.minLength) {
+                return;
+            }
+
             this.update({
                 start: Math.min(this.end + delta, this.start),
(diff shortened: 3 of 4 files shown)


example/regions/index.html
```diff
@@ -29,6 +29,13 @@
         <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/9.13.1/styles/default.min.css">
         <script src="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.13.1/highlight.min.js"></script>
         <script>hljs.initHighlightingOnLoad();</script>
+
+        <!-- Fix strange default css that makes handles hard to grab when the region is small -->
```
thijstriemstra, line 33: can you remove word strange here (or fix the 'strange' css)?

example/regions/index.html
```diff
@@ -29,6 +29,13 @@
...
+        <style>
+            .wavesurfer-region > handle {
+                width: 2px !important;
```
osheroff, line 36: if this is happening in the example, perhaps it's something worth putting in the library itself?  probably as a `minWidth` property in the handle css...

src/plugin/regions/region.js
```diff
@@ -606,13 +606,33 @@ export class Region {
...
+     * onResize(0.5, 'end') // Moves the end point 0.5 seconds forward
+     *
+     * @param {number} delta How much to add or subtract, given in seconds
+     * @param {string} direction 'start 'or 'end'
+     */
     onResize(delta, direction) {
         if (direction === 'start') {
+            // Check if changing the start by the given delta would result in the region being smaller than minLength
+            // Ignore cases where we are making the region wider rather than shrinking it
+            if (delta > 0 && Math.abs(this.end - (this.start + delta) < this.minLength)) {
```
osheroff, line 621: sorry, what's the call to `Math.abs` doing here?

src/plugin/regions/region.js
```diff
@@ -606,13 +606,33 @@ export class Region {
...
+            }
+
             this.update({
                 start: Math.min(this.start + delta, this.end),
                 end: Math.max(this.start + delta, this.end)
             });
         } else {
+            // Check if changing the end by the given delta would result in the region being smaller than minLength
+            // Ignore cases where we are making the region wider rather than shrinking it
+            if (delta < 0 && Math.abs((this.end + delta) - this.start) < this.minLength) {
```
osheroff, line 632: same question about `Math.abs`.  Also curious why we have to check for `delta > 0` and `delta < 0`; the checks are sorta fine, but was there a specific reason to put them in?

sundayz: Since minLength isn't checked everywhere, it's actually possible to manually call this.update() and set the region to a size smaller than minLength. The delta check makes sure that if the area is too small, you can still make it bigger, but not shrink it further when dragging it with the mouse. A better fix would be to make sure the region can never be smaller than minLength, but I'm not familiar enough with this plugin to make that change.

src/plugin/regions/region.js
sundayz: Wanted to make sure there's no case where start + delta is smaller than end, but maybe there's no point in checking. Also I see now there's a typo on this line. I'll just remove both Math.abs checks.

example/regions/index.html
sundayz: Changed default to 2px in the plugin. I'm not sure why it was 1% as default. Maybe I'm missing something. If you think it would be better, we can keep it as it was but add minWidth: 1px so the handle cant disappear

marizuccara:
Hi great idea and very useful! It would be even better if the `minLength `property was not related only to a single region which you define on initialization, but also when you create a region "manually", directly on the waveform. Because actually, for the regions you create with dragging, the problem persists. Maybe it would be better to define a general minLength prop, valid for all regions, even for the new ones that will be created in the future.

CHANGES.md
```diff
@@ -10,6 +10,7 @@ Next (unreleased)
 - Regions plugin:
   - Removed 'col-resize' cursor when resize is disabled (#1985)
   - Improved and unified loop playback logic (#1868)
+  - Check `minLength` before resizing region (#2001)
```
thijstriemstra, line 13: please move it to the 'Next' section, 4.0.0 is already released.

src/plugin/regions/region.js
```diff
@@ -203,8 +203,7 @@ export class Region {
                 cursor: 'col-resize',
                 position: 'absolute',
                 top: '0px',
-                width: '1%',
-                maxWidth: '4px',
+                width: '2px',
```
marizuccara, line 206: I don't know if a fixed width would be ideal for all people, because it can be for example too large for someone.. But I'm agree with you that with a width of 1%, at some point the handle disappeared.. But maybe it also can be passed from outside, so everyone can define a custom width?

CHANGES.md
sundayz: not sure what you mean, it's under Next (unreleased) for me?

src/plugin/regions/region.js
sundayz: These are default CSS styles that get applied if you don't provide any, so yes you're right you can provide a custom width. I can change it to width: 1%, minWidth: 1px if you think most people would want the handle size to scale with the region

marizuccara: I talk about of another prop, which you can pass from outside, so you can customize not only the defined regions but also those you draw dragging. But maybe it is outside the purposes of this PR

sundayz: Sounds like a good idea. The region dragging can be improved a lot, we can make another PR later to fix that.

CHANGES.md
thijstriemstra: sorry, my mistake.

marizuccara:
> Hi great idea and very useful! It would be even better if the `minLength `property was not related only to a single region which you define on initialization, but also when you create a region "manually", directly on the waveform. Because actually, for the regions you create with dragging, the problem persists. Maybe it would be better to define a general minLength prop, valid for all regions, even for the new ones that will be created in the future.

Hi, @sundayz, are you planning to open a PR for this, or can I proceed?

sundayz:
@marizuccara You can do it if you'd like. I will try to fix #2006 soon, should we do 1 PR to fix multiple issues with regions, or do a separate PR for each issue? #1925 also added something to regions in 4.0 which should be changed to a configuration option since it stops the seeker from working within regions

thijstriemstra:
> should we do 1 PR to fix multiple issues with regions, or do a separate PR for each issue?

One per issue, it makes it easier to track changes later on. I was thinking of doing a 4.1.0 release next week, but it can wait till regions plugin improvements you talk about are merged.

Issue #2008, opened 2020-07-18
How to prevent a region handle from resizing beyond its opposite handle?


adityakrshnn:
I wanted to know if there is any way to stop resizing when it touches its opposite handle. Attaching video.

![ezgif-4-32aa03ff4e72](https://user-images.githubusercontent.com/23035769/87847999-ebff2400-c8f9-11ea-80fa-642821f4fb0e.gif)

marizuccara:
https://github.com/katspaugh/wavesurfer.js/pull/2002

PR #2009, opened 2020-07-19
regions plugin: add regionsMinLength option 

marizuccara:
Added `generalMinLength ` parameter to control min length of regions created dragging on drawer (so not those passed on initialization in `regions` array), to avoid dragging regions to size 0.

NB. Maybe unify in some way `minLength` and `generalMinLength` parameters if necesssary?

PR https://github.com/katspaugh/wavesurfer.js/pull/2002
ISSUES https://github.com/katspaugh/wavesurfer.js/issues/2001 https://github.com/katspaugh/wavesurfer.js/issues/2008


thijstriemstra:
@sundayz can you also take a look?

src/plugin/regions/region.js
```diff
@@ -614,10 +616,19 @@ export class Region {
      * @param {string} direction 'start 'or 'end'
      */
     onResize(delta, direction) {
+        let minLength = null;
+
```
sundayz, line 620: Maybe add a comment so people know a region's minLength is used over generalMinLength if both are specified

sundayz:
Maybe `regionMinLength` would be a better name, but I don't mind what name is used. example/regions/index.html should be updated to show the settings that are being used in the .js file.

When I tried testing this, generalMinLength seems to always be `undefined`? The code looks right I don't know why it's not being passed to the plugin constructor

marizuccara:
> Maybe `regionMinLength` would be a better name, but I don't mind what name is used. example/regions/index.html should be updated to show the settings that are being used in the .js file.
> 
> When I tried testing this, generalMinLength seems to always be `undefined`? The code looks right I don't know why it's not being passed to the plugin constructor
> 
> edit:
> 
> okay so I see generalMinLength only applies to regions created by dragging the mouse.
> I think it would be better if generalMinLength applied to all regions. If you want to set a different min length for regions created via the mouse it should be done by listening to the `region-created` event and modifying the region after it's created. That makes more sense to me. for example:
> 
> (example code, might not work)
> 
> ```js
> // first wait for the region plugin to be loaded
> wavesurfer.on('plugin-added', name => {
>     if (name == 'regions') {
>         // now any regions created after the plugin was loaded (i.e. via mouse drag) have minlength of 1.5
>         wavesurfer.on('region-created', region => region.minLength = 1.5)
>     } 
> })
> ```

I'm agree, thanks. Later fix

marizuccara:
So so sorry for the absence..

marizuccara:
So sorry for the absence :(
If you approve these changes, I'll update changelog end `regions/index.html` example

thijstriemstra:
@marizuccara there are many old, unrelated changes in this PR, can you take a look at that?

marizuccara:
@sundayz have you take a look to this PR if you'll have time?

example/regions/app.js
```diff
@@ -13,6 +13,7 @@ document.addEventListener('DOMContentLoaded', function() {
         backend: 'MediaElement',
         plugins: [
             WaveSurfer.regions.create({
+                regionsMinLength: 2,
```
sundayz, line 16: the HTML file shows the settings used in the example, should update that too

sundayz:
I tested the change and it works as intended:


- [x] When setting minLength AND regionsMinLength, minLength is used
- [x] When setting only minLength and no regionsMinLength, minLength is used
- [x] When setting only regionsMinLength and no minLength, regionsMinLength is used

Works well, just need to update HTML for the examples page

src/plugin/regions/region.js
thijstriemstra: ping @marizuccara

PR #2012, opened 2020-07-20
[Regions Plugin] Fix: Region dragging and resizing can now continue outside canvas.

### LINKED ISSUE #2006: How to continue dragging a wavesurfer region even outside canvas area?
I wanted to know if there is any way to continue dragging/resizing a region even outside the wavesurfer.js canvas area. Attaching video.

qWaKi

adityakrshnn:
- region dragging and resizing can now continue outside canvas.
- the cursor remains consistent on the point of contact with the region while dragging.

fixes #2006


src/plugin/regions/region.js
```diff
@@ -36,7 +36,7 @@ export class Region {
         // no styling or can be assigned an object containing CSS properties.
         this.handleStyle = params.handleStyle || {
             left: {},
-            right: {}
+            right: {},
```
thijstriemstra, line 39: this extra comma is not needed.

sundayz:
really good job! I haven't read the code but I tested your changes and I can say that resizing regions is a lot smoother now. It definitely works as the end user would expect it to now.

adityakrshnn:
Haven't tested on mobile. But I have changed the scope of both mousemove and touchmove. I believe it should work the same.

src/plugin/regions/region.js
adityakrshnn: For now I have removed all the dangling commas that I had put.

thijstriemstra: thanks, I enabled the `comma-dangle` eslint rule in master branch.

PR #2023, opened 2020-07-26
regions plugin: edgescroll now works for both edges

### LINKED ISSUE #2011: Edgescroll does not work for the right edge
The edgescroll does not seem to work for the right edge. Attaching video.
edgescroll_gif

adityakrshnn:
- Edgescroll (i.e. dragging and resizing near the edges automatically continues) now works for both the edges

fixes #2011 

### Steps to test
- Enabled scrollParent in the zoom plugin example so that edgescroll works.


marizuccara:
Hi, thanks! Good  job! I've tested it and I have one observation:
- When I drag the region, the edgescroll scrolls well! But when I drag a region, it disappears from the view.. Maybe it will be better if It stayed in the visible view, instead of disappearing besides the edges?

adityakrshnn:
Yes. I did consider keeping the entire region visible at all times. I wanted to skip the 30px threshold area and enable dragging **when the edges of the region reached the edges of the visible wave**.  In an ideal case, when the edge is still inside the visible wave, it will only edge scroll when the edge comes close to the edge of the visible wave. However, there was one case that would've made that behaviour inconsistent. If we could decide to pick a middle ground then that could be solved.

The case when one or both the edges of the region are already beyond the visible wave:
In this case, if we start dragging towards the invisible edge then it would immediately start edge scrolling.

wavesurfer

What should be the behaviour over here, if we go with the ideal case (edge scroll only when region edge comes close to the visible edge)? One solution could be to not edge scroll at all in such a case and only scroll when an edge touches the visible edge

marizuccara:
I think that the case when the edges of the region are already beyond the visible view, because is bigger, is correct. 
I talk about the situation in which you have a smaller region. If you scroll, the region becomes not visible. Maybe in these scenarios, left a piece of region visible?

image

adityakrshnn:
I am unsure of what you meant. Is this what you want?

ezgif-7-56f67ce708b1


How will this work for large regions where the handles are beyond visible region?

marizuccara:
Mhmm.. You said "One solution could be to not edge scroll at all in such a case and only scroll when an edge touches the visible edge".. In this case I think that your solution is better..

@sundayz can you take a look at this PR?

thijstriemstra:
ps. can you add @sundayz to the team @katspaugh?

katspaugh:
Invited! 👍

sundayz:
Thanks for the invite guys :) I'll try out this PR this evening

sundayz:
I’m really busy atm getting a new floor fitted into the house so have very restricted access to PC right now, can’t review the PR unfortunately. Will probably have time Sunday or Monday evening 

When are you guys thinking of releasing 4.1?

thijstriemstra:
> When are you guys thinking of releasing 4.1?

Sounds good to me.

thijstriemstra:
I don't like the usage of a hard-coded value of 30px in this PR.

adityakrshnn:
I could change it to 5% of the waveform's width. Will that be ok?

thijstriemstra:
> I could change it to 5% of the waveform's width. Will that be ok?

I think it should be a value that can be configured. The default can be 30px or 5%, whatever works well.

adityakrshnn:
Could be done. What would the value be called?

thijstriemstra:
something with margin, or edge?

adityakrshnn:
I have pushed some new commits in case you haven't checked them out.

src/wavesurfer.js
```diff
@@ -61,6 +61,7 @@ import MediaElementWebAudio from './mediaelement-webaudio';
  * attributes.
  * @property {number} duration=null Optional audio length so pre-rendered peaks
  * can be display immediately for example.
+ * @property {number} edgeScrollWidth='5% from container edges' Optional width for edgeScroll to start
```
thijstriemstra, line 64: this new param applies to the regions plugin, not wavesurfer core, so it shouldn't be documented here.

adityakrshnn: Fixed it

src/plugin/regions/index.js
```diff
@@ -129,6 +130,7 @@ export default class RegionsPlugin {
             this.wrapper = this.wavesurfer.drawer.wrapper;
             if (this.params.regions) {
                 this.params.regions.forEach(region => {
+                    region.edgeScrollWidth = this.params.edgeScrollWidth ?? this.wrapper.clientWidth * 0.05;
```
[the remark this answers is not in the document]
thijstriemstra: good to know @sundayz. Let's not use them till it's supported in preset-env. Ideally, we add an eslint rule that enforces this in the codebase.

PR #2024, opened 2020-07-26
regions plugin: revert PR #1926 fix click propagation 

### LINKED ISSUE #2013: [Question] [Regions plugin] How to get time from region-click?
I wanted to know if there is anyway to get time from region-click event?

### LINKED ISSUE #2029: The regions can't be clicked
# The regions can't be clicked

### The previous version can be clicked and played from the click point.

wavesurfer_cannot_clicked

### The previous version can be clicked

wavesurfer_can_clicked

sundayz:
Revert change made in #1926 because as the author says, it's possible to take the event object and call e.stopPropagation() yourself if you need it. He probably didn't notice that the event was actually being passed in the region-click event.

If stopPropagation() is called too early, it's impossible to let users move the seeker while clicking inside a region.
If you want to play a region on click like in the example in the original PR just do:
```js
wavesurfer.on('region-click', (region, event) => {
    event.stopPropagantion();
    region.play();
});
```
and that plays the region you clicked without setting the seeker.

The double click event handler also calls stopPropagation(), I'm not sure if it should, but I dont think double clicking the waveform does anything.

Reverts #1926
Fixes #2013
Closes #2029


thijstriemstra:
can you take a look at this PR @iboshkov?

sundayz:
Added changelog, let me know if its ok

PR #2026, opened 2020-07-26
renamed requestHeaders to headers in util.fetchFile()

silentauben:


thijstriemstra:
You should be using the `requestHeaders` property of the `xhr` object, not `headers`. it's not a typo.

PR #2034, opened 2020-08-03
Trigger 'waveform-ready' event when provided peaks are drawn

### LINKED ISSUE #2031: BUG: waveform-ready event isn't triggered anymore if mediaelements and peaks and preload is none
Hi,

Wavesurfer v4.0.1.x, with FireFox

It seems waveform-ready event isn't triggered if backend is set to MediaElement, peaks are providen and preload is set to none.

It is weird cause the waveform is drawn so we might expect this event to be triggered.

Probably related to https://github.com/katspaugh/wavesurfer.js/pull/1990

Thanks !

marizuccara:
Trigger `waveform-ready` event when provided peaks are drawn and backend is of type of `MediaElement` or `MediaElementWebAudio`. 
This event becomes useful especially if the audio resource provided is an url and the preload attribute is of type `none`, because in this case the `ready` event is triggered only after the audio starts playing (`canplay`),

fixes #2031
PR: https://github.com/katspaugh/wavesurfer.js/pull/1990


thijstriemstra:
It would be good to have a unit test for this.

spec/test-helpers.js
```diff
@@ -10,6 +10,15 @@ const TestHelpers = {
     /** Length of example audio clip (in seconds) */
     EXAMPLE_FILE_DURATION: 21,
 
+    /** Example audio clip with waveform */
+    EXAMPLE_2_FILE_PATH: '/base/spec/support/stereo.mp3',
```
thijstriemstra, line 14: can you rename this to `EXAMPLE_STEREO_FILE_PATH` and change the other names too?

spec/mediaelement-shared.js
```diff
@@ -300,3 +308,11 @@ function loadElement() {
     audioElement.src = TestHelpers.EXAMPLE_FILE_PATH;
     wavesurfer.load(audioElement);
 }
+
+function loadAudioPeaks() {
```
thijstriemstra, line 312: can you give this function a docstring?

spec/mediaelement-shared.js
```diff
@@ -32,6 +32,14 @@ export function sharedTests(backend) {
         loadElement();
     });
 
+    /**
+     * @test {WaveSurfer#waveformReady}
+     */
+    it('should be waveform ready', function(done) {
```
thijstriemstra, line 38: can you mention the event in the test description?

sundayz, line 38: I think this test should check that `waveform-ready` is only emitted once (not 2 or 3 times). This catches bugs where someone changes the if statements inside `loadMediaElement`. We can do this with jasmine spy functions:

```js
const waveformReadySpy = jasmine.createSpy('waveform-ready-spy');
waveform.once('waveform-ready', () => {
    waveformReadySpy();
});

// catch bugs where waveform-ready is called more than once during loading
expect(waveformReadySpy).toHaveBeenCalledTimes(1);
```

source:
https://stackoverflow.com/questions/34938016/testing-if-an-event-has-been-triggered-in-jasmine

Note: I don't use jasmine so maybe there is a better way to do this :)

marizuccara: yeaah, thanks!

bwestergard:
@marizuccara I believe these changes are incomplete; there is at least one other place we need to be firing the `waveform-ready` event. I will make a PR into your branch shortly.

sundayz:
> 
> 
> @marizuccara Thank you so much for this contribution!
> 
> I believe these changes are ever-so-slightly incomplete. I believe we need to be firing the `waveform-ready` event here as well: [npr@52596c9#diff-4661d6bcf4ba307b2071080d540b7bc3R1418](https://github.com/npr/wavesurfer.js/commit/52596c9f6aef8dfe90051e0feb5be3b8c99e158b#diff-4661d6bcf4ba307b2071080d540b7bc3R1418)

I think that method is only called by WebAudio. The documentation says:

`waveform-ready Fires after the waveform is drawn when using the MediaElement backend. If you're using the WebAudio backend, you can use ready.`

So not emitting waveform-ready from loadBuffer is "technically" correct. But in my opinion the best solution would be to have both backends use `ready` to signal that the waveform has been drawn, and make a separate event that signals when the .mp3 file has been fully processed. Since I think that's what `ready` does on WebAudio - it only fires when audio is fully available

bwestergard:
@sundayz: In my case, I just need there to be an event that fires after peaks provided as the second argument to `.load` are rendered. I don't have an opinion as to which event this should be. I think the path of least resistance for @marizuccara's contribution would be to add [this line](https://github.com/npr/wavesurfer.js/commit/52596c9f6aef8dfe90051e0feb5be3b8c99e158b#diff-4661d6bcf4ba307b2071080d540b7bc3R1418). But introducing a new event, rather than extending the conditions under which `waveform-ready` is fired, would serve my purposes just as well.

sundayz:
I think “ready” does what you describe (fires after peaks are rendered). And “waveform-ready” fires after peaks are rendered AND audio is ready to play.

If that’s not how it works then yes we should probably add a new event.

bwestergard:
> I think “ready” does what you describe (fires after peaks are rendered). And “waveform-ready” fires after peaks are rendered AND audio is ready to play.
> 
> If that’s not how it works then yes we should probably add a new event.

I have just checked, and when I call the `load` method with peaks, `ready` does not fire. So yes, it seems we do need a new event that fires when `load` called with peaks completes.

thijstriemstra:
> So yes, it seems we do need a new event that fires when load called with peaks completes. 

Please make that change in a new pr if you want.

thijstriemstra:
Closing in favor of #2077

PR #2045, opened 2020-08-20
Fix performance issues with seeking while audio is playing

todoroff:
When making calls to the `seekTo` method while the audio is playing, we were experiencing a issues with console errors and performance (to the point of the UI becoming completely unresponsive in the case of high-frequency seeking) because of unnecessary calls to `start` and `pause` inside that method.

Making the `seekTo` method `async` and awaiting the calls to `pause` and `start` inside solved the errors and the performance hit for low-frequency seeks (less than ~3 seeks per second) but wasn't good enough for higher-frequency seeking due to too much delay to resolve the promises.

We finally decided to remove those calls altogether, because they appear to be redundant. They were first introduced in #918, and supposedly solved an issue with backward seeking on `MediaElement`, which I wasn't able to replicate (tested with a 30 min long .ogg file).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 wavesurfer.js changelog
 =======================
 
+Next (unreleased)
+-----------------
+
+- Fix performance issues with `seekTo` while audio is playing (#2045)
+
 4.1.1 (24.09.2020)
 ------------------
 - Revert Code cleanup for Observer class (#2069)
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -900,20 +900,12 @@ export default class WaveSurfer extends util.Observer {
         }
         this.fireEvent('interaction', () => this.seekTo(progress));
 
-        const paused = this.backend.isPaused();
-        // avoid draw wrong position while playing backward seeking
-        if (!paused) {
-            this.backend.pause();
-        }
         // avoid small scrolls while paused seeking
         const oldScrollParent = this.params.scrollParent;
         this.params.scrollParent = false;
         this.backend.seekTo(progress * this.getDuration());
         this.drawer.progress(progress);
 
-        if (!paused) {
-            this.backend.play();
-        }
         this.params.scrollParent = oldScrollParent;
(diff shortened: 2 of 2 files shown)


todoroff:
> It should probably be removed, but did you run the javascript profiler on your code to confirm that the calls to play() and pause() are slow?
> 
> It's been 6 years so the original issue with mediaelement is probably fixed and we can remove the play/pause calls

I haven't run a profiler.

What happens is that if you ca play and pause too fast one after the other, you get errors `DOMException: The play() request was interrupted`. After several of those, the entire browser bogs down until you hit pause.

If you do make sure those calls run sequentially, meaning after each one is complete, there are no UI responsiveness issues but you can't seek as fast as when the audio is not playing, which simply doesn't feel nice.

osheroff:
i'd definitely check this against a bunch of browsers and html-audio vs media-element combos.  iOS/Safari tends towards "wildly different in behavior" vs firefox/chrome

sundayz:
BTW if you want high performance you should be using WebAudio instead of MediaElement.

I assume you want to allow users to seek by dragging the cursor with the mouse during playback? With WebAudio you have more fine control over audio buffers etc which is needed for something like that.

realies:
Works fine on Safari 13.1.2 (15609.3.5.1.3) over here.

todoroff:
@osheroff Works on Safari.
@sundayz We need MediaElement for our usecase.

thijstriemstra (changes requested):
thanks @todoroff. I think we should give it a try; less is more. Can you add a changelog entry?

CHANGES.md
```diff
@@ -4,6 +4,7 @@ wavesurfer.js changelog
 4.1.1 (24.09.2020)
 ------------------
 - Revert Code cleanup for Observer class (#2069)
+- Fix performance issues with `seekTo` while audio is playing (#2045)
```
thijstriemstra, line 7: can you move this entry to a new section on top called 

```
Next (unreleased)
------------------
```

thijstriemstra (approved):

PR #2064, opened 2020-09-15
RFC: defer script node creation until play time

osheroff:
hey folks,

I've been try to trace down a weird issue where my webpage is using 50% of a cpu core when the webpage is entirely idle.  The webpage has many, many (idle) instances of wavesurfer present.

I dove all the way down to the bottom and found that chrome begins to chew CPU as soon as the `AudioScriptNode` is instantiated.  With this change, an idle webpage with 50 or so wavesurfer instances sits at around 7% cpu as opposed to about 16-22%.  It seems that the very act of creating a webaudio context at all incurs some cpu time.

Now, this is technically a breaking change as users who rely on the script node may be in for an unwelcome surprise.  We *technically* don't expose the interface, but hey, it's javascript, reach down into whatever you like.

thoughts?  I can also work around this issue by simply defering wavesurfer initialization until play-time entirely, but I thought it'd be nice to try to fix it the core.

Perhaps better would be to lazy-instantiate the node using `Object.defineProperty`?  dunno.

### DIFF
--- a/src/mediaelement-webaudio.js
+++ b/src/mediaelement-webaudio.js
@@ -30,7 +30,6 @@ export default class MediaElementWebAudio extends MediaElement {
         this.setPlaybackRate(this.params.audioRate);
         this.createTimer();
         this.createVolumeNode();
-        this.createScriptNode();
         this.createAnalyserNode();
     }
     /**
--- a/src/webaudio.js
+++ b/src/webaudio.js
@@ -21,6 +21,7 @@ export default class WebAudio extends util.Observer {
     stateBehaviors = {
         [PLAYING]: {
             init() {
+                this.createScriptNode();
                 this.addOnAudioProcess();
             },
             getPlayedPercents() {
@@ -156,7 +157,6 @@ export default class WebAudio extends util.Observer {
      */
     init() {
         this.createVolumeNode();
-        this.createScriptNode();
         this.createAnalyserNode();
 
         this.setState(PAUSED);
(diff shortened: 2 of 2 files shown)


sundayz:
7% still sounds a bit high if there is nothing happening on the page. We also use wavesurfer and it's not uncommon to have 50+ instances of wavesurfer on a single page, but when I test with 65 instances of wavesurfer, my browser sits at 1-2% CPU usage when idle. This is when I use preloaded peaks, do you use those too or do you let wavesurfer generate the waveform?

osheroff:
@sundayz preloaded peaks all the way.  in production it's actually more like 30%. what backend are you using?  what OS?

sundayz:
Windows 10, MediaElementWebAudio.

Tried with WebAudio and no preloaded peaks in firefox, CPU idles at 0.5% - 1%.

maybe check the performance tab to see if there are events firing somewhere even if youre not doing anything on the page.

osheroff:
trust me i went all the way down the rabbit hole.  it's that.  on macOS

thijstriemstra:
Haven't tested but this looks like it might break the mic plugin.

sundayz:
People including myself have noticed that audio with MediaElementWebAudio is choppy on iOS, could this be related?

thijstriemstra:
AudioWorklet seems to be supported in all major browsers now. Switching to that (#1281) would make this workaround obsolete, wouldn't it?

osheroff:
maybe?  depends on the implementation of it i suppose.  I ended up lazy-loading wavesurfer itself which works reasonably well, but let me know if you've got an implementation you'd like me to test with lots of instances.

PR #2069, opened 2020-09-24
Revert "Code cleanup for Observer class (#2041)"

### LINKED ISSUE #2067: this.handlers is undefined
### Wavesurfer.js version(s):
4.1.0

Region plugin doesn't load, check example page:

https://wavesurfer-js.org/plugins/regions.html

When calling `this.regions.add(options)`.
Line 199 of regions/index.js registers this event:
```
        region.on('remove', () => {
            delete this.list[region.id];
        });
```

Results in `TypeError: can't access property "remove", this.handlers is undefined`

Last change to observer: https://github.com/katspaugh/wavesurfer.js/commit/590d9ccc96f6c69090849f2e5cfb4587da647ee5
Looks like removing the lazy creation of handlers object broke the regions plugin.

Found this comment in observer.js: `* plugins borrow part of this class without calling the constructor`
Can we do something about that without breaking existing plugins? Not calling the constructor doesn't sound very good.

Worst case scenario we revert to what we were doing before.

sundayz:
This reverts commit 590d9ccc96f6c69090849f2e5cfb4587da647ee5.

Fixes #2067 (this.handlers is undefined)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 wavesurfer.js changelog
 =======================
 
+4.1.1 (unreleased)
+------------------
+- Revert Code cleanup for Observer class (#2069)
+
 4.1.0 (16.09.2020)
 ------------------
 
--- a/src/util/observer.js
+++ b/src/util/observer.js
@@ -14,17 +14,13 @@ export default class Observer {
      */
     constructor() {
         /**
-         * @type {string[]}
          * @private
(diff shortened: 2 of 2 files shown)


sundayz:
@thijstriemstra what version number

thijstriemstra:
@sundayz sounds good.

thijstriemstra (approved):

PR #2108, opened 2020-11-13
Introduce relativeNormalization option for split channels rendering

Claudiohbsantos:
Introduced new splitChannelsOption parameter relativeNormalization, that allows wavesurfer to normalize channels while keeping them proportional to each other. Currently normalize normalizes each channel independently, so it's impossible to visually tell when a channel is louder than the others after normalization.
Also fixed assignment of default parameters to splitChannelsOptions, so partial options won't cause bugs anymore. 

None

None. Added parameter to MultiCanvas.prepareDraw that can be left undefined in order to prevent breaking current calls. 

### Todos/Notes:
Added esdoc documentation for splitChannelsOptions. Still need to update gh-pages if PR is accepted

### DIFF
--- a/example/split-channels/index.html
+++ b/example/split-channels/index.html
@@ -128,7 +128,7 @@
                 <code>filterChannels</code> - array - Array of channel numbers. Channels included in the array will not be drawn.
             </p>
             <p>
-                <code>relativeNormalization</code> - boolean - Determines whether when <code>normalize</code> and <code>splitChannels</code> are both true the channels will be normalized individually or proportionally to each other. Defaults to false (each channel will be normalized in isolation). 
+                <code>relativeNormalization</code> - boolean - When <code>normalize</code> and <code>splitChannels</code> are both true the channels will be normalized individually or proportionally to each other. Defaults to <code>false</code> (each channel will be normalized in isolation). 
             </p>
 
 
--- a/example/split-channels/index.html
+++ b/example/split-channels/index.html
@@ -127,6 +127,9 @@
             <p>
                 <code>filterChannels</code> - array - Array of channel numbers. Channels included in the array will not be drawn.
             </p>
+            <p>
+                <code>relativeNormalization</code> - boolean - Determines whether when <code>normalize</code> and <code>splitChannels</code> are both true the channels will be normalized individually or proportionally to each other. Defaults to false (each channel will be normalized in isolation). 
+            </p>
 
 
             <div class="footer row">
--- a/src/drawer.multicanvas.js
+++ b/src/drawer.multicanvas.js
@@ -456,7 +456,7 @@ export default class MultiCanvas extends Drawer {
      * rendered
      * @param {function} fn The render function to call, e.g. `drawWave`
      * @param {number} drawIndex The index of the current channel after filtering.
-     * @param {number?} normalizedMax maximum modulation value accross channels for use with relativeNormalization. Ignored when undefined
+     * @param {number?} normalizedMax maximum modulation value across channels for use with relativeNormalization. Ignored when undefined
      * @returns {void}
      */
     prepareDraw(peaks, channelIndex, start, end, fn, drawIndex, normalizedMax) {
--- a/src/drawer.multicanvas.js
+++ b/src/drawer.multicanvas.js
@@ -477,6 +477,7 @@ export default class MultiCanvas extends Drawer {
 
                     let overallAbsMax;
                     if (this.params.splitChannelsOptions && this.params.splitChannelsOptions.relativeNormalization) {
+                        // calculate maximum peak across channels to use for normalization
                         overallAbsMax = util.max(channels.map((channelPeaks => util.absMax(channelPeaks))));
                     }
 
--- a/src/drawer.multicanvas.js
+++ b/src/drawer.multicanvas.js
@@ -456,7 +456,7 @@ export default class MultiCanvas extends Drawer {
      * rendered
      * @param {function} fn The render function to call, e.g. `drawWave`
      * @param {number} drawIndex The index of the current channel after filtering.
-     * @param {number?} normalizedMax maximum modulation value across channels for use with relativeNormalization. Ignored when undefined
+     * @param {number?} normalizedMax Maximum modulation value across channels for use with relativeNormalization. Ignored when undefined
      * @returns {void}
      */
     prepareDraw(peaks, channelIndex, start, end, fn, drawIndex, normalizedMax) {
--- a/src/drawer.multicanvas.js
+++ b/src/drawer.multicanvas.js
@@ -456,9 +456,10 @@ export default class MultiCanvas extends Drawer {
      * rendered
      * @param {function} fn The render function to call, e.g. `drawWave`
      * @param {number} drawIndex The index of the current channel after filtering.
+     * @param {number?} normalizedMax maximum modulation value accross channels for use with relativeNormalization. Ignored when undefined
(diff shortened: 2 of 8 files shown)


example/split-channels/index.html
```diff
@@ -127,6 +127,9 @@ <h4>splitChannelOptions</h4>
             <p>
                 <code>filterChannels</code> - array - Array of channel numbers. Channels included in the array will not be drawn.
             </p>
+            <p>
+                <code>relativeNormalization</code> - boolean - Determines whether when <code>normalize</code> and <code>splitChannels</code> are both true the channels will be normalized individually or proportionally to each other. Defaults to false (each channel will be normalized in isolation).
```
thijstriemstra, line 131: remove 'Determines whether'  and `false` needs to be wrapped in `<code>`

src/drawer.multicanvas.js
```diff
@@ -456,9 +456,10 @@ export default class MultiCanvas extends Drawer {
      * rendered
      * @param {function} fn The render function to call, e.g. `drawWave`
      * @param {number} drawIndex The index of the current channel after filtering.
+     * @param {number?} normalizedMax maximum modulation value accross channels for use with relativeNormalization. Ignored when undefined
```
thijstriemstra, line 459: typo: across

src/drawer.multicanvas.js
```diff
@@ -474,8 +475,14 @@ export default class MultiCanvas extends Drawer {
                         );
                     }
 
+                    let overallAbsMax;
+                    if (this.params.splitChannelsOptions && this.params.splitChannelsOptions.relativeNormalization) {
```
thijstriemstra, line 479: add a comment describing what this does

src/drawer.multicanvas.js
```diff
@@ -456,9 +456,10 @@ export default class MultiCanvas extends Drawer {
      * rendered
      * @param {function} fn The render function to call, e.g. `drawWave`
      * @param {number} drawIndex The index of the current channel after filtering.
+     * @param {number?} normalizedMax maximum modulation value accross channels for use with relativeNormalization. Ignored when undefined
      * @returns {void}
      */
-    prepareDraw(peaks, channelIndex, start, end, fn, drawIndex) {
+    prepareDraw(peaks, channelIndex, start, end, fn, drawIndex, normalizedMax) {
```
thijstriemstra, line 462: normalizedMax param should be documented.

src/util/absMax.js
```diff
@@ -0,0 +1,15 @@
+import utilMax from './max';
+import utilmin from './min';
+
+/**
+ * Get the largest absolute value
```
thijstriemstra, line 5: add unit test(s) for this new util

src/wavesurfer.js
```diff
@@ -149,6 +150,25 @@ import MediaElementWebAudio from './mediaelement-webaudio';
...
+ *         progressColor: 'green',
+ *         waveColor: 'pink'
+ *     },
+ *     1: {
+ *         progressColor: 'orange',
+ *         waveColor: 'purple'
+ *     }
+ * }
+ * @property {number[]} filterChannels=[] indexes of channels to be hidden from rendering
+ * @property {boolean} relativeNormalization=false determines whether normalization is done per channel or maintains proportionality between channels. Only applied when normalize and splitChannels are both true.
```
thijstriemstra, line 169: add version marker when this was introduced

thijstriemstra, line 169: try to keep lines below char length of < 100

src/drawer.multicanvas.js
Claudiohbsantos: I added  ESDocs documentation  for the parameter on line 459 in the original pull request. Is there another place I should document the parameter, or is the documentation insufficient?

Current docs:  `* @param {number?} normalizedMax Maximum modulation value across channels for use with relativeNormalization. Ignored when undefined`

src/wavesurfer.js
Claudiohbsantos: I'm sorry, I'm not sure I understand how to do that. I am checking out esdoc and jsdoc documentation for references for something similar but I haven't found anything that seems pertinent yet. Do you have an example I could look at for format?

src/drawer.multicanvas.js
thijstriemstra: ah yes.

thijstriemstra (approved):

osheroff:
might rename this `stereoNormalization` and default it to false

osheroff:
otherwise lgtm

Claudiohbsantos:
> might rename this `stereoNormalization` and default it to false

I was struggling with a name for it as well. I'm just not sure about "stereo" because that implies 2 channels, and this would apply to any multichannel audio. The use case inspired this addition for example is when showing audio files recorded on tv/film shoots, which contain any number of channels and are correlated by level.

thijstriemstra:

Claudiohbsantos:
> can you also add a changelog entry?

Sure, I'll add it to Changes.md.

Just realized the @version tag I added on commit 136f678 is probably wrong, right? I tagged 4.2.0 but it should be tagged for the next release, whichever version it ends up being?

thijstriemstra:
> I tagged 4.2.0 but it should be tagged for the next release, whichever version it ends up being?

4.2.0 is fine as is. you can also rename next in changelog to 4.2.0 if you want.

Claudiohbsantos:
> > I tagged 4.2.0 but it should be tagged for the next release, whichever version it ends up being?
> 
> 4.2.0 is fine as is. you can also rename next in changelog to 4.2.0 if you want.

I added under heading * Next (unreleased)* because I am a bit confused with the versioning since I thought 4.2.0 was already released.

thijstriemstra:
oops, my bad, i meant 4.3.0

Claudiohbsantos:
> oops, my bad, i meant 4.3.0

Cool, I thought i was missing something =]. Updated the version tag and the changelog to 4.3.0

src/wavesurfer.js
```diff
@@ -149,6 +150,28 @@ import MediaElementWebAudio from './mediaelement-webaudio';
...
+ *     1: {
+ *         progressColor: 'orange',
+ *         waveColor: 'purple'
+ *     }
+ * }
+ * @property {number[]} filterChannels=[] indexes of channels to be hidden from rendering
+ * @property {boolean} relativeNormalization=false determines whether
+ * normalization is done per channel or maintains proportionality between
+ * channels. Only applied when normalize and splitChannels are both true.
+ * @version 4.3.0
```
thijstriemstra, line 172: use https://jsdoc.app/tags-since.html instead

src/util/absMax.js
thijstriemstra, line 5: "in an array" or something like that

src/util/absMax.js
```diff
@@ -0,0 +1,15 @@
+import utilMax from './max';
+import utilmin from './min';
+
+/**
+ * Get the largest absolute value
+ *
+ * @param   {Array} values Array of numbers
+ * @returns {Number} Largest number found
+ * @example console.log(max([-3, 2, 1]), max([-3, 2, 4])); // logs 3 4
```
thijstriemstra, line 9: @since marker missing

src/wavesurfer.js
Claudiohbsantos: Thanks, I didn't know about this tag.

marizuccara (approved):

Issue #2109, opened 2020-11-16
RFC: Typescript for Wavesurfer.js


sundayz:
Hi all.

Over time, more and more people have suggested that wavesurfer uses typescript, or at least provides up to date typescript definitions. (#2038 #1209 etc)

I have started work on incrementally moving Wavesurfer to Typescript.

**Follow the progress here:**
- [ ] Part 1  #2110

Typescript has the following benefits:
- Static type checking means less chance to accidentally write broken code.
- Smaller barrier to entry for new contributors, they can write new code with confidence.
- The end product is still Javascript code, so end users get all the benefits for no extra cost. Typescript users don't have to download separate types.
- Typescript has some useful linting functionalities like no unused variables, etc.

The 3rd party wavesurfer types on the definitely typed repository are outdated. This is causing issues for some people who are trying to use wavesurfer in a typescript project. By slowly adopting typescript we remove the need for separate types which would need to be kept in sync with the project. 

Progress as of right now is that the code builds correctly as far as I can tell, and I moved util/max.js to util/max.ts and it seems like everything is going well.

Let me know what you think, if this is worth it, etc. There is still some things left to configure. We will need to use relaxed typescript rules as we convert js files to ts, then we can apply additional restrictions like no "any", strict function types etc.

thijstriemstra:
Sounds good.

sundayz:
Sure. I’ll start by converting the Wavesurfer class to typescript and we can also look at the tsconfig file for the first PR.

thijstriemstra:
I would start with something simple like util classes. But maybe entry point (Wavesurfer) needs to be done first, dunno.

sundayz:
It doesn't need to be first but I guess it would bring most value. We can start with util classes though.

nadineshmo:
Hi, thanks so much for your work! @sundayz do you have a rough estimate to when the cursor plugin might work for typescript? Just wondering. Thanks!

sundayz:
@nadineshmo Adding type definitions to the code is pretty easy and won't take long, but I have to rework #2110 and I'm a bit busy with work related things right now to do it. I might have time later this week to re-do the PR to change the build system to typescript.

caugner:
@sundayz I opened a PR in the DefinitelyTyped repository to complete the current TypeScript definitions (especially adding those for the Wavesurfer.js plugins) : https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52429

[...]

PR #434, opened 2015-04-11
add start of test suite

thijstriemstra:
refs #427 

Use `grunt`, `grunt test` or `grunt coverage` to run the tests. I added a single test for getId in util to show the setup is working.

PR #2110, opened 2020-11-20
WIP: Typescript Port Initial Setup

sundayz:
Part 1 of moving Wavesurfer to Typescript. As suggested in #2109 we start by rewriting the utility functions to typescript.

Because this is the first typescript PR we have to do some setup. Suggestions and improvements to the tsconfig are welcome. Also to the webpack config.

**Help wanted**
- [x] Import paths require you to specify the .ts extension right now. This is probably a webpack configuration error. It should be fixed before this is merged.
- Any improvements to the tsconfig file.
- Source maps for debugging? Typescript can output them but I think we need a webpack loader for them.

Most important thing about this PR is that once all this is finished, the code that is generated should not have different behaviour than before.

Sometimes we will need to change old code to make it work in typescript, so watch out for that! The end result should be the same even if the code is slightly different.

Helpful links: https://www.typescriptlang.org/docs/handbook/migrating-from-javascript.html


src/drawer.canvasentry.js
```diff
@@ -3,7 +3,7 @@
  */
 
 import style from './util/style';
-import getId from './util/get-id';
+import getId from './util/get-id.ts';
```
sundayz, line 6: Need help to fix this. you should be able to import typescript files without the extension

src/util/index.js
```diff
@@ -1,11 +1,11 @@
-export { default as getId } from './get-id';
-export { default as max } from './max';
-export { default as min } from './min';
-export { default as Observer } from './observer';
+export { default as getId } from './get-id.ts';
```
sundayz, line 1: imports here need to be fixed (.ts extension shouldnt be necessary)

src/util/clamp.ts
```diff
@@ -0,0 +1,11 @@
+/**
```
thijstriemstra, line 1: some files are properly renamed, but some are not like clamp. The diff shows a removal of .js and new .ts file. I would like to preserve the history by using a proper rename/move.

build-config/fragments/common.js
```diff
@@ -33,6 +33,10 @@ module.exports = {
                         loader: 'babel-loader'
                     }
                 ]
+            },
+            {
+                test: /\.tsx?$/,
```
thijstriemstra, line 38: this will only detect .tsx files?

sundayz: the ? means the x is optional

marizuccara:
Hi @sundayz! You're performing a job of no small effort!

sundayz:
> 
> 
> Hi @sundayz! You're performing a job of no small effort! I've taken into consideration your request to can import ts files without specifying the extension. I've made some tries, and inserting the resolution of extensions in the webpack file, the problem seems to be solved. particular, I've wrote this piece of code in the file `build-config/fragments/commonj.js`:

Thank you!

src/util/clamp.ts
sundayz: I think I have to use git rebase to rewrite the history, will that break the pull request?

src/util/fetch.ts
```diff
@@ -172,13 +196,13 @@ export default function fetchFile(options) {
...
             // fire progress event when during load
-            instance.onProgress = e => {
+            instance.onProgress = (e: ProgressEvent) => {
                 instance.fireEvent('progress', e);
             };
 
             return new Response(
                 new ReadableStream(
-                    new ProgressHandler(instance, contentLength, response)
+                    new ProgressHandler(instance, contentLength!, response)
```
thijstriemstra, line 205: ??

sundayz: Asserts not null :D 

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

Code above already checks `if (contentLength === null)` but typescript didn't understand that and still had a warning without the !

thijstriemstra: hmm i sure hope there are not many more of these weird checks necessary or it'll be really hard to contribute in a fun and easy way in the future.. i didn't expect the complete codebase to be converted into typescript..

sundayz: I think by rewriting the code in that function the `!` wouldn't be necessary. I think setting `progressAvailable = false` instead of returning is preventing typescript from understanding that contentLength cannot be null.

But I didn't want to rewrite any code in this PR unless necessary, so I don't accidentally change how the code works.

sundayz:
Any ideas on what to do next? Keep the first PR small or convert more files.

thijstriemstra:
I would say rework this PR for util only (+ build changes) and make sure the files are properly git moved, e.g. `git mv foo.js foo.ts`.

sundayz:
How about I open a new PR with just build changes only, no source code changes. Then after that is merged we can start changing .js to .ts

That might make things easier

thijstriemstra:
> I would say rework this PR for util only

Like #434 📡 where testing started with util as well.

sundayz:
new pr here will fix git history issues: #2184

PR #2122, opened 2020-12-02
FIX: IOS webview implementation of FileReader doesn't support `addEventListener`

bennordgengo:
- This bug was discovered while developing an IOS app using Ionic.
- Currently, when loading a blob via `wavesurfer.loadBlob(blob)`, the IOS webview implementation of FileReader throws an error complaining:
  ```
  TypeError: n.addEventListener is not a function. (In 'n.addEventListener("progress",(function(e){return t.onProgress(e)}))', 'n.addEventListener' is undefined)
  ```
  In this error message, `n` is initialized as the `new FileReader`
- By replacing the calls to `addEventListener` with the equivalent `onprogress`/`onload`/`onerror`, this bug is resolved and the blob load correctly.

### DIFF
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -1336,13 +1336,9 @@ export default class WaveSurfer extends util.Observer {
     loadBlob(blob) {
         // Create file reader
         const reader = new FileReader();
-        reader.addEventListener('progress', e => this.onProgress(e));
-        reader.addEventListener('load', e =>
-            this.loadArrayBuffer(e.target.result)
-        );
-        reader.addEventListener('error', () =>
-            this.fireEvent('error', 'Error reading file')
-        );
+        reader.onprogress = e => this.onProgress(e);
+        reader.onload = e => this.loadArrayBuffer(e.target.result);
+        reader.onerror = () => this.fireEvent('error', 'Error reading file');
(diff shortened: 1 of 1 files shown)


thijstriemstra:
This is framework specific issue, e.g. https://github.com/mpetroff/pannellum/issues/284#issuecomment-260669954 I would try to use some sort of polyfill instead because it's not a bug in wavesurfer.js

bennordgengo:
Is there any risk in taking the fix?  I agree it's specific to the IOS webview, but this resolves the issue for everyone.

thijstriemstra:
>  I agree it's specific to the IOS webview,

Sounds to me it's specific to the Ionic framework. And if we start making exceptions for every framework it would get messy and unmaintainable..

bennordgengo:
This isn't using anything specific to Ionic, as far as I see.  That was simply for context.

thijstriemstra:
> This isn't using anything specific to Ionic, as far as I see.

That would change things. Can you supply link to those ios webview filereader docs?

bennordgengo:
Doing some more digging to make sure I'm not mistaken.  If this is really Ionic using their own implementation of FileReader, then I apologize.  Thanks for taking a look so far.

PR #2127, opened 2020-12-06
spectrogram plugin: Use ImageData to draw pixel-by-pixel

ilovecomputers:
Use `ImageData` to perform the pixel-by-pixel drawing operation instead of calling `fillStyle` and `fillRect` thousands of times since that translates to thousands of DOM operations within one frame.

Removed `const length` because it was unused in the method.

none

none

### Testing

Since there are no automated tests for the spectro plugin, I manually performed the following steps with and without the changes:

1. Set the device pixel ratio (1, 2, and 3) using Chrome dev tool's Device Mode
2. Load the spectro example page
3. Right click the spectro canvas to save image (need to turn off Device Mode to do this step)

I did this for both the mono and stereo test audio files. Here are the saved images for you to compare: 
[Compare Spectro.zip](https://github.com/katspaugh/wavesurfer.js/files/5649398/Compare.Spectro.zip)

The images are identical except for the slight downward shift for mono audio DPR 1 and 3. Stereo audio doesn't have these differences. Perhaps having to manually draw two pixel tall data points rather than drawing two pixel tall rectangles on the canvas causes a minor change in the scaling? I leave it to the maintainers if this is acceptable.

### Justification

I discovered [this perf bottleneck while working on my own spectro](https://github.com/ilovecomputers/spectrogram/commit/78c8bfb087499abbc8a355451d921932eb849f63). Thought I would contribute what I learned to wavesurfer. You can see this perf bottleneck by using Chrome dev tool's Performance tool when you click "Start profiling and reload page":

Screen Shot 2020-12-06 at 10 34 16 AM

Here's the result with the change:

Screen Shot 2020-12-06 at 10 34 28 AM

The self and total time of `drawSpectrogram` went down significantly because instead of mutating the DOM thousands of times (from  calling `fillStyle` and `fillRect`), the DOM is mutated once with `putImageData`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@ wavesurfer.js changelog
 ------------------
 
 - Add `relativeNormalization` option to maintain proportionality between waveforms when `splitChannels` and `normalize` are `true` (#2108)
+- Spectrogram plugin: Use `ImageData` to draw pixel-by-pixel (#2127)
 
 4.2.0 (20.10.2020)
 ------------------
--- a/src/plugin/spectrogram/index.js
+++ b/src/plugin/spectrogram/index.js
@@ -250,34 +250,34 @@ export default class SpectrogramPlugin {
 
     drawSpectrogram(frequenciesData, my) {
         const spectrCc = my.spectrCc;
-        const length = my.wavesurfer.backend.getDuration();
         const height = my.height;
+        const width = my.width;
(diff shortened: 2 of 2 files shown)


thijstriemstra (changes requested):
thanks @ilovecomputers, can you also add a changelog entry?

thijstriemstra (approved):

PR #2136, opened 2020-12-15
lint: disable and replace usage of var

thijstriemstra:
refs #2110

### DIFF
--- a/src/webaudio.js
+++ b/src/webaudio.js
@@ -305,7 +305,7 @@ export default class WebAudio extends util.Observer {
                 );
             }
             audio.autoplay = true;
-            let dest = this.ac.createMediaStreamDestination();
+            const dest = this.ac.createMediaStreamDestination();
             this.gainNode.disconnect();
             this.gainNode.connect(dest);
             audio.srcObject = dest.stream;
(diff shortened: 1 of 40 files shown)


thijstriemstra:
anyone wants to help and resolve these lint errors?

src/webaudio.js
```diff
@@ -305,7 +305,7 @@ export default class WebAudio extends util.Observer {
                 );
             }
             audio.autoplay = true;
-            var dest = this.ac.createMediaStreamDestination();
+            let dest = this.ac.createMediaStreamDestination();
```
marizuccara, line 308: Hi, why `dist` and `audio`  were replaced with let instead of const, since immutable?

thijstriemstra: because it was too much work to think about every var. I did with some but perhaps should've used let everywhere for consistency.

thijstriemstra: but for core code it makes sense to be implicit, will change this one.

marizuccara: > should've used let everywhere for consistency.

I think that where you're used const is logically correct, so can left as is

marizuccara (approved):

PR #2147, opened 2020-12-30
Draft: Tests for WebAudio Backend

sundayz:
It looks like the WebAudio backend has no tests? I made it so we now run the shared audio backend tests with WebAudio too. But now `WaveSurfer/MediaElementWebAudio` tests started failing...

### DIFF
--- a/spec/mediaelement-shared.js
+++ b/spec/mediaelement-shared.js
@@ -312,10 +312,18 @@ export function sharedErrorTests(backend) {
         }).toThrow(new Error('media parameter is not a valid media element'));
     });
 }
+
 function loadElement() {
     // set src
     audioElement.src = TestHelpers.EXAMPLE_FILE_PATH;
-    wavesurfer.load(audioElement);
+
+    // These tests should work with all backends?
+    // WebAudio requires the URL instead of the HTMLAudioElement to be passed in
+    if (wavesurfer.params.backend === 'WebAudio') {
+        wavesurfer.load(audioElement.src);
+    } else {
+        wavesurfer.load(audioElement);
+    }
 }
 
 /** Retrieve normalized waveform peaks, then load an audio resource giving peaks and setting preload attribute to 'none' **/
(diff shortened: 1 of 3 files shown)


spec/mediaelement-shared.js
```diff
@@ -312,10 +312,18 @@ export function sharedErrorTests(backend) {
...
 }
+
 function loadElement() {
     // set src
     audioElement.src = TestHelpers.EXAMPLE_FILE_PATH;
-    wavesurfer.load(audioElement);
+
+    // These tests should work with all backends?
+    // WebAudio requires the URL instead of the HTMLAudioElement to be passed in
+    if (wavesurfer.params.backend === 'WebAudio') {
```
sundayz, line 322: This should actually be reverted and the Wavesurfer.load() function needs to be changed. The documentation says it accepts `{string|HTMLMediaElement}`, so if you pass a HTMLMediaElement to the load function with backend 'WebAudio', it should automatically extract the src URL from the HTMLMediaElement

thijstriemstra: that's unfortunate. Sounds like a wavesurfer 5.0, or is this a bug.. when was the change introduced?

sundayz: Currently if you use WebAudio and pass a HTMLMediaElement, it thinks you gave it a string, the fix is something like:

```js
if (params.backend === 'WebAudio' && url instanceof HTMLMediaElement) {
    // Set URL to HTMLMediaElement.src
    url = url.src;
}
```

Won't break any user code, but will fix any broken code :P Alternatively we could throw an error or console.warn but it looks like it was intended that you can pass a HTMLMediaElement to any backend, I don't see any reason not to

sundayz: Oh the change in the tests was mine, because shared tests didn't work with WebAudio. It can be reverted if the load() method is fixed

thijstriemstra (changes requested):
CI build is failing.

sundayz:
Closing because https://github.com/katspaugh/wavesurfer.js/commit/658bd921be60a64192331ce8511bf1bce68cc4ec added initial WebAudio tests. Just need to add more, or make the shared backend tests work with WebAudio too

PR #2148, opened 2020-12-30
WaveSurfer load() should accept HTMLMediaElement when using WebAudio

sundayz:
The WaveSurfer.load function's documentation says that the url parameter can be a string or a HTMLMediaElement, but when using the `WebAudio` backend, the loadBuffer function will assume the URL is always a string, and send incorrect HTTP requests like this:

image

Technically this affects the external API, but instead of breaking existing code, it will fix currently broken code.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@ x.x.x (unreleased)
 ------------------
 
 - Regions plugin: Stop region dragging when mouse leaves canvas (#2158)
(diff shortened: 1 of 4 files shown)


thijstriemstra (changes requested):
Looks good. Please add a test(s) as well.

sundayz:
I had to make a new file for WebAudio tests, let me know if it looks OK, I don't know how you'd like to structure the tests

thijstriemstra (approved):
please add a changelog entry, then squash and merge, thanks.

Issue #1722, opened 2019-08-20
Multi Color Bars


alisajidcs:
Hy, I want to give color to each and bar saperately. Is there anyway to do so in current package?

If no please guide me how can I access canvas object so that I can override selected bars myself.

The scenario I am in right now is that I want to give colors based on emotions. Like if audio was angry for 1st 10 seconds then color should be red and if next 20 seconds is a happy voice then I want to give may be yellow or green color.

Thanks

thijstriemstra:
> Is there anyway to do so in current package?

not yet.

> how can I access canvas object 

@alisajidcs use the `wavesurfer.drawer.canvases` array to access the canvas(es).

if you come up with something we can re-use, happy to review any pull requests.

thijstriemstra:
cc @kubk

kubk:
Hi @alisajidcs , you can hook into `setFillStyles` to override bar colors separately: https://github.com/katspaugh/wavesurfer.js/blob/9f1ed9064ea86e736cc9d4051ffe0a2bfb4d74b3/src/drawer.canvasentry.js#L141

This method is called for each bar. Let me know if you need more details.

PR #2150, opened 2021-01-05
Update to support multi channel colors for bars

adarshpastakia:
Added ability to draw multi channel bars with different colors similar to regular line wave

None

None

refs #1722


adarshpastakia:
@thijstriemstra this update does not address the linked issue, this update is for using splitChannel colors when rendering as bars. the issue however is for coloring sections of the wave in different colors

marizuccara (commented):
Can you add also an example in example dir?

thijstriemstra:
> Can you add also an example in example dir?

And a changelog entry?

PR #2155, opened 2021-01-16
examples: add missing sample.m3u

### LINKED ISSUE #2112: M3U Playlist Plugin example broken
Just noticed the [M3U Playlist Plugin example](https://wavesurfer-js.org/plugins/playlist.html) is broken, the M3U file is no longer hosted at the location it points to.

entonbiba:
fixes #2112


thijstriemstra (approved):
thanks.

PR #2158, opened 2021-01-17
regions plugin: stop dragging region when mouse exits wavesurfer div

jbeuckm:
It is difficult to set a region border close to the start or end of a wavesurfer. The intention here is that when the mouse leaves the wavesurfer, the region being adjusted stays at the most recent mousemove position instead of following the mouse after mouseup outside the wavesurfer.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,13 +1,10 @@
 wavesurfer.js changelog
 =======================
 
-
-4.4.1 (17.01.2021)
+x.x.x (unreleased)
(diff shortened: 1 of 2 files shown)


thijstriemstra (approved):
Thanks!

marizuccara (approved):
Thanks! Can you add a changelog?

jbeuckm:
> Thanks! Can you add a changelog?

a4cd4e6

PR #2164, opened 2021-01-23
ensure webaudio is removed from ram/cpu usage

### LINKED ISSUE #1940: Every call to wavesurferjs keeps on increasing RAM memory, without releasing  it on destroy.
I have came across a issue with `wavesurferjs` it does not release memory previously held upon destroy.

With high duration audio file, if we open 10 times it will hang browser itself (then we need to either hard refresh or close browser itself to continue).

Steps to reproduce on linux:

1.open terminal type $htop
2.keep on opening audio with destroy on closing.
3.memory will keep on adding

Though i'm not a good programmer, but i have found a exact solution with `canvas`  and `audio api`

Here is that solution: https://stackoverflow.com/questions/53241345/web-audio-api-memory-leak?noredirect=1&lq=1

i request @katspaugh , please implement it as soon as possible.

entonbiba:
fixes #1940

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,7 @@ wavesurfer.js changelog
 x.x.x (unreleased)
 ------------------
 
-- Fixed memory leak with `destroy()` in `WebAudio` backend (#1778)
+- Fixed memory leak with `destroy()` in `WebAudio` backend (#1940)
 - Regions plugin: Stop region dragging when mouse leaves canvas (#2158)
 - Fixed `WaveSurfer.load(url)` not working when passing a HTMLMediaElement as the url parameter, with the WebAudio backend. 
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ wavesurfer.js changelog
 x.x.x (unreleased)
 ------------------
 
+- Fixed memory leak with `destroy()` in `WebAudio` backend (#1778)
 - Regions plugin: Stop region dragging when mouse leaves canvas (#2158)
 - Fixed `WaveSurfer.load(url)` not working when passing a HTMLMediaElement as the url parameter, with the WebAudio backend. 
 
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -1732,7 +1732,7 @@ export default class WaveSurfer extends util.Observer {
             );
         }
         if (this.backend) {
-            this.backend.destroy();
+            this.backend.buffer = null;
         }
         if (this.drawer) {
             this.drawer.destroy();
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -1734,9 +1734,7 @@ export default class WaveSurfer extends util.Observer {
         if (this.backend) {
             this.backend.destroy();
             // clears memory usage
-            if (this.backend !== null) {
-                this.backend = null;
-            }
+            this.backend = null;
         }
         if (this.drawer) {
             this.drawer.destroy();
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -1733,7 +1733,10 @@ export default class WaveSurfer extends util.Observer {
         }
         if (this.backend) {
             this.backend.destroy();
-            this.backend = null;
+            // clears memory usage
+            if (this.backend !== null) {
+                this.backend = null;
+            }
         }
         if (this.drawer) {
             this.drawer.destroy();
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -1732,6 +1732,7 @@ export default class WaveSurfer extends util.Observer {
             );
         }
         if (this.backend) {
+            this.backend.destroy();
             this.backend.buffer = null;
         }
         if (this.drawer) {
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -1733,7 +1733,7 @@ export default class WaveSurfer extends util.Observer {
         }
         if (this.backend) {
             this.backend.destroy();
-            this.backend.buffer = null;
+            this.backend = null;
         }
         if (this.drawer) {
             this.drawer.destroy();
--- a/spec/wavesurfer.spec.js
+++ b/spec/wavesurfer.spec.js
@@ -356,7 +356,7 @@ describe('WaveSurfer/playback:', function() {
             done();
         });
         wavesurfer.destroy();
-        
+
         expect(wavesurfer.backend).toBeNull();
     });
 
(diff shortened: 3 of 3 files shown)


entonbiba:
@katspaugh @thijstriemstra `backend = null` will happen as long as `backend` is available

``` javascript
if (this.backend) {
  this.backend.destroy();
  this.backend.buffer = null;
}
```

sundayz:
Your comment on #1940 says we should set this.backend to null but this commit sets this.backend.buffer to null, so which is it :)

Your research into the issue is great.  if we know loading new audio allows the memory to be garbage collected, we can follow that code path and see what it is doing that we are not doing in the destroy() function.

marizuccara:
Hi, also the audio buffer is already set to null in the `WebAudio` backend (which is the only backend that uses the `AudioBuffer`).

I hope that setting the backend to null will resolve all your problems! :)

sundayz:
Here is something interesting I found that may be a piece of the puzzle...

WebAudio createSource sets a reference of the buffer on the source.
https://github.com/katspaugh/wavesurfer.js/blob/658bd921be60a64192331ce8511bf1bce68cc4ec/src/webaudio.js#L580


WebAudio disconnects the source, but never sets it to null.
https://github.com/katspaugh/wavesurfer.js/blob/658bd921be60a64192331ce8511bf1bce68cc4ec/src/webaudio.js#L509

So if when we call destroy() a reference to the backend still exists, then a reference to the buffer will still also exist. Might be part of the problem. And could explain why calling load() again as entonbiba has shown frees the memory, because a new source is set and the old one is collected.

marizuccara:
I found this interesting response on stackoverflow:

marizuccara:
I found this interesting response on stackoverflow:
https://stackoverflow.com/questions/55207924/does-the-web-audio-api-clear-out-a-source-node-after-it-finishes-playing

On destroy we can call also stop() on the source node, if it's playing (in WebAudio). On `onendend` event, we can disconnect it,  sets to null the buffer, and then the node, as you suggest

entonbiba:
@sundayz lol yes `backend = null` is correct, updated

thijstriemstra (changes requested):
thanks. can you also expand the existing test for `destroy` and add an assertion that `backend` is `null`. Also add a changelog entry please (with a reference to the ticket this pr is closing).

CHANGES.md
```diff
@@ -4,6 +4,7 @@ wavesurfer.js changelog
 x.x.x (unreleased)
 ------------------
 
+- Fixed memory leak with `destroy()` in `WebAudio` backend (#1778)
```
thijstriemstra, line 7: this should be #1940

entonbiba: hmm, updated to correct pr

entonbiba:
@thijstriemstra added null check,

@sundayz yes the observation about the `load()` is correct as its replacing backend with new. In the destroy case it would become null for this to work since even if we did .close or .disconnect it would still keep it as suspended taking up memory usage in webaudio.

thijstriemstra:
@entonbiba I meant more somewhere along these lines: https://github.com/katspaugh/wavesurfer.js/blob/entonbiba-patch-webaudio-usage-fix/spec/wavesurfer.spec.js#L351

entonbiba:
@thijstriemstra I'll add `expect(wavesurfer.backend).toBeNull();` after the `destroy();` like below:

``` javascript
it('destroy', function(done) {
        manualDestroy = true;

        wavesurfer.once('destroy', function() {
            TestHelpers.removeElement(element);
            done();
        });
        wavesurfer.destroy();

        expect(wavesurfer.backend).toBeNull();
    });
```

src/wavesurfer.js
```diff
@@ -1733,6 +1733,10 @@ export default class WaveSurfer extends util.Observer {
         }
         if (this.backend) {
             this.backend.destroy();
+            // clears memory usage
+            if (this.backend !== null) {
```
thijstriemstra, line 1737: this null check can be removed, it's already doing that on line 1734

thijstriemstra (approved):

katspaugh (approved):
Great work!

sundayz:
I haven't tried the new branch but in my project when I set backend = null, a little bit more memory gets collected but the 50MB arraybuffer still remains there.

thijstriemstra:
Feel free to re-open ticket closed by this pr: #1940

entonbiba:
@sundayz do you have the code you are using the create the instances ?
Just tried it again with `backend = null` and the results are working:

- create
- load
- backend = null
image

sundayz:
The file is 2000 lines long and it uses Vue so we should test in a code pen :D I will try that tomorrow. I just have a feeling that if the issue was as simple as `backend = null` then it would have been fixed a long time ago.

If your PR fixes the leak for most people then it's still a very good thing. I will try to reproduce the issue without Vue later.

entonbiba:
@sundayz  if it's 2000 lines of vue then it's likely rendering a whole lot more than the audio file that's taking up memory.

The best test is to the library without any additional things. I just do `create` `load` `destroy (backend=null) pr` and repeat that multiple times with multiple loads at once to see the differences of usage.

ping me when the sample code is available I'll test it as well :)

thijstriemstra:
Unfortunately this PR introduced a new issue with timing issues around `backend = null`. In unit tests for a wavesurfer related project I started seeing this error:

```
  TypeError: Cannot read property 'getCurrentTime' of null
      at WaveSurfer.getCurrentTime (node_modules/wavesurfer.js/dist/wavesurfer.js:4199:27)
      at d.value (dist/commons.js:8:71353)
      at d.value (dist/commons.js:8:69669)
      at d.value (dist/commons.js:8:74618)
      at node_modules/wavesurfer.js/dist/wavesurfer.js:3214:12
      at <Jasmine>
      at WaveSurfer.fireEvent (node_modules/wavesurfer.js/dist/wavesurfer.js:3213:28)
      at node_modules/wavesurfer.js/dist/wavesurfer.js:4133:16
      at node_modules/wavesurfer.js/dist/wavesurfer.js:3214:12
      at <Jasmine>
```

Commenting out the `backend = null` change from this PR makes the test pass again.

Basically any code in wavesurfer that relies on the `backend` property, e.g. `wavesurfer.getCurrentTime()` will now break if `destroy()` is called. This is not necessarily a bad thing, the instance was destroyed afterall, but should wavesurfer fix this (e.g. add a backend !== null check everywhere) or is it up to the user to check for an proper, existing `wavesurfer.backend` property?

thijstriemstra:
> but should wavesurfer fix this (e.g. add a backend !== null check everywhere) or is it up to the user to check for an proper, existing wavesurfer.backend property?

Turns out it was an issue with my tests and this change brought it to light, so no worries, sorry for the noise.

entonbiba:
sounds good

PR #2167, opened 2021-01-24
Split channels: overlay param now properly displays a single canvas

johnpaulmedina:
#2161 

When setting overlay: true on split channels the height of the waveform becomes equal to a single channel height but does not overlay the second channel. 

The issue is that the offsetY value is set based on height multiplied by the drawIndex or default to zero. In this case we should check for the splitChannelOptions.overlay and set the offsetY to zero regardless of the drawIndex.

PR #2173, opened 2021-01-25
build: remove publicPath from webpack dev config

johnpaulmedina:
Added rootDir const to webpack dev config. Should resolve issue with build we discussed on #2167

### DIFF
--- a/build-config/fragments/dev.js
+++ b/build-config/fragments/dev.js
@@ -7,14 +7,13 @@ module.exports = {
     mode: 'development',
     devtool: 'eval-source-map',
     output: {
-        publicPath: 'localhost:8080/dist/'
+        publicPath: 'localhost:8080/'
     },
     devServer: {
         static: [
             {
-                directory: path.join(rootDir, 'dist'),
+                directory: rootDir,
                 staticOptions: {},
-                publicPath: '/dist/',
                 serveIndex: true,
                 watch: {
                     ignored: [
--- a/build-config/fragments/dev.js
+++ b/build-config/fragments/dev.js
@@ -1,6 +1,7 @@
 /* eslint-env node */
 
 const path = require('path');
+const rootDir = path.resolve(__dirname, '..', '..');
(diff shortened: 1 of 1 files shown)


thijstriemstra (changes requested):
thanks @johnpaulmedina. Previously localhost:8080 showed a file listing of the rootdir and after some tweaking this is what most resembles prev situation:

```diff
diff --git a/build-config/fragments/dev.js b/build-config/fragments/dev.js
index 57bcc0e..c432996 100644
--- a/build-config/fragments/dev.js
+++ b/build-config/fragments/dev.js
@@ -1,6 +1,7 @@
 /* eslint-env node */
 
 const path = require('path');
+const rootDir = path.resolve(__dirname, '..', '..');
 
 module.exports = {
     mode: 'development',
@@ -11,9 +12,8 @@ module.exports = {
     devServer: {
         static: [
             {
-                directory: path.resolve(__dirname, '..', '..'),
+                directory: rootDir,
                 staticOptions: {},
-                publicPath: '/dist/',
                 serveIndex: true,
                 watch: {
                     ignored: [
```

Otherwise browsing to localhost:8080 will result in a GET error. Can you make test and make these changes?

johnpaulmedina (commented):
Got it. I tested it and it now serves the root dir as requested. So the localhost can access either directory now as previously.

build-config/fragments/dev.js
```diff
@@ -1,19 +1,19 @@
...
 
 const path = require('path');
+const rootDir = path.resolve(__dirname, '..', '..');
 
 module.exports = {
     mode: 'development',
     devtool: 'eval-source-map',
     output: {
-        publicPath: 'localhost:8080/dist/'
+        publicPath: 'localhost:8080/'
```
thijstriemstra, line 10: not sure if this var is actually used?

thijstriemstra (approved):
thanks!

Issue #2137, opened 2020-12-16
I want to use a 2-channel waveform and save the JSON data generated by the waveform; however, after using exportpcm, the result can only be null


guisha:
var wavesurfer = WaveSurfer.create({
			container: '#waveform',
			progressColor: 'red',
			waveColor: "#666666",
			height: 65,
			fillParent:true,
			forceDecode:true,
			partialRender:true,
			pixelRatio:1,
			splitChannels: true,
			barWidth: 0,
			
		
		});
		 wavesurfer.load("music/work.mp3"); 
		wavesurfer.on("ready", function() {
			
			console.log(wavesurfer.exportPCM('1024','10000',true,0));
		})
       result：
         Promise {<resolved>: "[null,null,null,null,null,null]"}；
       How to solve this problem！

sundayz:
#1954 

possibly related #2091 

Does the waveform display, or do you see a single line?

Also, in English please if possible.

guisha:
waveform display

guisha:
I want to use a 2-channel waveform and save the JSON data generated by the waveform; however, after using exportpcm, the result can only be null

thijstriemstra:
@guisha can you rename the ticket title in english?

guisha:
No one?

entonbiba:
continue on this ticket #1954

Issue #1896, opened 2020-03-30
exportPCM Wrong method return type


azimbialiev:
I'm actually using wavesurfer.js@3.3.1 with Typescript 3.8
VS Code suggest me that wavesurfer.exportPCM() is not an async function

`WaveSurfer.exportPCM(length: number, accuracy: number, noWindow?: boolean | undefined, start?: number | undefined): string`

But the official documentation says 

`exportPCM(length: number, accuracy: number, noWindow: boolean, start: number, end: number): Promise`

thijstriemstra:
> But the official documentation says

Where did you find this, can you link it?

aburai:
since 3.3.0  `wavesurfer.exportPCM` now accepts an optional `end` argument and returns
  a Promise (#1728)

the doc is correct, but wavesurfer has no typings file so i think vscode do not understand the jsdoc

    /**
     * Exports PCM data into a JSON array and opens in a new window.
     *
     * @param {number} length=1024 The scale in which to export the peaks
     * @param {number} accuracy=10000
     * @param {?boolean} noWindow Set to true to disable opening a new
     * window with the JSON
     * @param {number} start Start index
     * @param {number} end End index
     * @return {Promise} Promise that resolves with array of peaks
     */

thijstriemstra:
> since 3.3.0 wavesurfer.exportPCM now [...] returns a Promise

Closing as fixed.

Issue #1383, opened 2018-05-31
how to export pcm 


nongcunshaoye:
exportPCM ()   is not work in version 2.0.5,    any other method replace ?

thijstriemstra:
@nongcunshaoye it works: https://github.com/katspaugh/wavesurfer.js/blob/master/spec/wavesurfer.spec.js#L269

thijstriemstra:
no feedback, closing.

PR #2175, opened 2021-01-27
Fix exportPCM noWindow parameter

### LINKED ISSUE #1954: Unable to display 2 channel waveform when externally loading peaks
Amazing work what you have done! 

Can you please elaborate, if what I am trying to achieve is achievable at all at the moment? The description is below.

* https://wavesurfer-js.org/example/split-channels/index.html
This works like charmed! And everything else is okey with small files. 

But I'm dealing with 100mb+ files and for that I am using audiowaveform to generate peaks. I'm loading it in with:

```
this.wave.load('http://localhost:4200/assets/' + mp3, peaks.data);
```
and initializing wavesurfer:
```
  private initWaveSurfer() {
    this.wave = WaveSurfer.create({
      container: '#waveform',
      backend: 'MediaElementWebAudio',
      pixelRatio: 1,
      height: 80,
      splitChannels: true,
      normalize: true,
      minimap: true,
```
Waveform is displayed nicely, media plays nicely but there's only one merged wave. How can I still have two waveforms separately for each channel as in the example? Is this possible?

* [root@hop build]# ./audiowaveform -i 2_channel.mp3 -b 8 -o 2_channel.json --split-channels
* [root@hop build]# python scale.py 2_channel.json

Unfortunately I cant execute the following command:
this.wavesurfer.exportPCM();
Promise {<resolved>: "[null,null]"}
__proto__: Promise
[[PromiseStatus]]: "resolved"
[[PromiseValue]]: "[null,null]"

Results in exception:
unpkg.com/wavesurfer.js@3.3.3/dist/wavesurfer.js:5240 Not allowed to navigate top frame to data URL: data:application/json;charset=utf-8,%5Bnull%2Cnull%5D

What I wanted to do is to export PCM to see how the different channels are described in the json (as in the same split channel example thats on your webpage). Maybe I just have to write a wrapper to audiowaveform output to format it correctly.

Can you please give advice and maybe also post an example of this.wavesurfer.exportPCM(); when there is a 2 channel mp3 in play.

Thank You

entonbiba:
- [x] fixes 2 channel split peaks data output
- [x] fixes 1 channel peaks data output
- [x] fixes new window to now create a valid json blob file
- [x] fix promise call to return normal json if `noWindow` is set
- [x] promise now returns a blob if `noWindow` is not set

closes #1954

refs #2137 #1728 #1383 #1896


entonbiba:
will do some more tests for split channel to see how that behaves with multiple type of wav/mp3 files of different channel lengths first and update pr as needed.

src/wavesurfer.js
```diff
@@ -1624,17 +1624,22 @@ export default class WaveSurfer extends util.Observer {
...
             resolve(json);
-        });
+        }).then(response => {
+            if (!noWindow) {
+                window.open(URL.createObjectURL(
+                    new Blob([JSON.stringify(json)], {
+                        type: 'application/json;charset=utf-8'
+                    })
+                ));
+                return;
```
entonbiba: @katspaugh I'm thinking of returning the actual blob here or maybe the same json of the peaks instead? What you think or should I add the blob export as an option?

entonbiba (commented):
promise will now return a blob or a json if `noWindow` option is provided.

thijstriemstra (changes requested):
thanks. can you add some more tests for all use cases? The current `expect(pcmData).toBeNonEmptyString()` is too vague.

entonbiba:
yes, I'll add a promise check on that

src/wavesurfer.js
```diff
@@ -1624,16 +1624,22 @@ export default class WaveSurfer extends util.Observer {
...
-                        encodeURIComponent(json)
+                let blobJSON = URL.createObjectURL(
+                    new Blob([JSON.stringify(response)], {
+                        type: 'application/json;charset=utf-8'
+                    })
                 );
+                window.open(blobJSON);
+                return blobJSON;
+            } else {
+                return response;
```
katspaugh, line 1641: We could do an early return here, something like this:
```
return new Promise((resolve, reject) => {
  if (noWindow) {
    return arr;
  }
  const blobJSON = URL.createObjectURL(...);
  window.open(blobJSON);
  return blobJSON;  
});
  ```

Comment (entonbiba) [reply src/wavesurfer.js:1641]: yes we can do this as well,

**Review by thijstriemstra — CHANGES_REQUESTED** [review]
Build is failing and no unit tests were added.

**Comment by entonbiba** [comment]
added `expect(pcmData).to.be.a('promise');` but it fails the build test, should we just add a check to see if its not null ?

File: spec/wavesurfer.spec.js [hunk 3]
```diff
@@ -321,9 +321,9 @@ describe('WaveSurfer/playback:', function() {
     });
 
     /** @test {WaveSurfer#exportPCM} */
-    it('return Promise with PCM data formatted using JSON.stringify', function(done) {
+    it('return Promise with PCM data', function(done) {
         wavesurfer.exportPCM().then(pcmData => {
-            expect(pcmData).toBeNonEmptyString();
+            expect(pcmData).to.be.a('promise');
```
katspaugh, line 326: Isn't it supposed to be an array of numbers?

entonbiba: it's returned as a promise so it can be an array or a blob, I could add two checks one with the blob and the other just array.

image

thijstriemstra: sounds good.

thijstriemstra: please also add a test for case where wavesurfer.load wasn't called yet (and peaks are empty).

entonbiba:
Seems the test is failing since there's no actual blob data without loading audio during the tests. Do you think I should change the `toBeTruthy()` to simply check if it's empty?

`expect(pcmData instanceof Blob).toBeTruthy();`

`Error: Expected false to be truthy.`

https://github.com/katspaugh/wavesurfer.js/pull/2175/files#diff-5f387fea6eccc282d6f1763d525fad47b1f5f02e906897b40eb5e9af412018da

thijstriemstra:
> Seems the test is failing since there's no actual blob data without loading audio during the tests.

Yeah, you should check out the other tests and load some audio first @entonbiba

thijstriemstra:
>  promise now returns a blob if noWindow is not set

@entonbiba I made some changes in the branch, and I wonder why you initially changed it to resolve the promise with a blob and not simply with an array?

spec/wavesurfer.spec.js
thijstriemstra: > Isn't it supposed to be an array of numbers?

Agreed. It should always return an array IMO (and not a blob).

thijstriemstra:
ping @entonbiba

entonbiba:
we can leave it as array in this case, I was going to add the option when it opens in new window it converts it to either pure json blob or pcm blob file.

PR #2151, opened 2021-01-05
Fix seeking issue for WebAudio backends

todoroff:
Revert the changes from #2045 but only for `WebAudio` backends.

Fixes #2149

PR #2176, opened 2021-01-27
allow audio to be played if device is in silent mode

### LINKED ISSUE #1701: Playing with no audio on iOS 12.4
## The issue:
The waveform is rendered, and it seems like it's playing (you see the ticker moving), but there is no audio.
I've tried the audio context workaround suggested in #1415, but to no avail.
Thanks.

### Wavesurfer.js version(s):
latest (whatever https://wavesurfer-js.org uses)

iOS Safari 12.4

Just navigate to https://wavesurfer-js.org/ and click play

### LINKED ISSUE #1864: Playing without audio on iOS 13.3.1
### Wavesurfer.js version(s):
3.3.1 (2020-01-14)

Safari on iOS 13.3.1

```
    container: '#waveform',
    waveColor: '#ccc',
    progressColor: '#000000',
    height: 40
});
wavesurfer.load('sample.mp3');
```

Trying to play the audio does not create any errors and visually all seems to work. However, no sound can be heard. The same is true for examples that are supplied on the official Waversurfer website (https://wavesurfer-js.org/).

entonbiba:
allows devices such as ios and android to play audio while in silent mode for both webaudio and mediaelement.

test:
https://www.entonbiba.com/lab/ios2.html

closes #1864 
closes #1701


bishalbaaniya123 (approved):

thijstriemstra (changes requested):
- this element is never removed
- this workaround is only needed for safari right?
- it creates a new audio element on every play call

entonbiba:
Yes primarily for ios but also any devices that have the silent mode turned on including androids and windows phones. It should now remove the element after it plays.

thijstriemstra (commented):
I don't like adding this idea/hack at all.

entonbiba:
It should be fine to include this since the user is initiating the web audio. By default only web audio is disabled while actual html5 audio elements are allowed to continue playing in silent mode. Silent mode cannot be checked via javascript apis at the moment.

sundayz:
I'm guessing this is only needed for WebAudio right? Normal MediaElement streams behave as expected?

By the way there's places in the wavesurfer code where `play()` is called to work around browser/library bugs, see for example:
https://github.com/katspaugh/wavesurfer.js/pull/2151

So this code may be getting called more than you expect.

I think we should look into this more before deciding on this fix because it is a pretty weird hack fix. Ideas for example: is it enough to do this just once? Can it be done on page load once, or does it need to be done for every audio source?

entonbiba:
@sundayz no this fixes is for all backend options webaudio and medialement on ios and android devices.

Yes I was thinking of making it globally load once, I'll add it above this code in here unless there's a better place to put it ?
https://github.com/katspaugh/wavesurfer.js/blob/45a4732599c1ec42a82a0d2811d2f544e9cf02f1/src/html-init.js#L221

sundayz:
I'm doing some research.

When I go to this page on my iPhone with the mute switch on, I can hear audio: https://www.w3schools.com/html/tryit.asp?filename=tryhtml5_audio_all

Then when I visit the wavesurfer home page, which uses the MediaElement backend, I can't hear any audio.

Youtube and Soundcloud play too. Both the simple w3schools demo and Soundcloud correctly appear in the iOS media center and lock screen which lets you play/pause audio.

So it seems like Wavesurfer is playing in the wrong audio "channel", im not sure what Android and iOS call them. But the mute switch in iPhone should only mute the RINGER while it should allow media to play unmuted. The question is how do we make wavesurfer be considered a "media"? At the very least the MediaElement backend should work because you can see the `<audio>` tag is considered media, not a ringer sound.

I haven't tested with Android but I assume it has the same problem?

entonbiba:
In order for the mediaelement backend to work while it's silent mode the actual <audio> tag needs to be initiated by the audio element `.play()` function, not sure if this is happening anywhere.

This fix takes care of both media element and web audio to ensure they are considered initiated by the user on play.

thijstriemstra (commented):
> In order for the mediaelement backend to work while it's silent mode 

Can someone define 'silent mode'? From quote below it looks like the mute switch on an iPhone? Is there any proper documentation on the behavior of this mute switch for iOS and WebAudio?

> But the mute switch in iPhone should only mute the RINGER while it should allow media to play unmuted. The question is how do we make wavesurfer be considered a "media"? At the very least the MediaElement backend should work because you can see the <audio> tag is considered media, not a ringer sound.

entonbiba:
@thijstriemstra silent mode requires the user interact with the media element of initiating it before it can play. Since there is not actual element displays using the default browser audio element that the user can click, it will have to be initiated the way the patch does it. Below are the official documentation from apple.

https://developer.apple.com/design/human-interface-guidelines/ios/user-interaction/audio/

https://support.apple.com/en-ca/HT208353#:~:text=The%20Ring%2FSilent%20switch%20is,play%20through%20your%20iPhone%20speaker.

tyress:
Hi guys, chiming in to mention that a few people have been doing similar workarounds as in https://stackoverflow.com/a/46839941/1685167 for a while now, but I cannot find any documented reason for this. However the author of that SO answer chalks it up to the "janky" Safari browser in their [repo](https://github.com/swevans/unmute)

sundayz:
> 
> 
> @thijstriemstra silent mode requires the user interact with the media element of initiating it before it can play. Since there is not actual element displays using the default browser audio element that the user can click, it will have to be initiated the way the patch does it. Below are the official documentation from apple.

I think I understand, so if the playback doesnt start from the actual <audio> controls generated by the browser it's not considered "media"?

IDK if this fix could cause performance issues, but it it does, it should probably be controlled through a conditional flag like `iOSsilentModeFix: true`, or maybe even a webpack conditional so you can remove it from the build.

sundayz:
This stack overflow post shows this process being done once per audio element: https://stackoverflow.com/questions/21122418/ios-webaudio-only-works-on-headphones/46839941#46839941

So we probably only need to do it the first time per WaveSurfer instance, the first time it plays. I'm a lot more OK with this if it only needs to happen once, rather than every time play() is called.

entonbiba:
we can do it on the first load, however by doing it on every audio load we ensure this complies with any requirements that state the user has to initialize the media first as the silent physical button on the device itself has not changed and there's no way of knowing if that changes currently on androids and iphones.

thijstriemstra:
> I'm a lot more OK with this if it only needs to happen once, rather than every time play() is called.

Agreed.

> the silent physical button on the device itself has not changed and there's no way of knowing if that changes currently on androids and iphones.

I see.

thijstriemstra:
> This fix takes care of both media element and web audio to ensure they are considered initiated by the user on play.

I did some tests with iOS 14.6 and didn't encounter any issues with audio being silenced when enabling the when using the MediaElement backend. The issue is present when using the WebAudio backend. I tried the fix from 

> I'm guessing this is only needed for WebAudio right? Normal MediaElement streams behave as expected?

It looks like it to me. Can you confirm @entonbiba?

thijstriemstra (approved):

thijstriemstra:
I ended up adding the fix behind a `ignoreSilenceMode` boolean option that is disabled by default.

sundayz (approved):

sundayz:
👍 It's an inconsistency that Apple needs to fix, seems good to me to have it behind a boolean flag

jakemumu:
not sure how I missed this but very happy it was included! So much better than managing a hack ourselves!

katche70:
Currently it starts with the ignoreSilenceMode: true for 5 seconds, then it is inaudible again. Shortly pressed pause and play again, then it runs again. Is this bug known ?

thijstriemstra:
@katche70 without any extra info (ios version, safari version etc) it's impossible to know what you mean.

katche70:
Sorry, iOS 16.1.1 iPhone 14 pro

jroebu14:
> Currently it starts with the ignoreSilenceMode: true for 5 seconds, then it is inaudible again. Shortly pressed pause and play again, then it runs again. Is this bug known ?

I started using wavesurfer.js on a project recently and have occasionally experienced the same issue as quoted above. It doesn't happen 100% of the time. I'm wondering if anyone else has experienced it.

I'm using wavesurfer.js v6.4.0

iPhone 13 Pro
iOS 16.2
Chrome 105.0.5195.147

Many thanks for creating wavesurfer.js btw it's a really amazing piece of work!

PR #2184, opened 2021-02-07
WIP: Typescript Port

sundayz:
Sets up basic typescript build system. Fixes issues old PR had with git history.

The end result should be that the project correctly builds even with Javascript files and we can slowly migrate towards typescript where needed. 

Typescript should output the latest javascript our project can use (es2017?) because Webpack will turn that into ES5 as required.

Feel free to suggest changes to the tsconfig file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@
 
 package-lock.json
 .package.json.swp
+
+# Typescript build directory
+/build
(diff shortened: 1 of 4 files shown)


thijstriemstra:
The typings in the DefinitelyTyped project look very complete as well with this pull request: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52726

@sundayz can you help review that pull request, they're asking for help but I'm not a typescript dev.

PR #2186, opened 2021-02-12
Adding deinterleaving to scale_json.py script in FAQ to support multichannel peak data generated by audiowaveform

kickermeister:
BBC's audiowaveform is generating a interleaved version when using the --split-channels flag, see https://github.com/bbc/audiowaveform/blob/master/doc/DataFormat.md . wavesurfer.js, however, seems to expect an separate array for each audio channel for the externally generated peak data. This updated script should generate appropriate peak data for multichannel files (8 channels is currently the maximum supported number by BBC's audiowaveform).

None

None

### Todos/Notes:
Tested with 1 - 8 channels for different backends and it seemed to work as expected.

#356, #1954


thijstriemstra (changes requested):
great! can you add some docstrings and/or comments to the script?

kickermeister:
Glad you like it. I've added some comments and improved the script a little. Did also add another code example using the --split-channels flag. Hope it's good to go now?

PR #2194, opened 2021-02-23
Fix annotations example

osheroff:
Hi guys.  Went to see if annotations was the solution to my problem. 

It's not, but this fixes the public facing example.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,12 +1,16 @@
 wavesurfer.js changelog
 =======================
 
+x.x.x (unreleased)
+------------------
+- Markers plugin: add new plugin that allows for timeline markers (#2196)
+
 4.5.0 (14.02.2021)
 ------------------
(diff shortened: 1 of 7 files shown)


osheroff:
this also fixes the regions plugin example.

thijstriemstra:
> It's not, but this fixes the public facing example.

What was specifically the issue? The build for this PR is now failing.

osheroff:
> What was specifically the issue?

check out the examples for region and annotations (and anything that includes trivia.js); they're failing under chrome with "wavesurfer" is already defined.

> The build for this PR is now failing.

yeah, seems unlikely that this PR caused it but I'll see.

osheroff:
https://github.com/babel/babel-eslint/issues/487 is the same issue; class field declarations are throing a no-undef from eslint.

I've spent 10 minutes now and that's my yak shaving budget for the day.  Yours if you want it :)

thijstriemstra:
fixed in #2196

Issue #634, opened 2016-02-09
Vertical Track


elBarkey:
Hi, is it possible to display the waveform in vertical layout ?
i've tried using css transofrm 90deg to parrent element, and it works.
but that just rotating the parrent element to 90 degrees, that's all. offcourse mouse event, seeking etc on actual waveform  doesn't work by just doing "css hacks".
is there any workaround on this?

katspaugh:
Hi! Interesting question!

I think this function needs to be modified: https://github.com/katspaugh/wavesurfer.js/blob/master/src/drawer.js#L41-L45

It assumes the horizontal axis. Try changing it to take `clientY` instead of `clientX`, `bbox.top` instead of `bbox.left` etc.

Also make sure you set the right rotation origin in CSS.

elBarkey:
Thanks @katspaugh !
I followed your direction, it now works as as expected.
The only problem remains is, the rotated wrapper draws the width depends on its parent width.
can you point me to the width value origins ? so i can take its parent height instead of the width.

katspaugh:
The container's width is used in these places: https://github.com/katspaugh/wavesurfer.js/blob/master/src/drawer.js#L96
https://github.com/katspaugh/wavesurfer.js/blob/master/src/drawer.js#L102
https://github.com/katspaugh/wavesurfer.js/blob/master/src/drawer.js#L105
https://github.com/katspaugh/wavesurfer.js/blob/master/src/drawer.js#L130
https://github.com/katspaugh/wavesurfer.js/blob/master/src/drawer.js#L168

I'm not sure if *all* of these places need to be changed to `height`, though.

elBarkey:
yep, you're right.
that way, the main  drawer core will be messed up.
as a solution i do this to get the the wrapper displayed vertically:

```html
<style>
    #v-container{
        position: fixed;
        top: 0px;
        left: 0px;
        height: 100%;
    }
    #w-container{
        transform: translateX(100%) rotate(90deg);
        transform-origin: top left;
    }
</style>
<div id="v-container">
    <div id="w-container">
        <div id="wave"></div>
    </div>
</div>
```

and change the wrapper parent height on ready event:

```javascript
wavesurfer.on('ready',function(){
    $('#w-container').width($('#v-container').height());
});
```

anyway, thanks for the kind response ;)

katspaugh:
Cool, nice hack!

987123879113:
4 years later... (sorry for the necro)

[...]

PR #2195, opened 2021-02-24
Add first-class support for vertical orientation

valrus:

This PR implements support for rendering the waveform in a vertical rather than horizontal orientation. To do this with as little disruption to the existing code as possible, I use two main approaches:
1. A utility object, `Orientation`, has two subclasses. `HorizontalOrientation` is basically a no-op. `VerticalOrientation` contains a mapping of "horizontally-oriented" attribute names to "vertically-oriented" ones (e.g. "width" becomes "height"). We can replace references to attributes in the DOM with calls to the `attrFor` method to ensure we use the correct attributes regardless of orientation, while preserving the "horizontally oriented" variable names in the code and without festooning the codebase with `if (vertical)` or some such.
2.  Actual canvas drawing is done by transformations on the context, so that we don't need to swap x's and y's every time we draw something in vertical mode.

Nothing that I'm aware of. Default params all still assume horizontal orientation so none of their names have changed. This code caused no test breakages.

`CanvasEntry` now takes an `Orientation` parameter. This might be avoidable by having entries refer to the parent drawer's orientation instead, if need be.

### Todos/Notes:

I'm submitting this PR in an incomplete state because while I tried to keep it minimally disruptive, it does introduce an additional layer of indirection which might be a deal breaker to merge it. I wanted to see if the overall approach seemed mergeable before doing the following:

1. Porting more plugins. I ported the regions one since it's what I intend to use myself but I'd like to port the rest as well. It's a pretty mechanical process — just add an `attrFor` everywhere we touch the DOM — but a bit tedious.
2. Adding tests. Vertical orientation seems likely to cause regressions if future code changes omit an `attrFor` call somewhere, so tests are important. But I don't know the test framework so wanted to put that off until I know whether this might be merged.

#634 is a request for this feature and the thread contains various hacks people used to get it sort-of working, mostly based on CSS transforms. AFAIK, all these approaches worked only partially at best, and they also have the problem of CSS transforms causing the waveform to be rendered perpendicular to the actual container.


example/vertical/index.html
```diff
@@ -0,0 +1,124 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+        <title>wavesurfer.js | Vertical </title>
+
```
thijstriemstra, line 6: I would use 'Vertical wave' or 'waveform' here

src/drawer.canvasentry.js
```diff
@@ -13,7 +13,7 @@ import getId from './util/get-id';
  * render a waveform, depending on the zoom level.
  */
 export default class CanvasEntry {
-    constructor() {
+    constructor(orientation) {
```
thijstriemstra, line 16: your ticket description mentions no API breakage but  this new param breaks it.

src/drawer.canvasentry.js
```diff
@@ -103,19 +105,21 @@ export default class CanvasEntry {
     updateDimensions(elementWidth, totalWidth, width, height) {
         // where the canvas starts and ends in the waveform, represented as a
         // decimal between 0 and 1
-        this.start = this.wave.offsetLeft / totalWidth || 0;
+        this.start = this.wave[this.orientation.attrFor('offsetLeft')] / totalWidth || 0;
```
thijstriemstra, line 108: I don't like this `attrFor` approach and would definitely prefer to keep the old object notation, e.g. `wave.offsetLeft`.

src/drawer.canvasentry.js
```diff
@@ -128,32 +132,35 @@ export default class CanvasEntry {
...
     /**
-     * Set the fill styles for wave and progress
+     * Set the drawing context, including fill styles and transform,
+     * for wave and progress
      *
      * @param {string} waveColor Fill color for the wave canvas
      * @param {?string} progressColor Fill color for the progress canvas
      */
-    setFillStyles(waveColor, progressColor) {
+    setDrawingContext(waveColor, progressColor) {
```
thijstriemstra, line 157: This breaks the API. I'd rather keep the old setFillStyles around and introduce a new setDrawingContext-type method.

src/drawer.js
```diff
@@ -29,6 +29,11 @@ export default class Drawer extends util.Observer {
          */
         this.height = params.height * this.params.pixelRatio;
 
+        this.orientation = util.makeOrientation(
```
thijstriemstra, line 32: ideally this has some comment with short explanation

src/wavesurfer.js
```diff
@@ -313,6 +313,7 @@ export default class WaveSurfer extends util.Observer {
      */
     static create(params) {
         const wavesurfer = new WaveSurfer(params);
+        console.log("Hello from create");
```
thijstriemstra, line 316: this can be removed

thijstriemstra (changes requested):
thanks, really cool idea and good to have!

src/drawer.canvasentry.js
valrus: TBH, I don't especially like it either, but I'm not sure how to do what you propose. We need some kind of shim to translate attributes such as this one for vertically oriented waveforms; can you suggest a way to do it more transparently, per your request? It seems like it would require overriding `wave` with some kind of wrapper, but this would be necessary for every other object for which we access the DOM attributes as well. Is that what you had in mind?

thijstriemstra: yea let's brainstorm about that. paging @sundayz @marizuccara

valrus: Perhaps a [Proxy object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) would do the trick? I'll take a stab at it the next chance I have to work on this.

valrus: I just [checked in](https://github.com/katspaugh/wavesurfer.js/pull/2195/commits/117bbd4bfcfb41074866c01f45c6f89d61a9529e) a rough implementation of this but [Proxy support](https://caniuse.com/?search=proxy) may be prohibitively low. Babel flags it as undefined. I'm not sure what other options are but I may try a manual wrapper of some kind next.

thijstriemstra: huh, Proxy is supported in 99% of the browsers isn' it: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy#browser_compatibility

package.json
```diff
@@ -29,6 +29,7 @@
     "lint-js-html": "lint-staged",
     "doc": "esdoc",
     "test": "karma start karma.conf.js",
+    "test:debug": "karma start karma.conf.js --browsers=Chrome --single-run=false",
```
thijstriemstra, line 32: maybe rename to  `test:chrome`?

src/drawer.canvasentry.js
```diff
@@ -109,7 +109,9 @@ export default class CanvasEntry {
         // set wave canvas dimensions
         this.wave.width = width;
         this.wave.height = height;
-        let elementSize = { width: elementWidth + 'px' };
+        let elementSize = {
```
thijstriemstra, line 112: no need for this formatting

src/drawer.js
```diff
@@ -86,39 +87,46 @@ export default class Drawer extends util.Observer {
...
+    getProgressPixels(wrapperBbox, clientX) {
+        if (this.params.rtl) {
+            return wrapperBbox.right - clientX;
+        } else {
+            return clientX - wrapperBbox.left;
+        }
+    }
+
     setupWrapperEvents() {
         this.wrapper.addEventListener('click', e => {
```
thijstriemstra, line 121: `e` might as well be called `event`

src/drawer.multicanvas.js
```diff
@@ -82,6 +82,13 @@ export default class MultiCanvas extends Drawer {
          * @type {number}
          */
         this.barRadius = params.barRadius || 0;
+
+        /**
+         * Whether to render the waveform vertically.
+         *
+         * @type {boolean}
+         */
+        this.vertical = params.vertical;
```
thijstriemstra, line 91: docs should point out default is false

src/drawer.canvasentry.js
valrus: Support seems decent in modern browsers but you would lose it for some older ones. I'm not sure if that's a concern. I would like to hash out a solution to this piece of things before addressing your other comments (which I'm happy to do) in case this turns out to be a hard blocker. Here's the options I see:
1. Implement this using Proxy objects, drop support entirely for older browsers as necessary. Not sure what to do about Babel flagging it in this case though.
2. Implement using Proxy objects but with a warning that vertical rendering is not supported in older browsers. Do a check in `orientation.js` whether `Proxy` is defined and, if not, make its functionality a no-op. Not sure if this will pass the linter.
3. Add a [Babel plugin](https://www.npmjs.com/package/babel-plugin-proxy) for a polyfill. It's 5 years old and warns "It is not suitable for production environments because performance impact is huge," so it may not be a _great_ option, but it is an option.
4. Something I'm missing??? I don't know if there's a way to do this cleanly without Proxy.

src/drawer.js
valrus: `e` seems to be used everywhere else in the project. Happy to change this one but it will be inconsistent with what seems to be the convention.

src/util/capitalize.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Capitalize a string's first character
+ *
+ * @param   {String} str String to capitalize
+ * @returns {String} str but with the first letter uppercased
+ * @example console.log(capitalize("hello")); // logs "Hello"
+ */
```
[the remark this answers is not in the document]
valrus: Assuming I fix the API changes (besides the addition of the `vertical` param, obviously) should I assume this will be `@since 4.6.0`?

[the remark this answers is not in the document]
thijstriemstra: correct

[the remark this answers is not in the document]
thijstriemstra: outdated again, sorry, 4.7.0 would be better ;)

src/drawer.canvasentry.js
thijstriemstra: > 1. drop support entirely for older browsers as necessary.

What browsers are we talking about exactly? and there's always a polyfill right?

> Do a check in orientation.js whether Proxy is defined and, if not, make its functionality a no-op. Not sure if this will pass the linter.

Sounds good.

thijstriemstra: ps. https://github.com/katspaugh/wavesurfer.js/blob/master/.eslintrc.js probably needs `Proxy` as well..?

src/drawer.canvasentry.js
```diff
@@ -157,6 +157,21 @@ export default class CanvasEntry {
         }
     }
 
+    /**
+     * Set the canvas transforms for wave and progress
+     *
+     * @param {boolean} vertical Whether to render vertically
+     */
+    applyCanvasTransforms(vertical) {
+        if (vertical) {
```
thijstriemstra, line 166: add a comment for params passed to setTransform, it's black magic now ;)

src/plugin/regions/index.js
```diff
@@ -127,11 +127,15 @@ export default class RegionsPlugin {
         });
         this.wavesurfer.Region = Region;
 
+        const scrollWidthProportion = 0.05;
```
thijstriemstra, line 130: ideally, add comment why it's 0.05 or near that value

src/drawer.canvasentry.js
sundayz: Pretty sure IE11 is the only browser that doesn't have Proxy support and can't be polyfilled

src/drawer.canvasentry.js
valrus: Oops! I had one in a previous iteration but must have lost it in the shuffle.

src/plugin/regions/index.js
valrus: I didn't actually add that 0.05 to the code, just modified a line that contained it already, so I don't 100% know why it was used. But I can take a stab at it.

thijstriemstra (changes requested):
Can you also add unit tests?

src/wavesurfer.js
```diff
@@ -401,7 +401,11 @@ export default class WaveSurfer extends util.Observer {
         }
```
thijstriemstra, line 402: the vertical parameter needs to be documented in this file, and added to default params

thijstriemstra (changes requested):
great! one thing left.

src/wavesurfer.js
```diff
@@ -115,6 +115,8 @@ import MediaElementWebAudio from './mediaelement-webaudio';
  * @property {boolean} splitChannels=false Render with separate waveforms for
  * the channels of the audio
  * @property {SplitChannelOptions} splitChannelsOptions={} Options for splitChannel rendering
+ * @property {boolean} vertical=false Render the waveform vertically instead of horizontally.
+ * This may not be compatible with any plugins besides the regions plugin.
```
thijstriemstra, line 119: i would move that statement to changelog instead and add 'currently'

thijstriemstra (changes requested):

src/wavesurfer.js
thijstriemstra: and build is failing on ci

valrus: I don't see it. Where should I be looking?
Screen Shot 2021-03-14 at 2 58 53 PM

src/drawer.multicanvas.js
```diff
@@ -210,11 +221,11 @@ export default class MultiCanvas extends Drawer {
         let lastEntry = this.canvases[this.canvases.length - 1];
 
         // wave
-        lastEntry.wave.parentElement.removeChild(lastEntry.wave);
+        lastEntry.wave.parentElement.removeChild(lastEntry.wave.proxiedElement);
 
         // progress
         if (this.hasProgressCanvas) {
-            lastEntry.progress.parentElement.removeChild(lastEntry.progress);
+            lastEntry.progress.parentElement.removeChild(lastEntry.progress.proxiedElement);
```
thijstriemstra, line 228: I'm still a little confused why you use `proxiedElement` property only here, and not in other places in multicanvas. Ideally we don't deal with this attribute at all..

valrus: I explain in this comment: https://github.com/katspaugh/wavesurfer.js/pull/2195/files#diff-c016f970dc3c8172ada63105abcd2b2f0e837a3c24c7592df6046dacd1bd4e0dR53

To expand a bit: Unfortunately, it doesn't seem possible to add/remove proxied elements to the DOM directly. [This blog post](https://hacks.mozilla.org/2015/07/es6-in-depth-proxies-and-reflect/) explains it briefly:

> There are limits to the fidelity of the illusion. You’ll find that `proxy !== target`. And a proxy will sometimes flunk type checks that the target would pass. Even if a proxy’s target is a DOM Element, for example, the proxy isn’t _really_ an Element; so something like `document.body.appendChild(proxy)` will fail with a `TypeError`.

I encountered this myself when testing this. So I have to use `proxiedElement` where a proxied element would touch the DOM directly.

thijstriemstra: I see. Perhaps `domElement` or `proxiedDOMElement` would better reflect the nature of the attribute. `proxiedElement` is too vague for me to be honest. I would also prefer to split the regions plugins changes into a separate PR that gets merged once this PR is merged. It would make it easier to read the changes made to core. thoughts?

CHANGES.md
```diff
@@ -11,6 +11,8 @@ x.x.x (unreleased)
   - Fix mouseup not firing if click & drag a region handle & release outside browser window (#2213)
   - Added new `showTooltip` param allowing disabling region `title` tooltip (#2213)
 - Nullify `onaudioprocess` on remove to not execute in background (#2218)
+- Add new `vertical` param enabling displaying waveforms vertically; currently not compatible
```
thijstriemstra, line 14: the second part of the sentence can be used in the separate PR that adds regions plugin support. ps. are you planning on adding support in other plugins?

src/drawer.multicanvas.js
valrus: Renamed to `domElement` per your request. The code you asked about should work, as the proxy passes through `set` to the underlying DOM element: https://github.com/katspaugh/wavesurfer.js/pull/2195/files#diff-c016f970dc3c8172ada63105abcd2b2f0e837a3c24c7592df6046dacd1bd4e0dR91

CHANGES.md
valrus: I removed the regions plugin part of the PR as you requested and will submit it in a different PR, and updated CHANGES.md too. I should be able to add support for other plugins - I think it's mostly pretty straightforward. It's not immediately obvious to me what will have to be done for the svg stuff in the markers plugin but that might just be because I don't know svgs that well. I think I should be able to do it.

thijstriemstra (approved):
thanks @valrus for sticking through the review, looks good to me.

ps. can you add @valrus to the contributors/team @katspaugh?

thijstriemstra:
@valrus if you could open the regions plugin PR in next 48 hours, we can include it in 4.7.0, otherwise it'll be in a next release.

katspaugh:
@valrus I've sent you an invite, welcome! 🎉

src/drawer.canvasentry.js
thijstriemstra: > I just [checked in](https://github.com/katspaugh/wavesurfer.js/pull/2195/commits/117bbd4bfcfb41074866c01f45c6f89d61a9529e) a rough implementation of this but [Proxy support](https://caniuse.com/?search=proxy) may be prohibitively low. Babel flags it as undefined. I'm not sure what other options are but I may try a manual wrapper of some kind next.
> Add a [Babel plugin](https://www.npmjs.com/package/babel-plugin-proxy) for a polyfill. It's 5 years old and warns "It is not suitable for production environments because performance impact is huge," so it may not be a great option, but it is an option.

I'm trying to reproduce this `Babel flags it as undefined` because babel-proxy-plugin `is 5 years old and warns "It is not suitable for production environments because performance impact is huge` which is problematic of course, stupid I missed that.

thijstriemstra: > Pretty sure IE11 is the only browser that doesn't have Proxy support and can't be polyfilled

Good point. I removed it and compilation doesn't fail. Let's see if the tests succeed as well and I'll remove it. See #2568

PR #2196, opened 2021-02-24
Add Markers plugin

osheroff:
This adds a "markers" plugin to wavesurfer, very similar to what Pro Tools gives you.  Just a way to easily jump around in a waveform. 

Lots more customization options are possible, but I figured I'd wait for the community for ask for it. 

Screen Shot 2021-02-24 at 12 38 15 PM

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,12 +1,16 @@
 wavesurfer.js changelog
 =======================
 
+x.x.x (unreleased)
+------------------
+- Markers plugin: add new plugin that allows for timeline markers
+
 4.5.0 (14.02.2021)
 ------------------
 - Split channels: `overlay` param now properly displays a single canvas (#2161)
 - Fixed memory leak with `destroy()` in `WebAudio` backend (#1940)
 - Fixed `WaveSurfer.load(url)` not working when passing a HTMLMediaElement as
-  the url parameter, with the WebAudio backend. 
+  the url parameter, with the WebAudio backend.
 - Microphone plugin: remove deprecated `MediaStream.stop` call (#2168)
 - Regions plugin: stop region dragging when mouse leaves canvas (#2158)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,7 @@ wavesurfer.js changelog
 
 x.x.x (unreleased)
 ------------------
-- Markers plugin: add new plugin that allows for timeline markers
+- Markers plugin: add new plugin that allows for timeline markers (#2196)
 
 4.5.0 (14.02.2021)
 ------------------
--- a/example/annotation/app.js
+++ b/example/annotation/app.js
@@ -1,7 +1,7 @@
 /**
  * Create a WaveSurfer instance.
  */
-let wavesurfer;
+var wavesurfer; // eslint-disable-line no-var
 
 /**
  * Init & load.
@@ -58,12 +58,9 @@ document.addEventListener('DOMContentLoaded', function() {
             //         wavesurfer.getDuration()
             //     )
             // );
-            wavesurfer.util
-                .ajax({
-                    responseType: 'json',
-                    url: 'annotations.json'
-                })
-                .on('success', function(data) {
+            fetch('/example/annotation/annotations.json')
+                .then(r => r.json())
+                .then(data => {
                     loadRegions(data);
                     saveRegions();
                 });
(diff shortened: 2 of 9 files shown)


example/zoom/index.html
```diff
@@ -17,10 +17,11 @@
         <script src="../../dist/wavesurfer.js"></script>
         <script src="../../dist/plugin/wavesurfer.timeline.js"></script>
         <script src="../../dist/plugin/wavesurfer.regions.js"></script>
+        <script src="../../dist/plugin/wavesurfer.markers.js"></script>
```
thijstriemstra, line 20: i would like to keep the zoom example as clean as possible (and remove other plugins in the future).

src/plugin/markers/index.js
```diff
@@ -0,0 +1,256 @@
...
+        this.markerWidth = 11;
+        this.markerHeight = 22;
+
+        this._onBackendCreated = () => {
+            this.wrapper = this.wavesurfer.drawer.wrapper;
+            if (this.params.markers) {
+                this.params.markers.forEach(marker => this.add(marker));
+            }
+            window.addEventListener('resize', this._updateMarkerPositions.bind(this), true);
+            window.addEventListener('orientationchange', this._updateMarkerPositions.bind(this), true);
```
thijstriemstra, line 87: these event listeners need to be cleanup on destroy.

thijstriemstra (changes requested):
thanks! can you also add a version marker in the new plugin and a changelog entry?

example/zoom/index.html
osheroff: ok, I'll pull it out.  was really just using it to test that my code reacted to zooming/scrolling properly.

osheroff:
> thanks! can you also add a version marker in the new plugin and a changelog entry?

can you clarify what you mean by version marker?

osheroff:
I've addressed all other issue besides the version marker

thijstriemstra:
> can you clarify what you mean by version marker?

see https://jsdoc.app/tags-since.html, I would put it in the new plugin class. The version nr can be 4.x.0 where x is current version + 1

example/annotation/app.js
```diff
@@ -58,12 +58,9 @@ document.addEventListener('DOMContentLoaded', function() {
             //         wavesurfer.getDuration()
             //     )
             // );
-            wavesurfer.util
-                .ajax({
-                    responseType: 'json',
-                    url: 'annotations.json'
-                })
-                .on('success', function(data) {
+            fetch('/example/annotation/annotations.json')
```
thijstriemstra, line 61: no need to use an absolute url here, i'd prefer to use the prev url format (unless it somehow doesn't work with such a relative url but i doubt it)

src/plugin/markers/index.js
```diff
@@ -0,0 +1,267 @@
...
+
+        this.style(labelDiv, {
+            display: "flex",
+            "align-items": "center",
+            cursor: "pointer"
+        });
+
+        el.appendChild(labelDiv);
+
+        labelDiv.addEventListener("click", e => {
```
thijstriemstra, line 235: this event listener is never cleaned up is it?

CHANGES.md
```diff
@@ -1,12 +1,16 @@
 wavesurfer.js changelog
 =======================
 
+x.x.x (unreleased)
+------------------
+- Markers plugin: add new plugin that allows for timeline markers
```
thijstriemstra, line 6: pr nr is missing

src/plugin/markers/index.js
osheroff: no but the div is removed from the DOM... would think that was adequate but not sure.

thijstriemstra: add a console log in the click handler, destroy wavesurfer, create a new one and see if there are 2 click console logs printed..

example/annotation/app.js
osheroff: fixed

src/plugin/markers/index.js
osheroff: confirmed, the handler is removed.

osheroff:
issues addressed.

CHANGES.md
```diff
@@ -1,12 +1,16 @@
 wavesurfer.js changelog
 =======================
 
+x.x.x (unreleased)
+------------------
+- Markers plugin: add new plugin that allows for timeline markers (#2196)
```
thijstriemstra, line 6: use 4.6.0 here

thijstriemstra (changes requested):

thijstriemstra (approved):
thanks!

PR #2201, opened 2021-03-03
webaudio: fix decodeAudioData handling in Safari

thijstriemstra:
The previous check was not sufficient and broken in some cases.

Related webkit commit: https://trac.webkit.org/changeset/265352/webkit

```
BaseAudioContext.decodeAudioData() should return a Promise as per:
- https://webaudio.github.io/web-audio-api/#dom-baseaudiocontext-decodeaudiodata
 
Behavior is unchanged for prefixed WebKitAudioContext.decodeAudioData() to ensure
backward compatibility.
```

`window.webkitAudioContext` can be flipped on/off using the `Experimental Features > Modern WebAudio API` option in Safari.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,8 +1,9 @@
 wavesurfer.js changelog
 =======================
 
-x.x.x (unreleased)
+4.6.0 (unreleased)
 ------------------
+- Webaudio: fix `decodeAudioData` handling in Safari (#2201)
 - Markers plugin: add new plugin that allows for timeline markers (#2196)
 
 4.5.0 (14.02.2021)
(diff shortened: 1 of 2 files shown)


sundayz:
I was going to test this on my iPhone but there doesn't seem to be any examples which use WebAudio for quick testing.

Code looks OK to me clearly no browser that supports the full API will have webkit prefixed AudioContext 👍

sundayz (approved):

thijstriemstra:
> I was going to test this on my iPhone but there doesn't seem to be any examples which use WebAudio for quick testing.


Eh, wavesurfer-js.org? :)

sundayz:
Yeah I ran the server locally so I can test the changes you made, but we only seem to have examples that use `backend: MediaElement` unless I missed any that use the WebAudio backend.

thijstriemstra:
Oh.. a simple `webaudio` example should be added I guess.

PR #2204, opened 2021-03-04
drawer: use `barMinHeight` according to docs

indutny:

Documentation says that `barMinHeight` is:

    If specified, draw at least a bar of this height, eliminating
    waveform gaps

However, the code uses minimum bar height only in cases of absolute
silence, leading to waveforms with bars significantly smaller than
`barMinHeight`.

Update implementation to always apply `barMinHeight` even if there is
non-zero signal at the timestamp.

I would say this is a bugfix and not a major change since it makes module behave according to its documentation.

N/A.

### Todos/Notes:

N/A

N/A

### DIFF
--- a/src/drawer.multicanvas.js
+++ b/src/drawer.multicanvas.js
@@ -301,10 +301,7 @@ export default class MultiCanvas extends Drawer {
                         peaks[Math.floor(i * scale * peakIndexScale)] || 0;
                     let h = Math.round((peak / absmax) * halfH);
 
-                    /* in case of silences, allow the user to specify that we
-                     * always draw *something* (normally a 1px high bar) */
-                    if (h == 0 && this.params.barMinHeight)
-                        h = this.params.barMinHeight;
+                    h = Math.max(h, this.params.barMinHeight);
 
                     this.fillRect(
                         i + this.halfPixel,


thijstriemstra (commented):
I think I'd rather correct the docs than change behavior..

src/drawer.multicanvas.js
```diff
@@ -301,10 +301,7 @@ export default class MultiCanvas extends Drawer {
                         peaks[Math.floor(i * scale * peakIndexScale)] || 0;
                     let h = Math.round((peak / absmax) * halfH);
 
-                    /* in case of silences, allow the user to specify that we
-                     * always draw *something* (normally a 1px high bar) */
-                    if (h == 0 && this.params.barMinHeight)
-                        h = this.params.barMinHeight;
+                    h = Math.max(h, this.params.barMinHeight);
```
marizuccara, line 304: So if a bar is smaller of the barMinHeight because the audio peaks in that point reflects a "quieter" audio, that peaks height will be replaced by the barMinHeight parameter? I think it is a bit misleading..

indutny: Well, otherwise it looks rather awkward as well. I've set it to `4`, and the silent peaks were 4 time higher than the quiet ones. I really think this is the way it has to be.

entonbiba:
yea the docs need to be updated instead, if anything this can be added as an additional enhancement option but not replace the default behavior.

src/drawer.multicanvas.js
jie-chen: Hi, has this change verified yet? We really need this feature otherwise some parts of signal would be discontinuous.

sundayz:
In my opinion the code is broken, but I think it's OK to add a temporary param like `useFixedBarMinHeightImplementation` that defaults to false, and then in the next major release change barMinHeight to work as intended.

By reading its name, I expect that if i set `barMinHeight` then I will never have any peaks that are smaller than this number.

thijstriemstra:
This branch has conflicts that must be resolved.

Issue #2191, opened 2021-02-19
Automatically publish to NPM


katspaugh:
I've noticed [this repo](https://github.com/TypeFox/monaco-languageclient/blob/master/.travis.yml) is publishing a `@next` version to NPM on each push to master.

We could do the same for our semver'ed releases based on the branch name. Like if the branch is called `release-4.5.1`, it would release that version.

Would do you think, @thijstriemstra?

thijstriemstra:
Good idea @katspaugh

katspaugh:
Cool, I'll prepare a PR on the weekend.

thijstriemstra:
fixed via #2205

PR #2205, opened 2021-03-05
Publish to NPM via GitHub Actions

katspaugh:
As discussed in #2191, let's automate the publishing of our NPM package.

When a pull request is merged to the protected `release` branch, a new version of the package will be published. Ideally, we should choose the rebase merge option so that the `release` branch is identical to master.

Here's the first successful job: https://github.com/katspaugh/wavesurfer.js/runs/2037945086?check_suite_focus=true

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -2,10 +2,11 @@ name: NPM Package
 
 on:
   push:
-    branches: [ release ]
+    branches: [ master ]
 
 jobs:
   publish-npm:
+    if: startsWith(github.ref, 'refs/heads/release')
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -6,7 +6,10 @@ on:
 
 jobs:
   publish-npm:
-    if: startsWith(github.ref, 'refs/heads/release')
+    env:
+      OLD_VERSION: $(npm show wavesurfer.js version)
+      NEW_VERSION: $(node -p 'require("./package.json").version')
+    if: env.NEW_VERSION != env.OLD_VERSION
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
@@ -27,5 +30,5 @@ jobs:
       - run: npm publish
         env:
           NODE_AUTH_TOKEN: ${{secrets.npm_token}}
-      - run: git tag $(node -p 'require("./package.json").version')
+      - run: git tag "$NEW_VERSION"
       - run: git push --tags
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -6,10 +6,6 @@ on:
 
 jobs:
   publish-npm:
-    env:
-      OLD_VERSION: $(npm show wavesurfer.js version)
-      NEW_VERSION: $(node -p 'require("./package.json").version')
-    if: env.NEW_VERSION != env.OLD_VERSION
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


thijstriemstra (commented):
> When a pull request is merged to the protected release branch, a new version of the package will be published. Ideally, we should choose the rebase merge option so that the release branch is identical to master.

Sounds slightly complicated, needing some documentation. Can you add this to bottom of readme or elsewhere?

katspaugh:
> Can you add this to bottom of readme

Good point, will add a line about this!

thijstriemstra:
Also, the website on gh-pages also needs to be updated after each release, would be nice if this can be automated (here?) as well.

thijstriemstra:
Also, how do you limit this so not every contributor can publish a release? By limiting who can create a tag?

katspaugh:
> Also, the website on gh-pages also needs to be updated after each release, would be nice if this can be automated (here?) as well.

I can make another PR with a separate action for the gh-pages branch.

> Also, how do you limit this so not every contributor can publish a release? By limiting who can create a tag?

Anyone can create a pull request to the `release` branch but it needs to be approved by a contributor. And this is completely independent from tags. I'm not sure we really need tags at all. AFAIK, there's nothing that depends on them.

thijstriemstra:
>   I'm not sure we really need tags at all. AFAIK, there's nothing that depends on them.

True, but we created tags for every single release and that would stop here? I'd prefer to keep that going as a good development practice tbh. I see 4.6.0 was also not tagged..

katspaugh:
We can keep tagging major versions manually, I just don't want this to be a blocker for NPM releases.

Triggering the action on a push to a designated branch has a few advantages:
* The pull request gives you an overview of the commits that go to the new version
* It can all be done completely via GitHub's web UI
* Flexible protection rules based on the branch

thijstriemstra:
> We can keep tagging major versions manually

Who will? I thought we want to prevent any manual steps.

thijstriemstra:
> Are you proposing that we push the tags from the Action? It should be doable.

I would prefer that, yeah.

In another project I work one we also setup auto publish on tag: https://github.com/rm-hull/luma.core/commit/a250b5085a14a65cdbe4e42d49756f95ce052623 but not sure yet how it would work with npm packages.

thijstriemstra:
that's good news. But, correct me if i'm wrong, does this mean a release now requires two pull requests and 2 reviews?

katspaugh:
Hmm, that's true.

We could run this action on push from `release*` to `master`. WDYT?

thijstriemstra:
I thought it would be as simple as creating a branch with a release with a certain format, e.g. `1.2.3`, and once it's merged (into master) it will build and publish the tag (not master).

katspaugh:
OK, the best I could come up with is the following:

 * On every push to master, we check the current version on NPM and the local one from the package.json.
 * If they differ, we tag & publish.

Please see
https://github.com/katspaugh/wavesurfer.js/pull/2205/files#diff-551d1fcf87f78cc3bc18a7b332a4dc5d8773a512062df881c5aba28a6f5c48d7R30-R35

The syntax is pretty ugly but it works, I've tested on a different branch.

thijstriemstra:
Awesome, that's a good start. Updating the version nr in package.json and bower.json can be a local script someone can run when creating a release pr. Are you 100% sure it only runs when merged to master?

> I can make another PR with a separate action for the gh-pages branch.

Sounds good.

katspaugh:
> Are you 100% sure this publish action only runs when merged to master?
Yes.

> Can you start that pr before we merge this one? I'd like to see how they tie together.
I'd prefer to merge this one first.

thijstriemstra:
> We can immediately start using it and see if it works well or not.

Sounds good.

thijstriemstra (approved):
Thanks, go ahead and merge.

PR #2207, opened 2021-03-06
add `returnToPlayhead` parameter

osheroff:
This emulates default behavior in some DAWs (notably protools), where you may be working on audio from a specific point; in these cases it's a lousy experience to constantly seek back to your previous play position.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 wavesurfer.js changelog
 =======================
 
+4.6.1 (unreleased)
+------------------
+- Add `returnToPlayhead` parameter, emulating ProTools behavior
+
 4.6.0 (04.03.2021)
 ------------------
 - Webaudio: fix `decodeAudioData` handling in Safari (#2201)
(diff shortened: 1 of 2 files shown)


src/wavesurfer.js
```diff
@@ -721,7 +724,12 @@ export default class WaveSurfer extends util.Observer {
             this.fireEvent('finish');
         });
         this.backend.on('play', () => this.fireEvent('play'));
-        this.backend.on('pause', () => this.fireEvent('pause'));
+        this.backend.on('pause', () => {
+            if ( this.params.returnToPlayhead ) {
```
thijstriemstra, line 728: adding more complexity to these event triggers is something we should prevent at all times.

osheroff: the `pause()` method then?

thijstriemstra: let's give it a try.

sundayz:
Some audio tools let you change the return position independently of seeking. Feels like maybe this could be a plugin so it's easier to modify the behaviour, I suspect everyone will want it to work slightly differently :p

CHANGES.md
```diff
@@ -1,6 +1,10 @@
 wavesurfer.js changelog
 =======================
 
+4.6.1 (unreleased)
```
thijstriemstra, line 4: x.x.x instead of specific version

osheroff:
ok, I made it a plugin #2209

PR #2208, opened 2021-03-07
markers plugin: markers at the begin/end of file were screwing up seeking

osheroff:
markers placed at the beginning or the end of a region weren't properly overflowing, instead they pushed the wavesurfer container to be wider and that messed up calculations when seeking around a waveform.

Does this need a CHANGELOG?  not sure if you've released the marker stuff into the wild yet.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 wavesurfer.js changelog
 =======================
 
+x.x.x (unreleased)
+------------------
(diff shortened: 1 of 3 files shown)


thijstriemstra:
> Does this need a CHANGELOG? not sure if you've released the marker stuff into the wild yet.

Always.

osheroff:
ok, changelog entry added.

thijstriemstra (approved):
Looks good. Squash merge away.

PR #2209, opened 2021-03-08
Playhead plugin

osheroff:
@katspaugh as suggested in the other PR, here's a plugin for doing return-to-last-play-time functionality.  I allowed for configuration on the 3 bits of functionality introduced, defaulting them all to "on" (I'm just an egotist like that).

### DIFF
--- a/src/plugin/playhead/index.js
+++ b/src/plugin/playhead/index.js
@@ -1,0 +1,225 @@
+
+/**
+ * The playhead plugin separates the notion of the currently playing position from
+ * a "play-start" position.  Having a playhead enables a listening pattern
+ * (commonly found in DAWs) that involves listening to a section of a track
+ * repeatedly, rather than listening to an entire track in a linear fashion.
+ *
+ * @implements {PluginClass}
+ *
+ * @example
+ * import PlayheadPlugin from 'wavesurfer.playhead.js';
+ *
+ * // if you are using <script> tags
+ * var PlayheadPlugin = window.WaveSurfer.playhead;
+ *
+ * // ... initialising wavesurfer with the plugin
+ * var wavesurfer = WaveSurfer.create({
+ *   // wavesurfer options ...
+ *   plugins: [
+ *     PlayheadPlugin.create({
+ *        movePlayheadOnSeek: true,
+ *        drawPlayhead: false,
+ *        movePlayheadOnPause: false
+ *     })
+ *   ]
+ * });
+ */
+
+const DEFAULT_FILL_COLOR = "#CF2F00";
+
+export default class PlayheadPlugin {
+    /**
+     * @typedef {Object} PlayheadPluginParams
+     * @property {?boolean} draw=true Draw the playhead as a triangle/line
+     * @property {?boolean} moveOnSeek=true Seeking (via clicking) while playing moves the playhead
+     * @property {?boolean} returnOnPause=true Pausing the track returns the seek position to the playhead
+     */
+
+    /**
+     * Playhead plugin definition factory
+     *
+     * This function must be used to create a plugin definition which can be
+     * used by wavesurfer to correctly instantiate the plugin.
+     *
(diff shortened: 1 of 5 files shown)


src/plugin/playhead/index.js
```diff
@@ -0,0 +1,225 @@
...
+     */
+
+    /**
+     * Playhead plugin definition factory
+     *
+     * This function must be used to create a plugin definition which can be
+     * used by wavesurfer to correctly instantiate the plugin.
+     *
+     * @param {PlayheadPluginParams} params parameters use to initialise the plugin
+     * @since 4.6.0
```
thijstriemstra, line 46: should be 4.7.0

thijstriemstra (changes requested):
tests are failing..

src/plugin/playhead/index.js
```diff
@@ -0,0 +1,225 @@
...
+    _createPointerSVG() {
+        const svgNS = "http://www.w3.org/2000/svg";
+
+        const el = document.createElementNS(svgNS, "svg");
+        const path = document.createElementNS(svgNS, "path");
+
+        el.setAttribute("viewBox", "0 0 33 30");
+        path.setAttribute("d", "M16.75,0 L31.705211,25.4338659 C32.570163,26.8470269 32.1257518,28.6938024 30.7125908,29.5587544 C30.2411619,29.8473013 29.6991813,30 29.1464566,30 L4.35354336,30 C2.69668911,30 1.35354336,28.6568542 1.35354336,27 C1.35354336,26.4472753 1.50624201,25.9052947 1.794789,25.4338659 L16.75,0 L16.75,0 Z");
+        path.setAttribute("transform", "translate(16.750000, 15.500000) scale(1, -1) translate(-16.750000, -15.500000)");
+        path.setAttribute("id", "polygon");
```
katspaugh, line 161: I've optimised this icon using svgo, and it encoded the transform into the path itself, so you can remove it. Along with the id attribute.
```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 33 30">
  <path d="M16.75 31 31.705 5.566A3 3 0 0 0 29.146 1H4.354a3 3 0 0 0-2.56 4.566L16.75 31z" stroke="#979797" fill="..." /></svg>
```

src/plugin/playhead/index.js
```diff
@@ -0,0 +1,225 @@
...
+            width: this.markerWidth + "px",
+            height: this.markerHeight + "px",
+            cursor: "pointer",
+            "z-index": 5
+        });
+        return el;
+    }
+
+    _createPlayheadElement() {
+        const el = document.createElement('playhead');
```
katspaugh, line 177: I see a mixture of single and double quotes. Not sure if we have a lint rule for this. It would be nice to use just one type of quotes.

katspaugh:
A really nice plugin, great work! I use this feature in Audacity all the time, really useful.

src/plugin/playhead/index.js
```diff
@@ -0,0 +1,225 @@
...
+            position: "absolute",
+            height: "100%",
+            display: "flex",
+            "flex-direction": "column"
+        });
+
+        const pointer = this._createPointerSVG();
+        el.appendChild(pointer);
+
+        pointer.addEventListener("click", e => {
```
thijstriemstra, line 190: pointer event listener is not cleaned up (on destroy).

src/plugin/playhead/index.js
osheroff: fixed to use single

src/plugin/playhead/index.js
osheroff: fixed

src/plugin/playhead/index.js
osheroff: thanks, fixed.

osheroff:
> tests are failing..

definitely not as a result of this PR, some SSL error

I think I've hit all the relevant CR comments.

thijstriemstra (approved):

PR #2211, opened 2021-03-08
Fix setVolume not working with MediaElementWebAudio on iOS Safari

sundayz:
Previously, calling setVolume with the `MediaElementWebAudio` backend would not work on iOS Safari, because you cannot change the volume of a MediaElement node programmatically. By setting volume through the gain node, just like the WebAudio backend does, you can now change the volume level.

This means that setVolume now works on iOS with `MediaElementWebAudio` and `WebAudio`, but not with `MediaElement` backend.

Tested on latest iOS 14.4 on iPhone 8.
Also tested on Chrome Windows 10 to make sure this doesn't break setting volume elsewhere.


sundayz:
Not sure what's happening here with `getVolume` not working

src/mediaelement-webaudio.js
```diff
@@ -74,4 +74,28 @@ export default class MediaElementWebAudio extends MediaElement {
 
         this.destroyWebAudio();
     }
+
+    /**
+     * Set the audio volume.
+     *
+     * Previously, calling setVolume on this backend would use the MediaElement backend's setVolume,
+     * which set the volume on the media element. On iOS Safari, media element volume cannot be
+     * programmatically set. Using the gain node allows volume to be changed programmatically.
```
thijstriemstra, line 83: > On iOS Safari, media element volume cannot

But I don't see any exception made for Safari so this workaround is safe to apply everywhere then?

thijstriemstra:
> Tested on latest iOS on iPhone 8.

In a few months 'latest' iOS is a different version. Please be specific.

src/mediaelement-webaudio.js
```diff
@@ -74,4 +74,28 @@ export default class MediaElementWebAudio extends MediaElement {
...
+     * @param {number} value A floating point value between 0 and 1.
+     */
+    setVolume(value) {
+        this.gainNode.gain.setValueAtTime(value, this.ac.currentTime);
+    }
+
+    /**
+     * Get the current volume.
+     *
+     * @since 4.7.0 See {@link MediaElementWebAudio#setVolume}
```
thijstriemstra, line 95: `@since` should only contain a version nr

thijstriemstra (changes requested):
thanks. tests are failing though:

```
* should set volume : 1 ok, 1 failed
* should get volume : 2 failed

Browser results:

 - Chrome Headless 88.0.4324.182 (Linux x86_64): 125 tests
   - 123 ok, 2 failed
 - Firefox 86.0 (Ubuntu 0.0.0): 125 tests
   - 124 ok, 1 failed
```

sundayz:
https://github.com/katspaugh/wavesurfer.js/blob/a41859187e73c7eec7f2c9512cef21b0ffdbcdce/spec/mediaelement-shared.js#L218

Looks like "should get volume" is failing because it sets the volume directly on the HTMLMediaElement rather than using setVolume. Is this how the MediaElementWebAudio should be used? I think you should always use `setVolume`. Except maybe when WaveSurfer is first created we can call `this.setVolume(this.sourceMediaElement.volume)` which would correctly set the initial volume to what the HTMLMediaElement has.

thijstriemstra:
>  I think you should always use setVolume

Agreed.

sundayz:
I will be testing this again soon with 14.5 (has new webaudio implementation)

thijstriemstra:
> I will be testing this again soon with 14.5 (has new webaudio implementation)

14.6 is out now, any updates?

thijstriemstra:
I tested a wavesurfer.js (5.1.0) app with `MediaElementWebAudio` backend on iOS 14.7 and audio is silent but playhead is moving etc. Will give this PR a try.

sundayz:
I think this PR has issues but I don't remember if they were caused by the code or by Safari incompatibility. I will be reviewing this soon because I will be testing MediaElementWebAudio on iOS again in the coming weeks.

thijstriemstra:
First test failure:

```
03 08 2021 16:19:33.423:ERROR [Firefox 90.0 (Ubuntu 0.0.0) | WaveSurfer/MediaElementWebAudio: | should get volume]: Expected 1 to equal 0.3.
<Jasmine>
sharedTests/</<@http://localhost:9876/base/dist/commons.js?97c0b7b1abb84724813e5d4d4961f0a9e2cf6ed0:3474:22
fn@http://localhost:9876/base/dist/commons.js?97c0b7b1abb84724813e5d4d4961f0a9e2cf6ed0:26944:17
fireEvent/<@http://localhost:9876/base/dist/commons.js?97c0b7b1abb84724813e5d4d4961f0a9e2cf6ed0:27019:12
fireEvent@http://localhost:9876/base/dist/commons.js?97c0b7b1abb84724813e5d4d4961f0a9e2cf6ed0:27016:85
loadMediaElement/<@http://localhost:9876/base/dist/commons.js?97c0b7b1abb84724813e5d4d4961f0a9e2cf6ed0:41259:19
fn@http://localhost:9876/base/dist/commons.js?97c0b7b1abb84724813e5d4d4961f0a9e2cf6ed0:26944:17
fireEvent/<@http://localhost:9876/base/dist/commons.js?97c0b7b1abb84724813e5d4d4961f0a9e2cf6ed0:27019:12
fireEvent@http://localhost:9876/base/dist/commons.js?97c0b7b1abb84724813e5d4d4961f0a9e2cf6ed0:27016:85
_setupMediaListeners/this.mediaListeners.canplay@http://localhost:9876/base/dist/commons.js?97c0b7b1abb84724813e5d4d4961f0a9e2cf6ed0:20546:16
Firefox 90.0 (Ubuntu 0.0.0) WaveSurfer/MediaElementWebAudio: should get volume FAILED
	Expected 1 to equal 0.3.
	<Jasmine>
	sharedTests/</<@dist/commons.js:3474:22
	fn@dist/commons.js:26944:17
	fireEvent/<@dist/commons.js:27019:12
	fireEvent@dist/commons.js:27016:85
	loadMediaElement/<@dist/commons.js:41259:19
	fn@dist/commons.js:26944:17
	fireEvent/<@dist/commons.js:27019:12
	fireEvent@dist/commons.js:27016:85
	_setupMediaListeners/this.mediaListeners.canplay@dist/commons.js:20546:16
Chrome Headless 92.0.4515.107 (Linux x86_64): Executed 135 of 135 (2 FAILED) (5.494 secs / 4.648 secs)
```

And second failure:

```
 03 08 2021 16:19:29.976:ERROR [Chrome Headless 92.0.4515.107 (Linux x86_64) | WaveSurfer/MediaElementWebAudio: | should set volume]: Error: Expected 1 to equal 0.5.
    at <Jasmine>
    at WaveSurfer.<anonymous> (http://localhost:9876/base/dist/commons.js?97c0b7b1abb84724813e5d4d4961f0a9e2cf6ed0:3493:38)
    at fn (http://localhost:9876/base/dist/commons.js?97c0b7b1abb84724813e5d4d4961f0a9e2cf6ed0:26944:17)
    at http://localhost:9876/base/dist/commons.js?97c0b7b1abb84724813e5d4d4961f0a9e2cf6ed0:27019:12
Chrome Headless 92.0.4515.107 (Linux x86_64) WaveSurfer/MediaElementWebAudio: should set volume FAILED
	Error: Expected 1 to equal 0.5.
	    at <Jasmine>
	    at WaveSurfer.<anonymous> (dist/commons.js:3493:38)
	    at fn (dist/commons.js:26944:17)
	    at dist/commons.js:27019:12
Chrome Headless 92.0.4515.107 (Linux x86_64): Executed 135 of 135 (2 FAILED) (0 secs / 4.648 secs)
```

thijstriemstra:
> because you cannot change the volume of a MediaElement node programmatically.

You mean, you can't do this in Safari iOS but can do in other browsers? Any docs/link that backs up this claim?

sundayz:
> You mean, you can't do this in Safari iOS but can do in other browsers? Any docs/link that backs up this claim?

https://developer.apple.com/library/archive/documentation/AudioVideo/Conceptual/Using_HTML5_Audio_Video/Device-SpecificConsiderations/Device-SpecificConsiderations.html

"On iOS devices, the audio level is always under the user’s physical control. The volume property is not settable in JavaScript"

Also the same for `playbackRate`. And yes only iOS Safari has this limitation.

The way we get around it at work is by using `WebAudio`, because that uses a gain node. `MediaElementWebAudio` would be a great solution because it's much more performant but as you have noted yourself, it has very glitchy audio on Safari, basically unusable. But `WebAudio` is suitable for only a couple of mp3 files, loading 10-20 can crash the browser on iOS.

I did extensive testing with `MediaElementWebAudio`, nothing I could change seemed to help. I tried changing the sample rate, passing my own AudioContext, changing the buffer size of the script node.

I did some very short testing with the new WebAudio implementation in iOS14 and it still has problems. According to a comment here it's still a bug in iOS 15 beta: https://bugs.webkit.org/show_bug.cgi?id=221334

One thing I did not test was an Audio worklet. I was using the now deprecated ScriptProcessorNode. Maybe moving the work away from the main thread helps??  I understand that they are now implemented in Safari too and are replacing script processor node.

sundayz:
At work we are getting reports of choppy audio even using the `WebAudio` backend now, which I've never experienced problems with before.

Safari support is basically so bad even after their WebAudio rework that in the future we will be switching to a different architecture: instead of doing the work on the client, we will do all the work on the server and stream it to the browser, that way it's a simple MediaElement with a single source and single output.

Of course that's not realistic for everyone to do, but if your use case is for example an audio editing tool in the browser, that's what you gotta do in 2021 it seems.

thijstriemstra:
it's insane a company with such enormous resources can't get things right.

sundayz:
My last hope is that AudioWorklet magically makes things better on iOS

entonbiba:
@sundayz is this issue still relevant with MediaElement backend on safari?

sundayz:
> @sundayz is this issue still relevant with MediaElement backend on safari?

Yes, you still cannot set the `volume` property of a HTMLMediaElement on any iOS device, by design. I don't know if this issue is still relevant to Wavesurfer though. From what I remember, you can use `AudioContext.createMediaElementSource()` to pipe a MediaElement source into WebAudio and then modify its volume that way, but I don't think we have that implemented, and also that function used to be buggy(?) on iOS, but I have not worked with it in a while, maybe things have changed now.

entonbiba:
@sundayz can you test this on ios 16? I believe volume changes in code can only happen after a user initiates a interaction with the mediaelement. I'll test it on my end, but if you have the sample code you used that be great to look at?

sundayz:
> @sundayz can you test this on ios 16? I believe volume changes in code can only happen after a user initiates a interaction with the mediaelement. I'll test it on my end, but if you have the sample code you used that be great to look at?

Try on desktop and then on iOS

https://codepen.io/sundays/pen/wvErVXy

katspaugh:
@sundayz the code looks good! Could you please solve the git conflicts so that we can merge this?

entonbiba:
@katspaugh git conflicts fixed in this draft,

katspaugh:
Thanks for checking, makes sense!
I guess we should fix the test before merging it.

PR #2213, opened 2021-03-09
regions plugin: drag bugfix & showTooltip parameter

marcelblum:
Improvements addressing 2 minor annoyances (including 1 legit bug fix):

1. Fix a bug that prevented firing mouseup to end a region drag if the user clicked and dragged a region handle and released the mouse button outside of the browser window. Fixed by attaching the `mouseup` event listener to `document` instead of `document.body`.
2. Added new `RegionParams` option `showTooltip` allowing disabling the `title=...` tooltip (just because I found the unstyleable tooltip obtrusive and got tired of having to do `element.removeAttribute("title")` all the time).

Nothing. The default value for the new `showTooltip` parameter is the old behavior.

Nothing

### DIFF
--- a/src/plugin/regions/region.js
+++ b/src/plugin/regions/region.js
@@ -675,18 +675,18 @@ export class Region {
         document.body.addEventListener('mousemove', onMove);
         document.body.addEventListener('touchmove', onMove);
 
-        document.body.addEventListener('mouseup', onUp);
+        document.addEventListener('mouseup', onUp);
         document.body.addEventListener('touchend', onUp);
 
         this.on('remove', () => {
-            document.body.removeEventListener('mouseup', onUp);
+            document.removeEventListener('mouseup', onUp);
             document.body.removeEventListener('touchend', onUp);
             document.body.removeEventListener('mousemove', onMove);
             document.body.removeEventListener('touchmove', onMove);
         });
 
         this.wavesurfer.on('destroy', () => {
-            document.body.removeEventListener('mouseup', onUp);
+            document.removeEventListener('mouseup', onUp);
             document.body.removeEventListener('touchend', onUp);
         });
     }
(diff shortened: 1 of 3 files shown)


osheroff:
@thijstriemstra will have words if you don't add a CHANGELOG entry

osheroff:
two minor requests, otherwise fine.

src/plugin/regions/region.js
```diff
@@ -42,6 +42,7 @@ export class Region {
         this.handleRightEl = null;
         this.data = params.data || {};
         this.attributes = params.attributes || {};
+        this.showTooltip = params.showTooltip ?? true;
```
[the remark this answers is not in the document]
marcelblum: I figured it was ok because it gets converted to older format JS by webpack anyway.

src/plugin/regions/region.js
```diff
@@ -171,7 +172,7 @@ export class Region {
         const regionEl = document.createElement('region');
 
         regionEl.className = 'wavesurfer-region';
-        regionEl.title = this.formatTime(this.start, this.end);
+        if (this.showTooltip) regionEl.title = this.formatTime(this.start, this.end);
```
thijstriemstra, line 175: please use extended if statement

src/plugin/regions/region.js
```diff
@@ -301,7 +302,7 @@ export class Region {
                 );
             }
 
-            this.element.title = this.formatTime(this.start, this.end);
+            if (this.showTooltip) this.element.title = this.formatTime(this.start, this.end);
```
thijstriemstra, line 305: please use extended if statement

thijstriemstra (changes requested):

src/plugin/regions/region.js
marcelblum: done

thijstriemstra (approved):

PR #2227, opened 2021-03-17
get examples working again

osheroff:
many of them are sharing a wavesurfer instance with `trivia.js` via a global variable.  I can't say this is the best idea, but this is at least explicit about it.

If there's a better approach I'm all ears, I looked down the road of timeline.js fishing out the wavesurfer instance from the DOM but didn't see an easy way.

### DIFF
--- a/example/audio-element/main.js
+++ b/example/audio-element/main.js
@@ -1,7 +1,7 @@
 'use strict';
 
 // Create an instance
-var wavesurfer; //eslint-disable-line
+var wavesurfer;
 
 // Init & load audio file
 document.addEventListener('DOMContentLoaded', function() {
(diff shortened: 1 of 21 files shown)


example/audio-element/main.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 // Create an instance
-let wavesurfer;
+var wavesurfer; //eslint-disable-line
```
thijstriemstra, line 4: instead of adding this eslint exception everywhere, disable the no-var rule for examples only in https://github.com/katspaugh/wavesurfer.js/blob/master/.eslintrc.js#L47

thijstriemstra (changes requested):

example/audio-element/main.js
osheroff: k done

thijstriemstra (approved):
thanks!

PR #2242, opened 2021-04-05
regions plugin: support for vertical orientation

valrus:
This PR adds support for vertically-oriented waveforms to the regions plugin. It's a follow up to the original vertical waveform PR #2195.

### API breakages:
None AFAIK.

### Todos/Notes:
Other plugins still need vertical track support.

PR #2195, issue #634.


src/plugin/regions/region.js
```diff
@@ -130,7 +131,7 @@ export class Region {
     /* Remove a single region. */
     remove() {
         if (this.element) {
-            this.wrapper.removeChild(this.element);
+            this.wrapper.removeChild(this.element.proxiedElement);
```
thijstriemstra, line 134: this should be `domElement` afaik

thijstriemstra (changes requested):
lot's of changes going on here, hard to review as I don't use/focus on regions plugin.

katspaugh:
Thanks for the cool PR @valrus!

Do you have it deployed anywhere so that we can play around with the vertical waveform?

valrus:
> Thanks for the cool PR @valrus!
> 
> Do you have it deployed anywhere so that we can play around with the vertical waveform?

Hey - sorry, I don't. I'm using it in a personal project but it's not ready to be more publicly available yet. When I have the time maybe I can throw up a basic demo that doesn't depend on a server like the aforementioned project does.

valrus:
@katspaugh OK, I tossed up a quick demo for you! It's really just the existing `regions` example with `vertical: true` and some CSS tweaks also used in the current `vertical` example. Here it is: https://ian.mccowan.space/wavesurfer-vertical-demo/

katspaugh:
Works great, thank you!

katspaugh (approved):

thijstriemstra (approved):
thanks!

PR #2252, opened 2021-04-27
regions plugin: fix Intervention message on mobile devices

### LINKED ISSUE #2251: Region plugin: Moving region on mobile causes an [Intervention]
### Wavesurfer.js version(s):
4.6.0

Chrome 90

Using Mobile emulator on devtools

https://wavesurfer-js.org/plugins/regions.html

When dragging a region on mobile it causes an Intervention message
`[Intervention] Unable to preventDefault inside passive event listener due to target being treated as passive.`

### Possible fix
in `wavesurfer.region.js`
update to 
`document.body.addEventListener('touchmove', onMove, {passive: false});`

```
{
    capture: Boolean,  // listener will be triggered when event propagates to this event target in its capture phase
    once: Boolean, // whether or not you wish to call the handler only once
    passive: Boolean  // whether this listener will never call `preventDefault`
}
```

dimshik100:

fixes #2251

### DIFF
--- a/src/plugin/regions/region.js
+++ b/src/plugin/regions/region.js
@@ -680,7 +680,7 @@ export class Region {
         this.element.addEventListener('touchstart', onDown);
 
         document.body.addEventListener('mousemove', onMove);
-        document.body.addEventListener('touchmove', onMove);
+        document.body.addEventListener('touchmove', onMove, {passive: false});
 
         document.addEventListener('mouseup', onUp);
         document.body.addEventListener('touchend', onUp);


thijstriemstra:
First time I see this. Background info on https://developers.google.com/web/updates/2017/01/scrolling-intervention

sundayz:
For some reason this function, which is called by `onMove`, calls `preventDefault`. But judging by the comments it wasn't intended to handle dragging, since the comments talk about click events?

https://github.com/katspaugh/wavesurfer.js/blob/c1f0a4de02ac39ef286229751e7054b395fec1d8/src/drawer.js#L87

Having passive events where possible makes scrolling more responsive, it's not clear to me if calling preventDefault here is needed.

thijstriemstra (changes requested):
Does `preventDefault` really have to be called in that method @dimshik100? What about removing the `preventDefault` call instead of making it passive? Would this have any side effects?

dimshik100:
> Does `preventDefault` really have to be called in that method @dimshik100? What about removing the `preventDefault` call instead of making it passive? Would this have any side effects?

That's a good question.
I did not dive deeper to understand if that is needed or not, just added a fix to the issue I was experiencing.
The current proposed PR should fix the current errors and improve performance.
Maybe there is a way to improve this all thing with a more significant refactor.

thijstriemstra (changes requested):
> The current proposed PR should fix the current errors and improve performance.

Agreed. @dimshik100 can you also add a changelog entry?

thijstriemstra (approved):

PR #2269, opened 2021-05-13
markers plugin: added support for custom marker elements

bknower:
Markers can now be provided with an image to display instead of the default pointer image when being added.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@ wavesurfer.js changelog
 ------------------
 
 - Fix removing DOM element on `destroy()` (#2258)
+- Markers plugin: add the ability to use custom HTML elements in place of the default marker icon by passing the new `markerElement` parameter to the marker constructor (#2269)
 
 5.0.0 (02.05.2021)
 ------------------
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,11 +1,13 @@
 wavesurfer.js changelog
 =======================
 
+x.x.x (unreleased) 
+------------------
+- Markers plugin: add the ability to use custom HTML elements in place of the default marker icon by passing the new `markerElement` parameter to the marker constructor (#2269)
+
 5.0.1 (05.05.2021)
 ------------------
-
 - Fix removing DOM element on `destroy()` (#2258)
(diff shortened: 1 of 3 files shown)


thijstriemstra:
good idea. what if someone wants to add videos, we'd have to add another parameter to that method, I'd prefer a different approach that would make it easy to support various types (eg audio/video/img).

bknower:
> good idea. what if someone wants to add videos, we'd have to add another parameter to that method, I'd prefer a different approach that would make it easy to support various types (eg audio/video/img).

Does it make sense to display anything other than an image as a marker? This is just for the markers like this 
image

thijstriemstra (changes requested):
> Does it make sense to display anything other than an image as a marker?

Yeah, a canvas element. But also an html element like a button or checkbox. Can you make it more generic? Also, can you add this new feature to the markers example?

thijstriemstra (changes requested):
thanks. can you also document the new option here: https://github.com/katspaugh/wavesurfer.js/blob/554ba967c8da3e271fab911c4ca8b0eaa4260806/src/plugin/markers/index.js#L3

And add a changelog entry?

osheroff:
needs documentation.  Interface is highly flexible but seems ripe for shooting yourself in the foot by re-using the same HTMLElement; I could see the argument for detecting that case?  I was going to say "just use `cloneNode`" but maybe that hampers some use cases.   Dunno.

@thijstriemstra is there an example of a plugin accepting an HTMLElement we could follow?

thijstriemstra:
> is there an example of a plugin accepting an HTMLElement we could follow?

Not that I know of. I think this PR is good enough for a first implementation.

CHANGES.md
```diff
@@ -5,6 +5,7 @@ wavesurfer.js changelog
 ------------------
 
 - Fix removing DOM element on `destroy()` (#2258)
+- Markers plugin: add the ability to use custom HTML elements in place of the default marker icon by passing the new `markerElement` parameter to the marker constructor (#2269)
```
thijstriemstra, line 8: move this changelog entry to a new section called `x.x.x (unreleased)`

thijstriemstra (approved):
thanks.

PR #2287, opened 2021-05-31
markers plugin: trigger click event on wavesurfer

Wangbaogang:

Changes: When you click `labelDiv`, the `marker-click`event of wavesurfer is triggered (not previously).

Reasons:  In fact, when I want to do more when I click on marker, event is necessary

No

No

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,7 +4,6 @@ wavesurfer.js changelog
 x.x.x (unreleased)
 ------------------
 - Markers plugin: add the ability to use custom HTML elements in place of the default marker icon by passing the new `markerElement` parameter to the marker constructor (#2269)
-
 - Regions plugin: handle rollover cursor bug fix (#2293)
 - Markers plugin:  trigger `marker-click` event on wavesurfer (#2287)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,9 +1,10 @@
 wavesurfer.js changelog
 =======================
 
-x.x.x (unreleased) 
+x.x.x (unreleased)
 ------------------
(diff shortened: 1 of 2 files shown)


osheroff:
Offhand I'd say we probably don't want to fire a general "click" event, consumers will probably mostly be treating those as clicking inside the waveform.

I like the marker-click one tho

thijstriemstra:
> I like the marker-click one tho

So what do you suggest for @Wangbaogang to follow up with?

osheroff:
I think just remove the wavesurfer-level click event

Wangbaogang:
@osheroff In fact, using `marker-click` event on wavesurfer is a handy way, just like `region-click` of `RegionsPlugin`, I think.

osheroff:
I'm talking about this line:

```
            this.fireEvent("click", e);
```

I think this line is problematic as it will appear to an event listener that the user clicked on the waveform, which they did not.

thijstriemstra:
The new event also needs to be documented (at top of class) and a changelog entry needs to be added.

thijstriemstra (changes requested):
@Wangbaogang 

- the new event needs to be documented (at top of class)
- a changelog entry needs to be added
- remove the `click` fireEvent

Wangbaogang:
> @Wangbaogang
> 
> * the new event needs to be documented (at top of class)
> * a changelog entry needs to be added
> * remove the `click` fireEvent

Added changelog and removed click. But It seems there is no place to edit the plugin information in the repo. Maybe I didn't find it?

thijstriemstra:
> there is no place to edit the plugin information in the repo

@Wangbaogang it's in the `gh-pages` branch. Also add a jsdoc for the new event.

thijstriemstra (approved):
thanks!

PR #2293, opened 2021-06-01
regions plugin: cursor orientation bugfix

marcelblum:
Recent change to support vertical orientation in the Regions plugin introduced a bug that made the rollover cursor for region handles always show the vertical cursor even when in non-vertical mode.

Nothing

None


thijstriemstra (approved):
thanks.

PR #2294, opened 2021-06-01
regions plugin: take formatTimeCallback from plugin params

rdn32:
**Title:** regions plugin: take formatTimeCallback from plugin params

The region plugin documentation for formatTimeCallback gives it as being a plugin parameter, so that's where it's now taken from.

If anyone was relying on being able to set formatTimeCallback on a per-region basis, the this would be a breaking change.

None

### Todos/Notes:

My assumption here is that the existing behaviour - taking this setting from the region parameters - was an accident, and that the documented behaviour is correct. It would be possible, of course, to keep compatibility with the existing behaviour by having the code look to both sets of parameters.

There appears to be a similar discrepancy between documentation and implementation regarding edgeScrollWidth, but I haven't done anything about that.

### DIFF
--- a/src/plugin/regions/index.js
+++ b/src/plugin/regions/index.js
@@ -201,7 +201,7 @@ export default class RegionsPlugin {
             params = {...params, minLength: this.regionsMinLength};
         }
 
-        const region = new this.wavesurfer.Region(params, this.util, this.wavesurfer);
+        const region = new this.wavesurfer.Region(params, this.params, this.util, this.wavesurfer);
 
         this.list[region.id] = region;
 
--- a/src/plugin/regions/region.js
+++ b/src/plugin/regions/region.js
@@ -9,7 +9,7 @@
  * @extends {Observer}
  */
 export class Region {
-    constructor(params, regionsUtils, ws) {
+    constructor(params, pluginParams, regionsUtils, ws) {
         this.wavesurfer = ws;
         this.wrapper = ws.drawer.wrapper;
         this.util = ws.util;
@@ -77,7 +77,7 @@ export class Region {
             }
         }
 
-        this.formatTimeCallback = params.formatTimeCallback;
+        this.formatTimeCallback = pluginParams.formatTimeCallback;
         this.edgeScrollWidth = params.edgeScrollWidth;
         this.bindInOut();
         this.render();


thijstriemstra:
> If anyone was relying on being able to set formatTimeCallback on a per-region basis, the this would be a breaking change.

Yea, -1 for me.

rdn32:
> > If anyone was relying on being able to set formatTimeCallback on a per-region basis, the this would be a breaking change.
> 
> Yea, -1 for me.

OK, I'll fix that

src/plugin/regions/index.js
```diff
@@ -201,7 +201,7 @@ export default class RegionsPlugin {
             params = {...params, minLength: this.regionsMinLength};
         }
 
-        const region = new this.wavesurfer.Region(params, this.util, this.wavesurfer);
+        const region = new this.wavesurfer.Region(params, this.params, this.util, this.wavesurfer);
```
katspaugh, line 204: Could we maybe just merge both params here instead of passing a new arg?
```suggestion
        const region = new this.wavesurfer.Region({
          formatTimeCallback: this.params.formatTimeCallback,
          edgeScrollWidth: this.params.edgeScrollWidth,
          ...params
        }, this.util, this.wavesurfer);
```

rdn32: That makes more sense than what I was suggesting.

Regarding `edgeScrollWidth`, I hadn't wanted to touch it, as I had an inkling that the code in the if block in `_onBackendCreated` was there to work around this issue, and I was a bit unsure what to make of it. Having looked at it some more, it seems to me that is what it's doing, but that the logic for setting a default value of `edgeScrollWidth` doesn't work unless all the regions are supplied in the plugin parameters - and I think it's worth sorting that out.

Would it be acceptable to have this in `_onBackendCreated`:
```
    if (!this.params.edgeScrollWidth) {
        this.params.edgeScrollWidth = this.wrapper.clientWidth * scrollWidthProportion;
    }
```
or should `this.params` be treated as read-only?

katspaugh: Sounds like it's less trivial than just passing it from params.

For the sake of this PR (since you just wanted to fix `formatTimeCallback`), maybe let's skip `edgeScrollWidth` for now.

If you feel like tackling that, a new PR would be nice.

rdn32: Sorry, I just raised an issue for `edgeScrollWidth`, and then realised that PR is "pull request", not "performance report".

src/plugin/regions/index.js
```diff
@@ -197,6 +197,10 @@ export default class RegionsPlugin {
             return null;
         }
 
+        if (!params.formatTimeCallback && this.params.formatTimeCallback) {
```
thijstriemstra, line 200: can you add a code comment explaining why this is here

thijstriemstra (changes requested):
see inline comments. also add a changelog entry.

PR #2298, opened 2021-06-07
markers plugin: custom HTML elements are now centered over the marker line

bknower-cogito:
…vesurfer

Previously, the markerElement displayed when passed as a parameter to the add marker function was not centered over the line displayed on the wavesurfer, but now it is.


thijstriemstra:
> Previously, the markerElement displayed when passed as a parameter to the add marker function was not centered over the line displayed on the wavesurfer, but now it is.

Can you show a before/after screenshot?

bknower:
> > Previously, the markerElement displayed when passed as a parameter to the add marker function was not centered over the line displayed on the wavesurfer, but now it is.
> 
> Can you show a before/after screenshot @bknower-cogito ? And a changelog entry is also missing.

Here's the before and after.
image

image

thijstriemstra (approved):
thanks!

thijstriemstra:
@bknower there's a merge conflict in CHANGES.md, can you check it?

PR #2299, opened 2021-06-08
plugins: prevent runtime errors when canvas context is null (setFillStyles, setFonts)

felguerez:
The [`HTMLCanvasElement.getContext()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext) method can return a drawing context on the canvas or `null` in some cases. Safari's Canvas implementation (macOS, v14.0.2) seems to cause runtime errors in the Timeline plugin:

`null is not an object (evaluating canvas.getContext("2d").fillStyle = fillStyle`

The error seems difficult to reproduce but it appears in our team's error reports a few hundred times per day.

This change will guard `setFillStyles` and setFonts` methods from attempting to access/mutate canvas properties when the object is undefined.

This shouldn't cause any breaking changes in the external or internal API.


thijstriemstra (changes requested):
thanks @felguerez 

and can you check if there are any other plugins that run into this problem?

thijstriemstra:
@felguerez build's failing

thijstriemstra (approved):
thanks!

PR #2265, opened 2021-05-07
docs: add upgrade instructions for 5.0

caugner:
Ref: https://github.com/katspaugh/wavesurfer.js/commit/a262e6993ffded2f06b65bb4ac373312e9e6aff8#commitcomment-50513905

PR #2305, opened 2021-06-14
release: 5.0.2

felguerez:
* docs: add upgrade instructions for 5.0 (#2265)
* update dev dependencies
* markers plugin: added support for custom marker elements (#2269)
* regions plugin: cursor orientation bugfix (#2293) 
* markers plugin: trigger click event on wavesurfer (#2287) 
* plugins: prevent runtime errors when canvas context is null (setFillStyles, setFonts) (#2299)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,7 @@
 wavesurfer.js changelog
 =======================
 
-x.x.x (unreleased)
+5.0.2 (07.15.2021)
 ------------------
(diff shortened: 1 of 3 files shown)


felguerez:
@thijstriemstra is there a release cadence this library follows or are releases created ad hoc?

thijstriemstra:
@felguerez take a look at the past releases, it's once every 3 months or so. Thanks for making the pull request but I'd rather roll it myself when the time is there. There are several plugin related PRs still open that are 99% done, and I'd like to incorporate those in a 5.1.0 release as well.

thijstriemstra:
You could help speeding up the release by providing code reviews on pull requests.

PR #2310, opened 2021-06-21
Marker time option typo

jnorris441:
had "id" instead of "time"...also specified "in seconds" since there was no telling if it was milliseconds or what


thijstriemstra (approved):
thanks!

PR #2311, opened 2021-06-22
improve error handling in ws.decodeArrayBuffer and WebAudio backend

### LINKED ISSUE #1473: Failed to execute 'stop' on 'AudioScheduledSourceNode': cannot call stop without calling start first.
### Wavesurfer.js version(s):
wavesurfer.js 2.0.6 (Fri Jun 15 2018 10:57:58 GMT+0200 (CEST))

### Browser version(s):
Google Chrome: 69.0.3497.100 (Official Build) (64-bit)

### Code that causes the issue:

        /**
         * Pauses the loaded audio.
         */

    }, {
        key: 'pause',
        value: function pause() {
            this.scheduledPause = null;

            this.startPosition += this.getPlayedTime();
            this.source && this.source.stop(0);

            this.setState(PAUSED);

            this.fireEvent('pause');
        }

-A check is needed before executing this.source && this.source.stop(0);, because due to timing problems it might happen that the audio is not started when the stop is called. 

###
To reproduce the issue, you need to load new audio files, then quickly start and stop them.

### LINKED ISSUE #2279: TypeError: Cannot read property 'decodeArrayBuffer' of null
### Wavesurfer.js version(s):

5.0.1

Nw.js / Electron

```js
		fs.readFile(file.replace('file://', ''), (err, buffer) => {
			if (err) {
				console.error(err)
				return
			}
			const blob = new window.Blob([new Uint8Array(buffer)])
			this.wavesurfer.loadBlob(blob)
		})
```

I am using WaveSurfer wrapped in a React component, I have a list of sounds and I can navigate through this list with keyboard shortcuts, when I go very fast through the list, the situation occurs that the wavesurfer's backend is gone and WaveSurfer crashes with: `TypeError: Cannot read property 'decodeArrayBuffer' of null`

 This happens over here: https://github.com/katspaugh/wavesurfer.js/blob/4360968534076ee7ed834e5c0c4ad48279f133a9/src/wavesurfer.js#L1542

I think we are missing a "this.isDestroyed" check over here: https://github.com/katspaugh/wavesurfer.js/blob/4360968534076ee7ed834e5c0c4ad48279f133a9/src/wavesurfer.js#L1326

Because loadBlob is using the FileReader which is asynchronous: https://github.com/katspaugh/wavesurfer.js/blob/4360968534076ee7ed834e5c0c4ad48279f133a9/src/wavesurfer.js#L1358

Here are some screenshots from me walking up the stack:

Screenshot 2021-05-26 at 15 46 19
Screenshot 2021-05-26 at 15 46 26
Screenshot 2021-05-26 at 15 46 32

vespakoen:
Fixes

`Failed to execute 'stop' on 'AudioScheduledSourceNode'`

and

`Cannot read property 'decodeArrayBuffer' of null`

(both timing / state issues)

Adds the necessary checks to stop errors from happening.

none

none

### Todos/Notes:

none

fixes #2279
fixes #1473

### DIFF
--- a/src/webaudio.js
+++ b/src/webaudio.js
@@ -705,7 +705,12 @@ export default class WebAudio extends util.Observer {
         try {
             this.source && this.source.stop(0);
         } catch (err) {
-            // no-op, see https://github.com/katspaugh/wavesurfer.js/issues/1473
+            // Calling stop can throw the following 2 errors:
+            // - RangeError (The value specified for when is negative.)
+            // - InvalidStateNode (The node has not been started by calling start().)
+            // We can safely ignore both errors, because:
+            // - The range is surely correct
+            // - The node might not have been started yet, in which case we just want to carry on without causing any trouble.
         }
 
         this.setState(PAUSED);
--- a/src/webaudio.js
+++ b/src/webaudio.js
@@ -702,7 +702,11 @@ export default class WebAudio extends util.Observer {
         this.scheduledPause = null;
 
         this.startPosition += this.getPlayedTime();
-        this.source && this.source.stop(0);
+        try {
+            this.source && this.source.stop(0);
+        } catch (err) {
+            // no-op, see https://github.com/katspaugh/wavesurfer.js/issues/1473
+        }
 
         this.setState(PAUSED);
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 wavesurfer.js changelog
 =======================
 
+x.x.x (unreleased)
+------------------
+- Fixed unhandled `Failed to execute 'stop' on 'AudioScheduledSourceNode'` error (#1473)
+- Fixed unhandled `Cannot read property 'decodeArrayBuffer' of null` error (#2279)
+
 5.1.0 (20.06.2021)
 ------------------
 - Markers plugin:
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -1538,19 +1538,21 @@ export default class WaveSurfer extends util.Observer {
      * @param {function} callback The function to call on complete
      */
     decodeArrayBuffer(arraybuffer, callback) {
-        this.arraybuffer = arraybuffer;
-        this.backend.decodeArrayBuffer(
-            arraybuffer,
-            data => {
-                // Only use the decoded data if we haven't been destroyed or
-                // another decode started in the meantime
-                if (!this.isDestroyed && this.arraybuffer == arraybuffer) {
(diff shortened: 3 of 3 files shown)


src/webaudio.js
```diff
@@ -702,7 +702,11 @@ export default class WebAudio extends util.Observer {
         this.scheduledPause = null;
 
         this.startPosition += this.getPlayedTime();
-        this.source && this.source.stop(0);
+        try {
+            this.source && this.source.stop(0);
+        } catch (err) {
+            // no-op, see https://github.com/katspaugh/wavesurfer.js/issues/1473
```
thijstriemstra, line 708: can you add a little more info so you only have to visit the ticket if you need all details

thijstriemstra (changes requested):
thanks! changes make sense. I preferred not to add a isDestroyed check everywhere but for this method I certainly don't mind.

can you also add a changelog entry? a new section with `x.x.x (unreleased)` is probably needed.

thijstriemstra:
Also; bonus points for adding a unit test that would fail without this change.

vespakoen:
I added the items to the changelog and a (hopefully) better comment.

I tried adding tests for these but it's pretty hard, since I don't know how I'd verify that an error threw.

I guess the test case would be something like:

```

    /** @test {WaveSurfer#destroy} */
    it('destroy while loading', function(done) {
        manualDestroy = true;

        wavesurfer.once('destroy', function() {
            TestHelpers.removeElement(element);
            done();
        });
        wavesurfer.load(TestHelpers.EXAMPLE_FILE_PATH);
        wavesurfer.destroy();

        // EXPECT NO THROWN ERRORS SOMEHOW?
    });
```

For the other issue it seems to happen when quickly starting / stopping a sound, perhaps something like:

```
    it('can quickly toggle', function() {
        wavesurfer.load(TestHelpers.EXAMPLE_FILE_PATH);
        wavesurfer.once('ready', () => {
            wavesurfer.play();
            wavesurfer.pause();
            wavesurfer.play();
            wavesurfer.pause();
            wavesurfer.play();
            wavesurfer.pause();
        });
    });
```

But it probably won't work all the time and is not elegant.

Basically, these are edge cases that happen in a rare situation that is hard to re-create in a test.

thijstriemstra (approved):
thanks!

vespakoen:
Thank you too!

@thijstriemstra Would it be possible to do a minor release for this change? I'd like to include it in an update to one of my projects tomorrow.

If not, that's also fine ;)

PR #2316, opened 2021-07-04
minimap plugin: add touch events

jnorris441:
The minimap should be draggable on a phone

### DIFF
--- a/src/plugin/minimap/index.js
+++ b/src/plugin/minimap/index.js
@@ -307,27 +307,35 @@ export default class MinimapPlugin {
         }
 
         if (this.params.showOverview) {
-            this.overviewRegion.addEventListener('mousedown', event => {
-                this.draggingOverview = true;
-                relativePositionX = event.layerX;
-                positionMouseDown.clientX = event.clientX;
-                positionMouseDown.clientY = event.clientY;
-            });
+            let clientX = (event)=> event.changedTouches ? event.changedTouches[0].clientX : event.clientX
+        	let clientY = (event)=> event.changedTouches ? event.changedTouches[0].clientY : event.clientY
+			let layerX = (event)=> clientX(event) - event.target.getBoundingClientRect().x
 
-            this.drawer.wrapper.addEventListener('mousemove', event => {
+        	let mouseDown = event => {
+                this.draggingOverview = true;
+                relativePositionX = layerX(event);
+                positionMouseDown.clientX = clientX(event);
+                positionMouseDown.clientY = clientY(event);               
+            }
+            this.overviewRegion.addEventListener('mousedown', mouseDown);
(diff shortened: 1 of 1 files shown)


src/plugin/minimap/index.js
```diff
@@ -307,35 +307,46 @@ export default class MinimapPlugin {
...
                 this.draggingOverview = true;
-                relativePositionX = event.layerX;
-                positionMouseDown.clientX = event.clientX;
-                positionMouseDown.clientY = event.clientY;
-            });
+                relativePositionX = layerX(event);
+                positionMouseDown.clientX = clientX(event);
+                positionMouseDown.clientY = clientY(event);
+            };
+            this.overviewRegion.addEventListener('mousedown', mouseDown);
```
thijstriemstra, line 320: these event listeners should be cleaned up as well.

src/plugin/minimap/index.js
```diff
@@ -307,35 +307,46 @@ export default class MinimapPlugin {
...
                     this.moveOverviewRegion(
-                        event.clientX -
+                        clientX(event) -
                             this.drawer.container.getBoundingClientRect().left -
                             relativePositionX
                     );
                 }
-            });
+            };
+            this.drawer.wrapper.addEventListener('mousemove', mouseMove);
```
thijstriemstra, line 332: these event listeners should be cleaned up as well.

src/plugin/minimap/index.js
```diff
@@ -307,35 +307,46 @@ export default class MinimapPlugin {
...
                 ) {
                     seek = true;
                     this.draggingOverview = false;
                 } else if (this.draggingOverview) {
                     seek = false;
                     this.draggingOverview = false;
                 }
-            });
+            };
+            this.drawer.wrapper.addEventListener('mouseup', mouseUp);
```
thijstriemstra, line 347: these event listeners should be cleaned up as well.

thijstriemstra (changes requested):

src/plugin/minimap/index.js
jnorris441: Makes sense. What is the best way to do that? How are the existing mousedown/mousemove/mouseup cleaned up?

thijstriemstra: take a look at other plugins like regions or timeline.

thijstriemstra: and others in `destroy`: https://github.com/katspaugh/wavesurfer.js/blob/bfa9c76040058ac59bab19e7bd9d347521c6e425/src/plugin/minimap/index.js#L181

thijstriemstra:
@jnorris441 can you take a look at this PR this week, I'd like to release a new wavesurfer.js with this fix included if possible.

thijstriemstra:
no feedback in a year, closing.

PR #2333, opened 2021-08-02
timeline plugin: fix null is not an object in fillText

felguerez:
Followup to PR #2299, this is an additional instance of the same pattern that caused `null is not an object` elsewhere

The [`HTMLCanvasElement.getContext()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext) method can either return a drawing context on the canvas or `null`. Safari's Canvas implementation (v14) seems to cause runtime errors in the Timeline plugin:

`null is not an object (evaluating context.canvas)`

The error seems difficult to reproduce but it appears in our team's error reports a few hundred times per day.

This change will guard `fillText` from attempting to access canvas as an object if it doesn't exist.

This shouldn't cause any breaking changes in the external or internal API.


thijstriemstra (changes requested):
thanks. can you add a changelog entry?

thijstriemstra:
ping @felguerez

felguerez:
@thijstriemstra sorry to drop the ball moving this one forward, many thanks for adding to the changelog and releasing  🙏

nahukas:
Hi all I still have problems with large files on Safari, the wave and the audio is cut somewhere in the middle. I don't know if this supposed to fix this issue but in my case it didn't work out. Thank you all

thijstriemstra:
@nahukas no, that sounds unrelated to this pull request.

PR #2339, opened 2021-08-10
regions plugin: add direction and action to region-updated event params

msuhov:
Add new `direction` and `action` fields to `region-updated` event params. That is going to help to understand in what direction is moving or resizing a region and what exactly happened.
`direction` can be in 3 states: `null`, `start` and `end`
`action` can be `drag` or `resize`

```
wavesurfer.on('region-updated', function(region, { direction, action }) {
    console.log(region, direction, action);
});
```

No breaking changes

No breaking changes

### DIFF
--- a/src/plugin/regions/region.js
+++ b/src/plugin/regions/region.js
@@ -87,7 +87,7 @@ export class Region {
     }
 
     /* Update region params. */
-    update(params) {
+    update(params, eventParams) {
         if (params.start != null) {
             this.start = Number(params.start);
         }
@@ -125,7 +125,7 @@ export class Region {
 
         this.updateRender();
         this.fireEvent('update');
-        this.wavesurfer.fireEvent('region-updated', this);
+        this.wavesurfer.fireEvent('region-updated', this, eventParams);
     }
 
     /* Remove a single region. */
@@ -708,10 +708,33 @@ export class Region {
             delta = this.start * -1;
         }
 
+        const eventParams = {
+            direction: this._getDragDirection(delta),
+            action: 'drag'
+        };
+
         this.update({
             start: this.start + delta,
             end: this.end + delta
-        });
+        }, eventParams);
+    }
+
+    /**
+     * Returns the direction of dragging region based on delta
+     * Negative delta means region is moving to the left
+     * Positive - to the right
+     * For zero delta the direction is not defined
+     * @param {number} delta Drag offset
+     * @returns {string|null} Direction 'left', 'right' or null
+     */
+    _getDragDirection(delta) {
+        if (delta < 0) {
+            return 'left';
+        }
+        if (delta > 0) {
+            return 'right';
+        }
+        return null;
     }
 
     /**
@@ -724,6 +747,11 @@ export class Region {
      */
     onResize(delta, direction) {
         const duration = this.wavesurfer.getDuration();
+        const eventParams = {
+            action: 'resize',
+            direction: direction === 'start' ? 'right' : 'left'
+        };
+
         if (direction === 'start') {
             // Check if changing the start by the given delta would result in the region being smaller than minLength
             // Ignore cases where we are making the region wider rather than shrinking it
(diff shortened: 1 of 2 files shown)


thijstriemstra (changes requested):
Can you also document the new parameter here: https://github.com/katspaugh/wavesurfer.js/blob/1e31868f1958d87423a597e59b83d39e1d295a77/src/plugin/regions/index.js#L20

And a changelog entry?

thijstriemstra:
ps. why is it called `updateDirection`? Maybe something like `currentDirection` or `activeDirection`?

msuhov:
> ps. why is it called `updateDirection`? Maybe something like `currentDirection` or `activeDirection`?

thanks for the feedback, `activeDirection` makes more sense, gonna change it soon

msuhov:
> Can you also document the new parameter here:
> 
> https://github.com/katspaugh/wavesurfer.js/blob/1e31868f1958d87423a597e59b83d39e1d295a77/src/plugin/regions/index.js#L20
> 
> And a changelog entry?

The thing is that `activeDirection` is not a param, it's a field of Region instance. So I'm not sure about adding it there

thijstriemstra:
If it's not a param but a public property, there's no need to add it to `params` here: https://github.com/katspaugh/wavesurfer.js/pull/2339/files#diff-e7155624d86da0e6e00723224e8d2069806b92c756b6b1ae11af3168e866e82cR127

src/plugin/regions/region.js
```diff
@@ -708,7 +713,16 @@ export class Region {
             delta = this.start * -1;
         }
 
+        let activeDirection;
```
thijstriemstra, line 716: not sure why you're redeclaring it here?

msuhov: Ok, you are right, I can simply update the `this.activeDirection` without `update`

sundayz:
It's a good idea but I think you should rethink how it is used.

- "left" and "right" are directions, "start" and "end" are not. I understand why it is like that, but it will be confusing for new users what that means.
- Currently the direction is only set when onResize runs, but not when moving the region?
- When you are using `activeDirection` you probably want to use it inside of an event. So the direction should be an event parameter, not a property on the region object.
- Also, right now `activeDirection` remains set to the last value. So if I drag the region, wait 5 seconds, now it's still set to that direction, even though I am not moving the region. Does that make sense? Maybe?

I think if I was using wavesurfer for the first time I would expect to be able to do something like this:
```
wavesurfer.on('region-updated', function (region, direction) {
    console.log(`Region ${region} is moving in direction ${direction}`);
});
```

2nd parameter can be an object so we can add more things in the future, and you can use destructuring like `{ direction }`

msuhov:
@sundayz 
Thank you for your feedback!

1. Agree with left/right naming, but I was trying to make as less changes as possible. Since resize already had start/end, I've just reused it
2. It's updated on both events move and resize
3. It's not really a property of Region, you are right. Happy to change it

What do you think about the following:

```
// Region
update(params, eventParams = {}) {
  ...

  this.wavesurfer.fireEvent('region-updated', this, eventParams)
}

onResize(delta, direction) {
   if (direction === 'start') {
    ...
    this.update({...}, { direction: 'right' })
   } else {
    ...
    this.update({...}, { direction: 'left' })
   }
}

onDrag(delta) {
  ...
  const eventParams = {
     direction: this._getDragDirection(delta)
   }

  this.update({...}, eventParams)
}

_getDragDirection (delta) {
   if (delta < 0) {
        return 'right';
   } else if (delta > 0) {
        return 'left';
  }
  return null
}

```

msuhov:
@sundayz I don't see any cons about adding `action` param to the event. Will do, thanks

src/plugin/regions/region.js
```diff
@@ -708,10 +708,25 @@ export class Region {
...
+        };
+
         this.update({
             start: this.start + delta,
             end: this.end + delta
-        });
+        }, eventParams);
+    }
+
+    _getDragDirection(delta) {
```
thijstriemstra, line 722: this function needs documentation.

msuhov: done, thanks

sundayz (approved):

thijstriemstra (approved):

PR #2341, opened 2021-08-14
add support of sample step

zouchengzhuo:
add support of sample step for webaudio.js
When displaying a long audio waveform in "scrollParent:false" mode, some information will be lost due to the strategy, which makes the waveform look different from that on some PC software. With this option,  when the user passes in 1, the waveform display effect is the same as that on the PC software. This option sacrifices some of the performance for an accurate presentation.

no

no

### DIFF
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -87,6 +87,7 @@ import MediaElementWebAudio from './mediaelement-webaudio';
  * `'audio'|'video'` ('video' only for `MediaElement`)
  * @property {number} minPxPerSec=20 Minimum number of pixels per second of
  * audio.
+ * @property {number} sampleStep=null If set, use the sampleStep in webaudio. Default behavior is to use ~~(sampleSize / 10).
  * @property {boolean} normalize=false If true, normalize by the maximum peak
  * instead of 1.0.
  * @property {boolean} partialRender=false Use the PeakCache to improve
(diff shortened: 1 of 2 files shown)


thijstriemstra:
@zouchengzhuo thanks for this fix. Could you see if there are any bug reports in wavesurfer that would be solved with this pr?

zouchengzhuo:
> @zouchengzhuo thanks for this fix. Could you see if there are any bug reports in wavesurfer that would be solved with this pr?

no, probably because it was an improvement rather than a bug fix.

image
image

from top to bottom is:

- wavesurfer before set sample step to 1
- wavesurfer after set sample step to 1
- adobe audition

entonbiba:
@zouchengzhuo can you explain the reasoning for doing the following change below, should this only happen if the `sampleSize` is specified?

```

if (value > max) {
    max = value;
} else if (value < min) {
    min = value;
}

```

zouchengzhuo:
> @zouchengzhuo can you explain the reasoning for doing the following change below, should this only happen if the `sampleSize` is specified?
> 
> ```
> 
> if (value > max) {
>     max = value;
> } else if (value < min) {
>     min = value;
> }
> ```

Sorry I didn't see it until now, when value is bigger than max, it can't be smaller than min. So we can use 'else' to avoid unnecessary 'if' logic, it works for all scene.

PR #2357, opened 2021-09-10
plugins: support vertical mode in cursor and minimap plugins

Ivaka:
**Title:** Please make sure the name of your PR is as descriptive as possible (Describe the feature that is introduced or the bug that is being fixed).

Use oriented event and DOM targets in cursor and minimap plugins, so that they work properly in vertical mode

no

no

### DIFF
--- a/example/minimap-vertical/index.html
+++ b/example/minimap-vertical/index.html
@@ -1,0 +1,147 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+        <title>wavesurfer.js | Vertical waveform</title>
+
+        <link href="data:image/gif;" rel="icon" type="image/x-icon" />
+
+        <!-- Bootstrap -->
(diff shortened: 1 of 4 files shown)


thijstriemstra (changes requested):
thanks! I tested the cursor plugin, which looks good. But the minimap plugin doesn't seem to work when specifying `vertical: true` but I might be missing some styling. Can you add a new `minimap-vertical` example that demonstrates vertical usage?

Ivaka:
> thanks! I tested the cursor plugin, which looks good. But the minimap plugin doesn't seem to work when specifying `vertical: true` but I might be missing some styling. Can you add a new `minimap-vertical` example that demonstrates vertical usage?

Thanks for taking the time to review and test this @thijstriemstra !

You are probably missing proper styling on the waveform and minimap elements:
```css
#wave-minimap, #waveform {
    display: inline-flex;
    height: 400px;
}
```

I've added a new `minimap-vertical`example

thijstriemstra (approved):
thanks!

PR #2367, opened 2021-09-28
Add option hideCursor

### LINKED ISSUE #2366: hideCursor
Hello, I saw that there is a hideCursor() function but it doesn't work.

Is it possible to have an example to use it?
Thanks in advance.

entonbiba:
Add option `hideCursor`.
- hides the cursor if set to true, by default it will always show.

fixes #2366

### DIFF
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -69,6 +69,7 @@ export default class Drawer extends util.Observer {
         if (this.params.fillParent || this.params.scrollParent) {
             this.style(this.wrapper, {
                 width: '100%',
+                cursor: this.params.hideCursor ? 'hidden' : 'auto',
                 overflowX: this.params.hideScrollbar ? 'hidden' : 'auto',
                 overflowY: 'hidden'
             });
--- a/src/drawer.js
+++ b/src/drawer.js
@@ -69,7 +69,7 @@ export default class Drawer extends util.Observer {
         if (this.params.fillParent || this.params.scrollParent) {
             this.style(this.wrapper, {
                 width: '100%',
-                cursor: this.params.hideCursor ? 'hidden' : 'auto',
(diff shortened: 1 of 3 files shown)


src/drawer.js
```diff
@@ -69,6 +69,7 @@ export default class Drawer extends util.Observer {
         if (this.params.fillParent || this.params.scrollParent) {
             this.style(this.wrapper, {
                 width: '100%',
+                cursor: this.params.hideCursor ? 'hidden' : 'auto',
```
katspaugh, line 72: As per [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor), It seems it should be `none` instead of `hidden`?

katspaugh:
Please don't forget to add it [in the docs](https://wavesurfer-js.org/docs/options.html) afterwards.

src/wavesurfer.js
```diff
@@ -270,6 +270,7 @@ export default class WaveSurfer extends util.Observer {
         forceDecode: false,
         height: 128,
         hideScrollbar: false,
+        hideCursor: false,
```
thijstriemstra, line 273: also needs to be documented at top of this file.

and changelog entry is missing.

thijstriemstra (changes requested):

src/drawer.js
entonbiba: yes

entonbiba:
updated,

entonbiba:
> Please don't forget to add it [in the docs](https://wavesurfer-js.org/docs/options.html) afterwards.

ready once this is merged
#2372

katspaugh (approved):
Cool, thanks!

thijstriemstra (approved):

PR #2369, opened 2021-09-28
add accessibility features

entonbiba:
Announce event changes as well as any content updates by the various plugins and events in order for the changes to be announced to screen readers.

to do:
- [x] add check to see if the element already exists
- [x] add function that creates the announcement element
- [x] add function that updates the announcement element
- [x] add function that clears the announcement element
- [x] add announcements for various events

### DIFF
--- a/src/util/accessibility.js
+++ b/src/util/accessibility.js
@@ -14,15 +14,15 @@ export class Accessibility {
     this.announcement.setAttribute('aria-live', 'polite');
 
     style(this.announcement, {
-      position: 'absolute',
-      width: '1px',
-      height: '1px',
-      padding: '0',
-      margin: '-1',
-      overflow: 'hidden',
-      clip: 'rect(0,0,0,0)',
-      white-space: 'nowrap',
-      border: '0'            
+      position: "absolute",
+      width: "1px",
+      height: "1px",
+      padding: "0",
+      margin: "-1",
+      overflow: "hidden",
+      clip: "rect(0,0,0,0)",
+      "white-space": "nowrap",
+      border: "0"
     });
 
     this.updateAnnouncementText('Loading waveform');
--- a/src/util/accessibility.js
+++ b/src/util/accessibility.js
@@ -7,41 +7,41 @@
 import style from './style';
 
 export class Accessibility {
-  constructor() {
-    this.announcement = document.body.appendChild(
-      document.createElement('div')
-    );
-    this.announcement.setAttribute('aria-live', 'polite');
+    constructor() {
+        this.announcement = document.body.appendChild(
+            document.createElement('div')
+        );
+        this.announcement.setAttribute('aria-live', 'polite');
 
-    style(this.announcement, {
-      position: "absolute",
-      width: "1px",
-      height: "1px",
-      padding: "0",
-      margin: "-1",
-      overflow: "hidden",
-      clip: "rect(0,0,0,0)",
-      "white-space": "nowrap",
-      border: "0"
-    });
+        style(this.announcement, {
+            position: "absolute",
+            width: "1px",
+            height: "1px",
+            padding: "0",
+            margin: "-1",
+            overflow: "hidden",
+            clip: "rect(0,0,0,0)",
+            "white-space": "nowrap",
+            border: "0"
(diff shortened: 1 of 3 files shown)


entonbiba:
provide your feedback on the changes above and any optimization to the util code.

src/util/accessibility.js
```diff
@@ -0,0 +1,51 @@
...
+ * Add accessibility features to help support announcing events and content changes for screen readers.
+ *
+ * @since 5.2.0
+ */
+
+/**
+ * Create the announcement element
+ */
+export default function createAnnouncementElement() {
+    var announcement = document.getElementById("announce-waveform");
```
msuhov, line 11: Thanks for your solution!
In general I see couple of issue with it:
- We can use `const` instead of `var`
- Do not repeat yourself with `document.getElementById("announce-waveform")`
- Use existing style utils for styling, to not repeat `Object.keys(styles).forEach..`

Also I think the number of `document.getElementById("announce-waveform")` calls, which are quite expensive for the performance. I'd suggest to use class instance or closures. Since almost everything in the project is done by using classes, I'd go for it:

```
import style from './style';

export class Accessibility {
  constructor() {
    this.announcement = document.body.appendChild(
      document.createElement('div')
    );
    this.announcement.setAttribute('aria-live', 'polite');

    style(this.announcement, {
      position: 'absolute',
      width: '1px',
      height: '1px',
      ...other,
    });

    this.updateAnnouncementText('Loading waveform');
  }

  updateAnnouncementText(text) {
    this.announcement.textContent = text;
  }

  clearAnnouncementText() {
    this.announcement.textContent = '';
  }
}
```

And then in WaveSurfer class:

```
class WaveSurfer {
  constructor() {
    this.accessibility = new Accessibility();
  }

  play() {
    this.accessibility.updateAnnouncementText('audio playing');
  }
}
```

sundayz:
If the Wavesurfer instance is destroyed, the announcer element should be removed too, otherwise they are never cleaned up.

Also, I feel like this could be implemented nicely with a plugin? 😃

thijstriemstra:
> Also, I feel like this could be implemented nicely with a plugin?

+1

src/util/accessibility.js
entonbiba: yea using classes is better, I'll update it

entonbiba:
> If the Wavesurfer instance is destroyed, the announcer element should be removed too, otherwise they are never cleaned up.
> 
> Also, I feel like this could be implemented nicely with a plugin? 😃
> 
> Edit: Actually it looks like the function is only called once so even if you have 2 wavesurfers there is only one announcer element. Still though it never goes away, so maybe it's good to let people choose if they should enable this or not.

true we can have it as an option, however I think making it as part of the wavesurfer code by default is better since being accessible for screen readers is a priority these days it ensures everyone has a meaningful experience from a accessibility and user experience. In addition to these announcements, I'll review the default styles as well to see what enhancements can be made to make it have better contrast.

thijstriemstra (changes requested):
I would suggest creating a new event (?) that fires when you need accessibility features (when the option is enabled), which a new plugin can handle (including translation of text etc). This way core wavesurfer isn't "bloated" with additional processing of stuff you might not need (so i would also suggest the option is disabled by default, no events are triggered. You need to enable the plugin to start trigger of the new event so the plugin can handle it.

> however I think making it as part of the wavesurfer code by default is better since being accessible for screen readers is a priority these days it ensures everyone has a meaningful experience from a accessibility and user experience.

Many wavesurfer projects might want to provide their own implementation or disable it completely. We shouldn't force things on users and keep core code lightweight as possible.

sundayz:
Yes, this should definitely be a plugin, think about all the functionality that a user might want:

- Translations
- Disable certain announcements
- Supply your own live region where the announcements appear

Sooner or later someone will ask for those, and it's much better to separate all that code in a plugin rather than directly under the Wavesurfer class.

thijstriemstra:
> Sooner or later someone will ask for those, and it's much better to separate all that code in a plugin rather than directly under the Wavesurfer class.

No feedback in a year, closing.

PR #2377, opened 2021-10-10
Add "Camelia Studio" in the Projects section of the documentation

julienbeisel:

I added the Camelia Studio project in the projects section of the documentation. We use it to display the waveform of uploaded songs and select an extract.

Thanks for maintaining wavesurfer.js, we love this project :)

No breaking changes

No breaking changes


thijstriemstra:
thanks!

PR #2390, opened 2021-10-31
regions plugin: fix region-in and region-out events

### LINKED ISSUE #2140: Regions plugin: 'out' event incorrectly emitted
### Wavesurfer.js version(s):
4.1.1

Firefox Windows 10

https://codepen.io/sundays/pen/VwKzmmL

See codepen.

1. Click inside region
2. Click play, then quickly pause when still inside the region.
3. Click outside the region to seek.
4. Click play. Region emits the 'out' event, but it should not do it in this case.

Maybe need to unbind the in/out event handlers when seeking outside the region.

### LINKED ISSUE #2307: Region events "region-in" and "region-out" fired in wrong order
## Description
Incorrect event firing order.
When clicking on region in reverse order, evens fired in wrong order "region-in" => "region-out".
When clicking in direct order, evens fired in correct order "region-out" => "region-in".

## Steps to reproduce
1)Have at least two regions (i.e. №1 and №2).
2) Start to play audio\video.
3) When audio\video entering in region №2, click in region №1.

## Results
Events fired in "backward" order: "region-in" => "region-out".
### Expected
Events fired in order: "region-out" => "region-in".

## Versions
Video.js 7.12.3
slighly modified parse-headers 2.0.2 <https://github.com/kesla/parse-headers/>
videojs-wavesurfer @version 3.8.0
wavesurfer.js regions plugin 4.0.1 (2020-06-24) https://github.com/katspaugh/wavesurfer.js
wavesurfer.js timeline plugin 4.0.1 (2020-06-24) https://github.com/katspaugh/wavesurfer.js
wavesurfer.js cursor plugin 4.0.1 (2020-06-24) https://github.com/katspaugh/wavesurfer.js
Reproduced also on latest version.
### Browsers
Presumably all. Tested on Chrome 91.0.4472.106
### OSes
Presumably all. Tested on Windows 10 

Region events "region-in" and "region-out" fired in wrong order

muhammad-sayed-mahdy:
- Make firing region-out events run before firing region-in events regardless of the order of the creation of regions. (Fixes #2307) 
- Fire region-out and region-in when seeking while the audio is paused. (Fixes #2140)

### TODOs and Notes
I know that handling the order of running the events should not be the responsibility of the observer and the code is ugly, but that what I came up with, but you got the idea anyway, and the code can be modified.

#2307 
#2140

### DIFF
--- a/src/util/observer.js
+++ b/src/util/observer.js
@@ -136,9 +136,24 @@ export default class Observer {
         }
 
         const handlers = this.handlers[event];
-        handlers &&
-            handlers.forEach(fn => {
-                fn(...args);
-            });
+        // run fireRegionOut handlers first
+        if (handlers) {
+            if (event == "audioprocess") {
+                handlers.forEach(fn => {
+                    if (fn.name == "fireRegionOut") {
+                        fn(...args);
+                    }
+                });
+                handlers.forEach(fn => {
+                    if (fn.name != "fireRegionOut") {
+                        fn(...args);
+                    }
+                });
+            } else {
+                handlers.forEach(fn => {
+                    fn(...args);
+                });
+            }
+        }
     }
 }
--- a/src/plugin/regions/region.js
+++ b/src/plugin/regions/region.js
@@ -325,9 +325,10 @@ export class Region {
         this.firedIn = false;
         this.firedOut = false;
 
-        const onProcess = (time) => {
-            let start = Math.round(this.start * 10) / 10;
-            let end = Math.round(this.end * 10) / 10;
+        let start = Math.round(this.start * 10) / 10;
+        let end = Math.round(this.end * 10) / 10;
+
+        const fireRegionOut = (time) => {
             time = Math.round(time * 10) / 10;
 
             if (
@@ -340,6 +341,11 @@ export class Region {
                 this.fireEvent('out');
                 this.wavesurfer.fireEvent('region-out', this);
             }
+        };
+
+        const fireRegionIn = (time) => {
+            time = Math.round(time * 10) / 10;
+
             if (!this.firedIn && start <= time && end > time) {
                 this.firedIn = true;
                 this.firedOut = false;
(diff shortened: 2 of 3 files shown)


src/util/observer.js
```diff
@@ -136,9 +136,24 @@ export default class Observer {
...
-        handlers &&
-            handlers.forEach(fn => {
-                fn(...args);
-            });
+        // run fireRegionOut handlers first
+        if (handlers) {
+            if (event == "audioprocess") {
+                handlers.forEach(fn => {
+                    if (fn.name == "fireRegionOut") {
+                        fn(...args);
```
thijstriemstra, line 144: you shouldn't put plugin related stuff into classes that are used by core wavesurfer.

thijstriemstra (changes requested):
see inline comment.

src/util/observer.js
muhammad-sayed-mahdy: I have another approach in mind, We can add a property `order` to the handler functions, and make the observer `handlers` a priority queue instead of an array, and sort that priority queue on the `order` and in case of equal order, the handlers get executed in the default order of attaching thme.

muhammad-sayed-mahdy: @thijstriemstra When I wrote this PR I knew that it will get rejected, but I just wanted to show the general idea to solve those 2 issues, now I want your input to decide how to do that.

thijstriemstra: your suggestion sounds ok to me. Would probably be a good idea to put this enhancement in a separate PR that only touches core code.

muhammad-sayed-mahdy: @thijstriemstra Can I use a priority queue dependancy like  [this](https://www.npmjs.com/package/js-priority-queue), or should it be implemented in util and tested ?

thijstriemstra: using a library is probably best as long as it's maintained actively and used a lot. Not a fan of the one you linked based on these criteria.

thijstriemstra:
no feedback in a year, closing.

PR #2402, opened 2021-11-10
regions plugin: ensure edgeScrollWidth is set

### LINKED ISSUE #2401: Edgescroll doesn't work properly for regions created via addRegion or drag selection
Edgescroll works well with regions that are specified in the region plugin params, but not so well for regions created later. Here's an example which shows the problem:
```
<!DOCTYPE html>
<html>
        <script src="https://unpkg.com/wavesurfer.js/dist/wavesurfer.js"></script>
        <script src="https://unpkg.com/wavesurfer.js/dist/plugin/wavesurfer.regions.js"></script>

    </head>

    <body>
        <div style="width: 200px;">
            <div id="waveform">
            </div>
        </div>

        <script>
document.addEventListener('DOMContentLoaded', function() {
    wavesurfer = WaveSurfer.create({
        container: '#waveform',
        height: 100,
        pixelRatio: 1,
        scrollParent: true,
        normalize: true,
        backend: 'MediaElement',
        plugins: [
            WaveSurfer.regions.create({
                regions: [{start: 1, end: 2, color: '#f00'}]
            })
        ]
    });

    wavesurfer.load('https://wavesurfer-js.org/example/media/demo.wav');

    wavesurfer.on('ready', function() {
        wavesurfer.addRegion({start: 3, end: 4, color: '#00f'});
        wavesurfer.enableDragSelection({color: '#0f0'});
    });
});
        </script>
    </body>
</html>
```

Initially there is a red region and a blue region; and a green region can be created by dragging on the waveform. All three regions can be moved and resized, but only with the red region does this cause the waveform to scroll.

This can be worked around by passing a value for `edgeScrollWidth` in the parameters passed to `addRegion` and `enableDragSelection`.

Wavesurfer.js version: 5.2.0
Browser and operating system version: Firefox 85.0.1, Ubuntu 18.04

rdn32:
**Title:** Fix edgescroll for regions created after region plugin

Looking for a value for edgeScrollWidth for a region, either from plugin params or a default value, was only happening for regions specified in the plugin params; it now happens for all regions. The consequence of it not being set was that the logic for doing edgescrolling was doing calculations with `undefined`, and so never doing any scrolling.

None

None

fixes #2401.

The problem was noticed in the course of pull request #2294.

### DIFF
--- a/src/plugin/regions/index.js
+++ b/src/plugin/regions/index.js
@@ -129,15 +129,14 @@ export default class RegionsPlugin {
         this.wavesurfer.Region = Region;
 
         // By default, scroll the container if the user drags a region
-        // within 5% of its edge
+        // within 5% (based on its initial size) of its edge
         const scrollWidthProportion = 0.05;
         this._onBackendCreated = () => {
             this.wrapper = this.wavesurfer.drawer.wrapper;
             this.orientation = this.wavesurfer.drawer.orientation;
+            this.defaultEdgeScrollWidth = this.wrapper.clientWidth * scrollWidthProportion;
             if (this.params.regions) {
                 this.params.regions.forEach(region => {
-                    region.edgeScrollWidth = this.params.edgeScrollWidth ||
-                        this.wrapper.clientWidth * scrollWidthProportion;
                     this.add(region);
                 });
             }
@@ -196,6 +195,11 @@ export default class RegionsPlugin {
         if (this.wouldExceedMaxRegions()) {
             return null;
         }
+
+        params = {
+            edgeScrollWidth: this.params.edgeScrollWidth || this.defaultEdgeScrollWidth,
+            ...params
+        };
 
         // Take formatTimeCallback from plugin params if not already set
         if (!params.formatTimeCallback && this.params.formatTimeCallback) {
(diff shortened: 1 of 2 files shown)


src/plugin/regions/index.js
```diff
@@ -134,10 +134,9 @@ export default class RegionsPlugin {
         this._onBackendCreated = () => {
             this.wrapper = this.wavesurfer.drawer.wrapper;
             this.orientation = this.wavesurfer.drawer.orientation;
+            this.defaultEdgeScrollWidth = this.wrapper.clientWidth * scrollWidthProportion;
```
katspaugh, line 137: Will this still work if the container is resized? (E.g. when resizing the window)
It could be fine if it doesn't, maybe we can just add a comment.

src/plugin/regions/index.js
```diff
@@ -197,6 +196,11 @@ export default class RegionsPlugin {
             return null;
         }
 
+        params = {
+            edgeScrollWidth: this.params.edgeScrollWidth || this.defaultEdgeScrollWidth,
```
katspaugh, line 200: Looks good, I like that you found a way not to modify `this.params`.

src/plugin/regions/index.js
rdn32: I've added a comment.

An alternative approach would have been for this calculation to be done in the relevant bit of onMove in region.js if no value for edgeScrollWidth had been set. That would guarantee it would be up-to-date, although I'd worry that getting the value of clientWidth out of an HTML element might be one of those things that's surprisingly expensive on some browsers, and the effect of doing that for every mouse movement (whilst a region was being dragged) would be noticeable.

thijstriemstra (changes requested):
can you add a reference to this pr in the changelog entry for #2401

PR #2404, opened 2021-11-13
Support order for observer event handlers

muhammad-sayed-mahdy:
- Make `handlers` a sorted set instead of an array.
- Sort the handlers based on their order if exists or the order of insertion.

[#2390](https://github.com/katspaugh/wavesurfer.js/pull/2390#discussion_r740429675)


muhammad-sayed-mahdy:
If the used `collections` dependancy is large and you don't want `wavesurfer` to be heavy, and you don't have a suggestion for another package. I have another approach in mind, keep the `handlers` as an array and when firing an event, check if the array is sorted and if it's not, sort it. (most probably it will not be sorted frequently unless someone constantly add handlers or remove them and fire the event)

thijstriemstra:
> If the used collections dependancy is large and you don't want wavesurfer to be heavy,

How many bytes are we talking about?

muhammad-sayed-mahdy:
@thijstriemstra it's almost doubled
### Before
wavesurfer.js 215.2 kB
wavesurfer.min.js 72 kB
### After
wavesurfer.js 418.8 kB
wavesurfer.min.js 140.2 kB

thijstriemstra:
ok yea that is too much for such a small feature

sundayz:
It's so big because you are using `require` instead of `import` which can do tree shaking (only import the functions that you are using)

muhammad-sayed-mahdy:
@sundayz I tried it, it's the same

thijstriemstra:
An es6 `Set` with a sort implementation sounds like most lightweight, e.g. https://stackoverflow.com/a/46267489/11355791 and  https://javascript.plainenglish.io/how-to-sort-a-set-in-javascript-es6-51b53f6ef71a

muhammad-sayed-mahdy:
@thijstriemstra This will not cause performance issues ? because it will increase the complexity of adding an event handler from O(1) (currently) to O(n log n)
Also I tried to extend the `Set` class and override the `add` method,
```js
export default class SortedSet extends Set {
    constructor(compare) {
        super();
        this.compare = compare;
    }

    add(item) {

        const arr = Array.from( this );
        arr.push( item );
        arr.sort(this.compare);
        this.clear();
        for ( let item of arr ){
            super.add(item);
        }
    }
}
```
 when I tried it, it gave me this error
image

I'm not a javascript expert tbh.

thijstriemstra:
what's the plan for this PR @muhammad-sayed-mahdy?

muhammad-sayed-mahdy:
I don't know, I just need some help, anyway the issue is in the implementation, but the idea is valid and based on it, #2307 will be solved

muhammad-sayed-mahdy:
@thijstriemstra turns out that this error was because I just forgot to run `npm install` :sweat_smile: sorry for that.
However I think this implementation will increase the complexity of adding a handler from `O(1)` (in the case of the array), to `O(n log n)` in case of sorting every time when inserting a handler.

thijstriemstra:
> However I think this implementation will increase the complexity of adding a handler from O(1) (in the case of the array), to O(n log n) in case of sorting every time when inserting a handler.

I guess we need to keep searching for a better, small size library..

entonbiba (changes requested):
I don't think we need this library to simply do the sorted set. We can rewrite it in js directly which will be very small.

src/util/observer.js
```diff
@@ -38,15 +38,15 @@ export default class Observer {
 
         let handlers = this.handlers[event];
         if (!handlers) {
-            handlers = this.handlers[event] = SortedSet([], null, (a, b) => {
+            handlers = this.handlers[event] = new SortedSet((a, b) => {
                 if (a.order && b.order && a.order != b.order) {
```
entonbiba, line 42: would it be better to do `if (a.order && b.order && a.order !== b.order)` ?

katspaugh:
I don’t think this is a good solution to the linked issue. If certain events need to be fired in a certain order, the code that triggers them should call the fire method in that order, topologically speaking.

PR #2409, opened 2021-11-21
regions plugin: disable "region-removed" event during plugin teardown

### LINKED ISSUE #2403: Destroying wavesurfer region plugin causus events to emit that appear the same as user events
I've setup a react component with wavesurfer and it works great! One thing I observed that I thought was a bit odd is if I provide functionality where a user interaction causes wavesurfer to be destroyed and recreated there are events emitted during the wavesurfer `destroy()` teardown that seem like internal events mixed with user events.

Specifically for the region plugin when [`destroy`](https://github.com/katspaugh/wavesurfer.js/blob/f04e9253feba0cbcf91e316d1a0c8818db3a363b/src/plugin/regions/index.js#L175) is called [`clear`](https://github.com/katspaugh/wavesurfer.js/blob/f04e9253feba0cbcf91e316d1a0c8818db3a363b/src/plugin/regions/index.js#L225) calls [`remove`](https://github.com/katspaugh/wavesurfer.js/blob/f04e9253feba0cbcf91e316d1a0c8818db3a363b/src/plugin/regions/region.js#L133) from `region.js` which emits `region-removed`. This is the same event that gets emitted when a user interaction causes a region to be removed, but instead it is being emitted because of an internal mechanism in the plugin teardown. It's also a public event in the [docs](https://wavesurfer-js.org/plugins/regions.html) so I think it is meant to be used for tracking user space interactions, not internal wavesurfer events if that makes sense?

This is the stack trace from devtools.

```
remove (region.js:133)
(anonymous) (index.js:220)
clear (index.js:219)
destroy (index.js:175)
destroyPlugin (wavesurfer.js:660)
(anonymous) (wavesurfer.js:674)
destroyAllPlugins (wavesurfer.js:673)
destroy (wavesurfer.js:1740)
```

I wanted to get the maintainers thoughts to see if it makes sense to change the plugin `destroy` and `clear` functions work so that `clear` doesn't emit a `region-removed` event (because it calls `remove`). Instead it would do the same operations, but emit a `regions-destroyed` or other event. I know that `clear` is removing the regions, but this feels like an internal event being mixed with events that can be tied to user interactions. If breaking this up/changing this makes sense I'd be happy to help out with that, just let me know.

### Wavesurfer.js version(s):

- 5.2.0

- OS: macOS Monterey 12.0.1
- Browser: 
  - edge 95.0.1020.44 
  - chrome 95.0.4638.69 
  - safari 15.1

Tried to setup an [example](https://codepen.io/n0mn0m/pen/XWaPorZ?editors=1010) in codepen, but running into some audio loading errors, hopefully this provides an idea, and it should be able to run locally with an audio file.

Trigger an event where wavesurfer is destroyed with the region plugin registered and observe the region plugin emit a 'region-removed' event.

n0mn0m:

Currently when wavesurfer is destroyed the `RegionPlugin` will emit a `region-removed` event the same way it would when a user removes a region. This can be confusing in applicaton code since `region-removed` can emit without users interacting with the plugin.

If anybody relied on `region-removed` being emitted during plugin destruction this would break that functionality.

### Todos/Notes:

Spent some time looking around at different ways to handle `RegionPlugin.destroy()` without emitting `region-removed` in the call chain. I came across `setDisabledEventEmissions` and that seems like a good way to suppress the event without duplicating functionality, adding conditional paths, etc.

fixes #2403

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,10 @@ wavesurfer.js changelog
   - Use of default `edgeScrollWidth` value no longer dependent on regions being created via
     plugin params (#2401)
 
+5.3.0 (21.11.2021)
+------------------
+- Regions plugin: disable `region-remove` event emission during plugin teardown (#2403)
+
 5.2.0 (16.08.2021)
 ------------------
 - Add `ignoreSilenceMode` option to ignore iOS hardware silence switch when using the
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,10 +15,7 @@ wavesurfer.js changelog
   - Allow `formatTimeCallback` from plugin params to be used (#2294)
   - Use of default `edgeScrollWidth` value no longer dependent on regions being created via
     plugin params (#2401)
-
-5.3.0 (21.11.2021)
-------------------
-- Regions plugin: disable `region-remove` event emission during plugin teardown (#2403)
+  - Disable `region-remove` event emission during plugin teardown (#2403)
 
 5.2.0 (16.08.2021)
 ------------------
--- a/src/plugin/regions/index.js
+++ b/src/plugin/regions/index.js
@@ -170,6 +170,7 @@ export default class RegionsPlugin {
     destroy() {
         this.wavesurfer.un('ready', this._onReady);
         this.wavesurfer.un('backend-created', this._onBackendCreated);
+        this.wavesurfer.setDisabledEventEmissions(['region-removed']);
         this.disableDragSelection();
         this.clear();
     }
--- a/src/plugin/regions/index.js
+++ b/src/plugin/regions/index.js
@@ -170,6 +170,10 @@ export default class RegionsPlugin {
     destroy() {
         this.wavesurfer.un('ready', this._onReady);
         this.wavesurfer.un('backend-created', this._onBackendCreated);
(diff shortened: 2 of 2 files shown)


thijstriemstra (changes requested):
thanks! Can you also add a changelog entry? And add a code comment explaining why it's disabled in destroy?

CHANGES.md
```diff
@@ -16,6 +16,10 @@ wavesurfer.js changelog
   - Use of default `edgeScrollWidth` value no longer dependent on regions being created via
     plugin params (#2401)
 
+5.3.0 (21.11.2021)
+------------------
+- Regions plugin: disable `region-remove` event emission during plugin teardown (#2403)
```
thijstriemstra, line 21: this change will be part of 6.0.0 (which is unreleased atm)

thijstriemstra (changes requested):

CHANGES.md
n0mn0m: Sorry about that. Updated. Let me know if you want me to squash these commits (you may be planning to do that, but wanted to mention it).

thijstriemstra (approved):

ApayRus:
Hi. It still happens: firing 'region-removed' on destroy wavesurfer

Issue #2381, opened 2021-10-19
How can get 'from' position when seek event fires?


Adrian-Swifter:
I am getting a position when seek event fires, and that serves as 'to', but I also need a 'from' position and that is something I didn't find looking through docs. Can you please explain how can I achieve this?

entonbiba:
@Adrian-Swifter you can use the `wavesurfer.getCurrentTime()` method that returns the current position by default you know it's the start. So next time the seek event happens inside the function save it as the current time and use that as the from.

We will add this as an enhancement to the current seek event to return the from position as well.

entonbiba:
pr created #2410

Gigamick:
I use this code which I think is what you've suggested:

```
    wavesurfer{{$p}}.on('seek', function () {
        var t = wavesurfer{{$p}}.getCurrentTime();

        $.ajax({
            type: 'POST',
            url: '/updateseek',
            data: {
                trackID: {{$item->track_id}},
                token: "{{$songbox->token}}",
                songbox: "{{$songbox->songbox}}",
                time: t
            }
        });
```

But the getCurrentTime here is the time the users seeks TO, not FROM. Am I missing something?

entonbiba:
@Adrian-Swifter use `wavesurfer.backend.getPlayedTime()`

PR #2410, opened 2021-11-22
add new seekfromto event

entonbiba:
provides the from and the to time in the new seekfromto event in a object

```js
wavesurfer.seekfromto(0.5);
wavesurfer.on("seekfromto",function(e){
  console.log(e);
  /*
    from: 0.0,
    to: 0.5
  */
});
```

#2381

### DIFF
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -927,6 +927,7 @@ export default class WaveSurfer extends util.Observer {
      * @param {number} progress Between 0 (=beginning) and 1 (=end)
      * @emits WaveSurfer#interaction
      * @emits WaveSurfer#seek
+     * @return {object} `from` being time started, `to` being time seeked for the `seek` event as {from:0, to:5}
      * @example
      * // seek to the middle of the audio
      * wavesurfer.seekTo(0.5);
@@ -947,6 +948,7 @@ export default class WaveSurfer extends util.Observer {
 
         const isWebAudioBackend = this.params.backend === 'WebAudio';
         const paused = this.backend.isPaused();
+        const currentTime = this.getCurrentTime();
 
         if (isWebAudioBackend && !paused) {
             this.backend.pause();
@@ -963,8 +965,7 @@ export default class WaveSurfer extends util.Observer {
         }
 
         this.params.scrollParent = oldScrollParent;
-        this.fireEvent('seekfromto', {from:this.getCurrentTime(), to:progress});
-        this.fireEvent('seek', progress);
+        this.fireEvent('seek', {from:currentTime, to:progress * this.getDuration()});
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/wavesurfer.js
```diff
@@ -963,6 +963,7 @@ export default class WaveSurfer extends util.Observer {
         }
 
         this.params.scrollParent = oldScrollParent;
+        this.fireEvent('seekfromto', {from:this.getCurrentTime(), to:progress});
```
katspaugh, line 966: As an alternative, `{ from, to }` could be a second parameter to the existing `seek` event. WDYT?
Also, if `from` is a time in seconds, `to` should probably be as well a timestamp, not a float.
Please don't forget to document it.

thijstriemstra: > a second parameter to the existing seek event.

Agreed.

entonbiba: yea I was thinking of adding it to the existing one but older projects using the event might break

thijstriemstra: we're planning on releasing 6.0 where it's ok to break some backwardcompat imo. as long as it's documented in the upgrade document..

katspaugh: The chance of something breaking because of an extra argument at the end seems slim tbh.

thijstriemstra (changes requested):
see inline comments

src/wavesurfer.js
entonbiba: yes I'll return both as the actual timestamp

src/wavesurfer.js
```diff
@@ -963,7 +964,7 @@ export default class WaveSurfer extends util.Observer {
         }
 
         this.params.scrollParent = oldScrollParent;
-        this.fireEvent('seek', progress);
+        this.fireEvent('seek', {from:currentTime, to:progress * this.getDuration()});
```
msuhov, line 967: This creates 2 breaking changes..
First is the object instead of number - can we pass `from` as second argument?
Second - fraction is transformed to seconds. Shouldn't we stick to fraction between 0 and 1 as unified format for events?
On other hand, if we want to use seconds in event, we should accept seconds in `seekTo` method as well
```
seekTo(amount: number, type?: 'seconds' | 'fraction'): void;
```

katspaugh: Exactly, `progress` should remain the first argument. `from` and `to` should come next. They don't even have to be an object, could be just positional arguments.
@entonbiba also be wary that `progress * this.getDuration()` may yield many digits after the comma. You might want to round it.

thijstriemstra:
Marking this is a draft PR. Please submit for review once it's ready @entonbiba

thijstriemstra:
> Marking this is a draft PR. Please submit for review once it's ready

Sadly no feedback, closing.

Issue #1248, opened 2017-11-06
Fixed frequency range spectrogram plugin 


pcprince:
I'm attempting to plot wav files which can possibly have a sample rate of 250kHz in order to pick up bats or as low as 8kHz. I would assume the plugin would take the sample rate into account and draw the spectrogram from 0kHz to half the sample rate, but that's not the case. The spectrogram plugin seems to only show 0kHz to 24kHz, which isn't useful when producing plots requiring a different range of frequencies.

I can't find any documentation on changing the ranges so was wondering if this is simply not an included feature or am I missing something?

pcprince:
Could it have something to do with the fact that, no matter the sample rate of the file being loaded, `wavesurfer.backend.ac.sampleRate` is always 48000?

entonbiba:
@pcprince may you post some test code?

pcprince:
This is using WaveSurfer 2.0.0-beta01:
```
var WaveSurfer = require('wavesurfer.js');
var SpectrogramPlugin = require('wavesurfer.js/dist/plugin/wavesurfer.spectrogram.min.js');

var spectrogram = SpectrogramPlugin.create({
    container: "#wave-spectrogram",
    labels: true
});

var wavesurfer = WaveSurfer.create({
    container: '#waveform',
    waveColor: '#FF0000',
    progressColor: '#46B54D',
    hideScrollbar: true,
    plugins: [
        spectrogram
    ]
});

wavesurfer.load('bat.wav');
console.log(wavesurfer.backend.ac.sampleRate);
```
The bat wav file is a recording at 250kHz. The console outputs 48000 and only shows the spectrogram up to 24kHz. If it actually was a 48kHz file, stopping at 24kHz would make sense. It seems to use that range of 0-24kHz for every file and there doesn't seem to be any way to either change the spectrogram sample rate or make WaveSurfer treat files as having any other sample rates.

thijstriemstra:
Closing question, no feedback.

PR #2424, opened 2022-01-06
spectrogram: Add multi-channel, Remove inaccurate doubling

### LINKED ISSUE #2232: Spectrogram for mono files doubles the frequency
### Wavesurfer.js version(s):
4.6

Windows 10; Chrome 89, Firefox 86

The drawSpectrogram function inside the spectrogram plugin has a heightFactor:

const heightFactor = my.buffer ? 2 / my.buffer.numberOfChannels : 1;

For mono files the heightfactor is 2 what is wrong. This should always be 1 because the frequency doesn't change just because it's a mono file.

(Please reduce your code as much as possible and only post the minimum code needed to reproduce the issue. [A Code pen](http://codepen.io/) is an excellent way to share such code)

Add the spectrogram plugin with labels to any wavesurfer session and load a mono file. The frequencies will be wrong.

dofuuz:
1. Remove inaccurate doubling of spectrogram (#2232)
2. Setting `splitChannels: true` going to draw spectrograms for each channels.

3ch spectrogram example
3ch spectrogram example

None.

None.

### Todos/Notes:
**This PR will change output of spectrogram.**

It seems like web browser resamples audio to fixed samplerate(in my case, 48kHz). #1248
So top of spectrogram frequency will be fixed regardless of original samplerate.

We can scale spectrogram, if there are nice way to get original samplerate(I couldn't find it).

#1248

fixes #2232

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -16,6 +16,9 @@ wavesurfer.js changelog
   - Use of default `edgeScrollWidth` value no longer dependent on regions being created via
     plugin params (#2401)
   - Disable `region-remove` event emission during plugin teardown (#2403)
+- Spectrogram plugin:
+  - Remove inaccurate frequency doubling of spectrogram (#2232)
+  - Add `splitChannels` option to draw spectrograms for each channels (#2424)
 
 5.2.0 (16.08.2021)
 ------------------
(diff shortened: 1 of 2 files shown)


thijstriemstra (changes requested):
thanks! can you also add a changelog entry @dofuuz?

thijstriemstra (approved):
awesome, thanks.

PR #2428, opened 2022-01-09
fixed bars sizing using max peak value in range

lorenzos:
This changes the way drawer renders bars. Prior to this, the height of each bar was [calculated only on the single peak at the start of bar, in `drawer.multicanvas.js:311`](https://github.com/katspaugh/wavesurfer.js/blob/master/src/drawer.multicanvas.js#L311). This made the bars render almost randomly, especially at smaller widths, since highest peaks were very easily ignored, if not falling exactly where a bar started.

Here's an animated example of **before**, as you can see the original waveform is translated poorly, and then when resizing, bars jump around for no good reason:

Bars size before

Now instead, the height is [calculated on the whole range of peaks that each bar falls into](https://github.com/lorenzos/wavesurfer.js/commit/243a595a5c5684c44b06619d1aed4fe38fa93429#diff-9e1d85325663a8f99b981eac351eef64b1ed40011d998b4d502b6b6ac4115609R311). I used the [maximum peak value](https://github.com/lorenzos/wavesurfer.js/commit/243a595a5c5684c44b06619d1aed4fe38fa93429#diff-9e1d85325663a8f99b981eac351eef64b1ed40011d998b4d502b6b6ac4115609R315); average works too, but it would require normalization to make sure the taller bar is at maximum height.

Here's the same animated example **after**, the translation is better and bars re-sampling is definitely more consistent:

Bars size after

---

### Addendum

While fixing this, I noticed another issue when choosing the peak index to use for heights calculation. There is a [`peakIndexScale` to "skip every other value if there are negatives"](https://github.com/katspaugh/wavesurfer.js/blob/master/src/drawer.multicanvas.js#L292) in the peaks array. It made sense, but was used incorrectly [in the actual formula](https://github.com/katspaugh/wavesurfer.js/blob/master/src/drawer.multicanvas.js#L312), before the rounding (floor) instead of [after](https://github.com/lorenzos/wavesurfer.js/commit/243a595a5c5684c44b06619d1aed4fe38fa93429#diff-9e1d85325663a8f99b981eac351eef64b1ed40011d998b4d502b6b6ac4115609R312).

This made it possible to actually choose odd indexes, and therefore negative peak values. I guess this went unnoticed since the bar is mirrored anyway. Here's a comparison of chosen peaks before and after, note how the fix "snaps" to even indexes only:

Peak indexes before Peak indexes after

### DIFF
--- a/src/drawer.multicanvas.js
+++ b/src/drawer.multicanvas.js
@@ -305,20 +305,22 @@ export default class MultiCanvas extends Drawer {
                 const scale = length / this.width;
                 const first = start;
                 const last = end;
-                let i = first;
-
-                for (i; i < last; i += step) {
+                let peakIndex = first;
+                for (peakIndex; peakIndex < last; peakIndex += step) {
 
                     // search for the highest peak in the range this bar falls into
                     let peak = 0;
-                    let peakIndex = Math.floor(i * scale) * peakIndexScale; // start index
-                    const peakIndexEnd = Math.floor((i + step) * scale) * peakIndexScale;
+                    let peakIndexRange = Math.floor(peakIndex * scale) * peakIndexScale; // start index
+                    const peakIndexEnd = Math.floor((peakIndex + step) * scale) * peakIndexScale;
                     do { // do..while makes sure at least one peak is always evaluated
-                        peak = Math.max(peak, peaks[peakIndex] || 0); // highest
-                        peakIndex += peakIndexScale; // skip every other value for negatives
-                    } while (peakIndex < peakIndexEnd);
+                        const newPeak = peaks[peakIndexRange];
+                        if (newPeak > peak) {
+                            peak = newPeak; // higher
+                        }
+                        peakIndexRange += peakIndexScale; // skip every other value for negatives
+                    } while (peakIndexRange < peakIndexEnd);
 
-                    // calculate the height this bar according to highest peak found
+                    // calculate the height of this bar according to the highest peak found
                     let h = Math.round((peak / absmax) * halfH);
 
                     // in case of silences, allow the user to specify that we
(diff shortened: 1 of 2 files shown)


thijstriemstra (changes requested):
thanks @lorenzos 

could you add as much comments in that piece of code as possible? It's hard to understand without any.

Also, please add a changelog entry.

lorenzos:
Done!

src/drawer.multicanvas.js
```diff
@@ -308,12 +308,21 @@ export default class MultiCanvas extends Drawer {
                 let i = first;
 
                 for (i; i < last; i += step) {
-                    const peak =
-                        peaks[Math.floor(i * scale * peakIndexScale)] || 0;
+
```
thijstriemstra, line 311: can you remove this empty line, and rename `let i` to something more meaningful?

src/drawer.multicanvas.js
```diff
@@ -308,12 +308,21 @@ export default class MultiCanvas extends Drawer {
...
+                    // search for the highest peak in the range this bar falls into
+                    let peak = 0;
+                    let peakIndex = Math.floor(i * scale) * peakIndexScale; // start index
+                    const peakIndexEnd = Math.floor((i + step) * scale) * peakIndexScale;
+                    do { // do..while makes sure at least one peak is always evaluated
+                        peak = Math.max(peak, peaks[peakIndex] || 0); // highest
+                        peakIndex += peakIndexScale; // skip every other value for negatives
+                    } while (peakIndex < peakIndexEnd);
+
+                    // calculate the height this bar according to highest peak found
```
thijstriemstra, line 321: `of` and `the` are missing in this comment

thijstriemstra (approved):

PR #2432, opened 2022-01-20
Readme: add info about a 10-year NFT

katspaugh:
I've created an NFT on Polygon that represents the initial commit to this repo.
I think it would be a nice way to support the project and celebrate the 10 years.
WDYT, @thijstriemstra?

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,16 @@ Interactive navigable audio visualization using Web Audio and Canvas.
 
 See a [tutorial](https://wavesurfer-js.org/docs) and [examples](https://wavesurfer-js.org/examples) on [wavesurfer-js.org](https://wavesurfer-js.org).
 
+## 10 years of wavesufer.js 🎉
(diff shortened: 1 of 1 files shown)


thijstriemstra:
please explain how this would work. ps i think nft's is one of the dumbest and most disingenuous technological concepts ever created.

katspaugh:
It will work like a regular donation, just with a symbolic token granted to the donator.

If it gets sold I'd be happy to send you 50% (0.5 ETH). Just tell me your wallet address. If you don't have one yet, [Metamask](https://metamask.io) is a good option.

thijstriemstra:
because im against nft's shouldn't stop the project from participating in this area so give it a try; if it doesn't get picked up within a year, let's remove it again from the readme nxt year. 1 eth seems cheap though ;)

thijstriemstra (approved):

thijstriemstra:
and i can give you a wallet address if you email me once it's sold. email can be found in git log.

katspaugh:
Perfect, thank you! 😄

PR #2446, opened 2022-02-10
fixed bars rendering for peaks array starting with a negative value

### LINKED ISSUE #2439: Pre-rendered Peaks Don't Work for Bars Rendering on 6.0.0

The following below **works** on Wavesurfer.js version 5.2.0 but *not* on 6.0.0,

```javascript
let wavesurfer = WaveSurfer.create({
  container: '#wavesurfer,
  backend: 'MediaElement',
  barWidth: 5 // Works when this is omitted, so applies only to bar rendering
})

wavesurfer.load('https://example.com/audio.mp3', arrayOfPeaks)
```

Observe peaks aren't pre-rendering, and bar rendering is empty/flat/silent. If i omit the `barWidth: 5` option, it works as usual with normal a waveform rendering.

Thanks,

David

lorenzos:
Fixes #2439, an "unmasked" regression introduced with #2428. That regression broke bars rendering in the case a given pre-computed array of peaks alternates negative/positive values, instead of alternating positive/negative values (as wavesurfer itself generates).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 wavesurfer.js changelog
 =======================
 
+6.0.1 (unreleased)
+------------------
+- Fixed a regression that broke bars rendering when using a certain format for the peaks array (#2439, #2446)
+
 6.0.0 (07.02.2022)
 ------------------
 - Add additional type to `waveColor` and `progressColor` parameters to support linear gradients (#2345)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,7 @@ wavesurfer.js changelog
 
 6.0.1 (unreleased)
 ------------------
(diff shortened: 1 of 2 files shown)


dtcooper (changes requested):
👏👍🖖

thijstriemstra (commented):
thanks @lorenzos can you also add a comment for this line, explaining why abs is needed?

CHANGES.md
```diff
@@ -1,6 +1,10 @@
 wavesurfer.js changelog
 =======================
 
+6.0.1 (unreleased)
+------------------
+- Fixed a regression that broke bars rendering when using a certain format for the peaks array (#2439, #2446)
+
```
thijstriemstra, line 7: the issue nr is enough, the pr nr is not needed here

thijstriemstra (approved):

PR #2454, opened 2022-02-20
Correct type of maxRegions property of RegionsPlugin

JanMisker:
Changed the type of `maxRegions` from `number[]` to `number`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 wavesurfer.js changelog
 =======================
 
+6.1.0
+------------------
+- Fixed the type annotation of `maxRegions` in the regions plugin.
+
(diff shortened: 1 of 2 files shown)


caugner (approved):
LGTM, see usage here: https://github.com/katspaugh/wavesurfer.js/blob/b740bb15c2985a223d2c0349200a5359ca06c25d/src/plugin/regions/index.js#L189

thijstriemstra:
thx, can you also add a changelog entry? lets use 6.1.0 as new version in changelog

JanMisker:
@thijstriemstra added a changlog entry. 

En bedankt ;)

thijstriemstra (approved):

PR #2470, opened 2022-03-04
timeline plugin: fix negative offset rendering

### LINKED ISSUE #2463: timeline plugin: negative offset rendering issue
First of all: thank you for this fantastic library  
So cool that nearly all of my special needs can be solved by configuration - i never would have expected such a great flexibility!!!  

Unfortunately i found a side effect that may be a kind of bug caused by "over-configuration" :D  
Currently i am doing some prototyping for a DJ-like web application for playing acapellas for synth jam sessions with precise control for pitching & nudging.  

Within this app i need a beat-grid indicating **quarter notes** and **8 bars** as the "longest" unit. It was dead easy to achieve this with the timeline plugin. Thanks to the `options.offset` i can exactly place the **8 bars** to the downbeat.  

## Approach 1: `timeline.options.offset = positive float`
This has the side effect that there is **no rendering** of the beat-grid between 0 and this configured second  
image  

## Approach 2: `timeline.options.offset = negative float`
My idea was to subtract the 8 bars duration from the downbeat to get rid of the missing rendering at the begin of the waveform.  
This basically works but now i have no rendering of the beat-grid **at the end** of the waveform.  

image  
Then i found `timeline.options.duration` which overrides the duration. but adding this negative offest to the duration messes up the resolution of the beat-grid

**So here is my question:**  
Can you confirm that this very special need/usecase can't be solved by configuration?  
Thank you

othmar52:
when using negative values as `timeline.options.offset` the timeline is now rendered from start to the very end.  
before that change the timeline for those seconds of negative offset had not been rendered at the end of the waveform

fixes #2463

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 wavesurfer.js changelog
 =======================
 
+6.0.4 (04.03.2022)
+------------------
+- Timline plugin:
+  - Fix rendering issue on negative offset values (#2463)
+
 6.0.3 (01.03.2022)
 ------------------
 - Cursor plugin:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,10 +1,10 @@
 wavesurfer.js changelog
 =======================
 
-6.0.4 (04.03.2022)
+6.0.4 (unreleased)
 ------------------
-- Timline plugin:
-  - Fix rendering issue on negative offset values (#2463)
+- Timeline plugin:
+  - Fix rendering issue for negative `offset` values (#2463)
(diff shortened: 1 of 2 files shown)


thijstriemstra (changes requested):
thanks! Can you also add a changelog entry (6.0.4)?

thijstriemstra (approved):
thanks.

othmar52:
youre welcome!  
sorry for the typo - i will pay more attention next time i contribute to wavesurfer

PR #2477, opened 2022-03-09
[DOCUMENTATION] change description of audioprocess event

### LINKED ISSUE #2475: documentation of `audioprocess` event needs improvement
related to [issue #618](https://github.com/katspaugh/wavesurfer.js/issues/618)  

The [documentation](https://wavesurfer-js.org/docs/events.html) reads:  
 - `audioprocess` – Fires continuously as the audio plays. **Also fires on seeking.**  

`Also fires on seeking` is not true when audio is paused  
 
### Wavesurfer.js version:
wavesurfer.js 6.0.2

open https://wavesurfer-js.org/example/minimap/index.html  
open js console and add those 2 lines  
```js
window.wavesurfer.on('audioprocess', (sec) => { console.log('audioprocess', sec); });
window.wavesurfer.on('seek', (value) => { console.log('seek', value); });
```
check console output when seeking during play  
check console output when seeking during pause  `<-- audioprocess event is not fired`  

@thijstriemstra do you think its better to add a note to the documentation like   
`...Also fires on seeking as long as the audio is not paused`  

or does it make sense to fire a single `audioprocess` event after seeking?  
or should the hint `Also fires on seeking` be removed?  
or do you think documentation of the `audioprocess` event is perfect and there is no need to change anything at all?

othmar52:
fixes #2475 

remove hint `Also fires on seeking.` of description of `audioprocess` event as it's not true during paused player

### DIFF
--- a/docs/events.html
+++ b/docs/events.html
@@ -13,7 +13,7 @@ layout: default
   <p>Here's the list of available events:</p>
 
   <ul>
-    <li><code>audioprocess</code> – Fires continuously as the audio plays. Also fires on seeking.</li>
+    <li><code>audioprocess</code> – Fires continuously as the audio plays.</li>
     <li><code>dblclick</code> – When instance is double-clicked.</li>
     <li><code>destroy</code> – When instance is destroyed.</li>
     <li><code>error</code> – Occurs on error.  Callback will receive (string) error message.</li>


thijstriemstra (approved):
thanks!

PR #2480, opened 2022-03-10
minimap plugin: clear waveform when load new audio

### LINKED ISSUE #2479: minimap plugin: clear minimap when loading a new track
maybe related to [issue #617](https://github.com/katspaugh/wavesurfer.js/issues/617)  
### Wavesurfer.js version(s):
wavesurfer.js 6.0.2

open https://wavesurfer-js.org/example/minimap/index.html
open js console and add this line
```js
window.wavesurfer.load('https://ia804502.us.archive.org/13/items/DontWatchTheDJ128BPMDry/Don%27t%20Watch%20The%20DJ%20128BPM%20Dry.mp3');
```
check minimap during load `<-- minimap of previous track still visible until load+analyzing is finished`  

Shouldn't the minimap be cleared automatically when loading a new track?  
or is it necessary to `destroy()` and `initPlugin('minimap')` manually?

othmar52:
As soon as a new audio file starts to load the current minimap is cleared

fixes #2479

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 wavesurfer.js changelog
 =======================
 
+6.0.5 (unreleased)
+------------------
+- Minimap plugin:
+  - Remove waveform of previous audio when starting to load new audio (#2479)
+
 6.0.4 (09.03.2022)
 ------------------
 - Spectrogram plugin:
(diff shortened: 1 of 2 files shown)


thijstriemstra (commented):
what is the behavior of the various other plugins when you load new audio?

othmar52:
> what is the behavior of the various other plugins when you load new audio?

there are no changes that affect any behaviour of events. no events are added removed or modified.    
so i assume all other plugin's behaviour is unchanged as well.

but i have to admit that i don't even know all the wavesurver plugins that exists.  
do you think it's necessary to test all permutations of plugin combinations that includes the timeline plugin?  
or can the testing be limited to a few scenarios?

PR #2485, opened 2022-03-20
responsive debounce optimizing: don't fire redraw if waveform is invisible

marcelblum:
Avoid unnecessary calculations when performing the responsive debounced resize check.

Background: In my app I use many simultaneous Wavesurfer instances and toggle showing/hiding different waveforms at different times. I noticed that on window resize each instance was firing `redraw` from the debounce function even if the waveform container was unpainted/had 0 width, leading to unneeded calculations. So I propose adding a check for a nonzero width value before incurring the expense of calling `redraw`.

This is a bit of a microoptimization that won't noticeably affect users with only a single Wavesurfer instance but for those using dozens or hundreds of instances the benefit compounds and this will reduce cpu use during resizes.

nothing

nothing

### DIFF
--- a/src/wavesurfer.js
+++ b/src/wavesurfer.js
@@ -505,7 +505,9 @@ export default class WaveSurfer extends util.Observer {
                     !this.params.scrollParent
                 ) {
                     prevWidth = this.drawer.wrapper.clientWidth;
-                    this.drawer.fireEvent('redraw');
+                    if (prevWidth) {
+                        this.drawer.fireEvent('redraw');
+                    }
                 }
             },
             typeof this.params.responsive === 'number'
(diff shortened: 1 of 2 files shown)


src/wavesurfer.js
```diff
@@ -505,7 +505,9 @@ export default class WaveSurfer extends util.Observer {
                     !this.params.scrollParent
                 ) {
                     prevWidth = this.drawer.wrapper.clientWidth;
-                    this.drawer.fireEvent('redraw');
```
thijstriemstra, line 508: can you add a comment why you add this condition?

thijstriemstra (changes requested):

src/wavesurfer.js
marcelblum: I've added a comment

thijstriemstra (approved):

Issue #2481, opened 2022-03-15
Many calls to setSinkId results in no sound


kross-italk:
- [X] I have searched for [already open issues](https://github.com/katspaugh/wavesurfer.js/issues) which desribe my problem.
https://github.com/katspaugh/wavesurfer.js/issues/2060
- [X] The issue I'm having is related to and caused by wavesurfer.js, not by other software (which maybe packages and uses wavesurfer incorrectly) – In that case you should open the issue on the respective project pages.

### Wavesurfer.js version(s):
Version: 5.2.0

### Browser and operating system version(s):
Browser: Chrome/latest Version 99.0.4844.51 (Official Build) (64-bit)

### Code needed to reproduce the issue:
https://jsfiddle.net/m57h3qcg/4/
1. Run wavesurfer.setSinkId lots of times in any tab
2. After 50+ attempts, new Audio within that tab will stop being heard
3. After 500+ attempts, Browser is now exhausted of audio nodes 
4. Go to YouTube and try to play any video. 
5. You will see Loading, then a play button. Click play and you will see "Audio rendering error"

### Use behaviour needed to reproduce the issue:
We call setSinkId inside a React componentDidMount. If a user saves audio settings and navigates between different players they will eventually run out of audio nodes. 

Here is the code: https://github.com/katspaugh/wavesurfer.js/blob/19afccbe8cde790986989105d99f5935c8088f03/src/webaudio.js#L301

You can see that invoking `setSinkId()` will always create a new Audio node. I suggest moving this audio node to be a member of the enclosing class. Happy to submit a PR.

thijstriemstra:
> Happy to submit a PR.

Sounds good!

PR #2492, opened 2022-03-25
track and disconnect audio resources used by setSinkId

kross-italk:
Short description of changes:

Invoking setSinkId will create an Audio and MediaStream.
Create them once per wavesurfer and then disconnect/NULL them explicitly
on destroy.

After this change observe chrome://media-internals and create/setSinkId/destroy
many wavesurfers in a loop

Related Issues and other PRs:
fixes https://github.com/katspaugh/wavesurfer.js/issues/2481

Demo of media-internals after this change:
kross-2481-media-internals


src/webaudio.js
```diff
@@ -298,17 +302,22 @@ export default class WebAudio extends util.Observer {
              * output. Here we create an HTMLAudioElement, connect the
              * webaudio stream to that element and setSinkId there.
              */
-            let audio = new window.Audio();
+            if (!this.audio) {
+                this.audio = new window.Audio();
+                this.audio.autoplay = true;
+            }
+            let audio = this.audio;
```
msuhov, line 309: Hi @kross-italk! Thanks for your commit

Quick suggestion: since we already have `this.audio` variable, I don't think we need to reassign it to another var, just use it directly in this method

CHANGES.md
```diff
@@ -1,6 +1,11 @@
 wavesurfer.js changelog
 =======================
 
+6.0.5 (25.03.2022)
+------------------
+- setSinkId:
```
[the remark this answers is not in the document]
thijstriemstra: and no need to add a release date (use `unreleased` instead)

CHANGES.md
```diff
@@ -1,6 +1,11 @@
...
+  - Fix #2481 Many calls to setSinkId results in no sound
```
thijstriemstra, line 7: ticket nr should go at end of sentence (see other changelog entries)

src/webaudio.js
```diff
@@ -145,6 +145,10 @@ export default class WebAudio extends util.Observer {
         this.state = null;
         /** @private */
         this.explicitDuration = params.duration;
+        /** @private */
+        this.dest = null;
```
thijstriemstra, line 149: i don't like `dest` as a var name on a class level. Can you call it something like `sinkDestination` or `sinkStreamDestination`? Same goes for `this.audio`, something like `sinkAudioElement`?

src/webaudio.js
```diff
@@ -541,6 +549,13 @@ export default class WebAudio extends util.Observer {
             // clear the offlineAudioContext
             window.WaveSurferOfflineAudioContext = null;
         }
+
```
thijstriemstra, line 552: add a comment making it clear this is related to `setSinkId`?

src/webaudio.js
```diff
@@ -298,19 +302,23 @@ export default class WebAudio extends util.Observer {
              * output. Here we create an HTMLAudioElement, connect the
              * webaudio stream to that element and setSinkId there.
              */
-            let audio = new window.Audio();
-            if (!audio.setSinkId) {
+            if (!this.sinkAudioElement) {
+                this.sinkAudioElement = new window.Audio();
+                this.sinkAudioElement.autoplay = true;
```
thijstriemstra, line 307: is this autoplay option necessary?

src/webaudio.js
```diff
@@ -298,19 +302,23 @@ export default class WebAudio extends util.Observer {
...
+            }
+            if (!this.sinkAudioElement.setSinkId) {
                 return Promise.reject(
                     new Error('setSinkId is not supported in your browser')
                 );
             }
-            audio.autoplay = true;
```
kross-italk, line 307: > is this autoplay option necessary? if so, add a comment why
It was already here, I do not think it's necessary since we're not invoking playback via `.play()`.

thijstriemstra: ah yea missed that. I guess it doesn't hurt to leave it in for now.

thijstriemstra (approved):

src/webaudio.js
kross-italk: @thijstriemstra Interesting. I tested with and without the autoplay. It no longer allocates the items in media-internals but doesn't playback any audio to the chosen sink. 
So now, I think a comment about this being load bearing is important.

thijstriemstra: if you want to add another comment, sure, please do

thijstriemstra: also, please make sure to test with firefox as well.

thijstriemstra (approved):

src/webaudio.js
kross-italk: > also, please make sure to test with firefox as well.

I'm having trouble getting setSinkId/enumerateDevices to work in Firefox. and a harder time following the tracker. 
https://bugzilla.mozilla.org/show_bug.cgi?id=1610319
https://bugzilla.mozilla.org/show_bug.cgi?id=1493982

thijstriemstra: you essentially didn't change the implementation, only the cleanup, and I tested it in Firefox (first) before so I can merge this and if you see any issues with Firefox, please open a ticket or new PR?

PR #2499, opened 2022-04-12
Fixes potential error with clientWidth on resizing

### LINKED ISSUE #2498: TypeError: Cannot read properties of null (reading 'clientWidth')
### Wavesurfer.js version(s):
6.1

All

Hi! Thanks for this great library. One comment: 
When, using react, I set the "responsive" param to true (or any number, **including ZERO**) if I have a hook that gets the screen resolution and I do this:

`	
        if (!isLgAndAbove) {
		return (<WaveSurfer />)
        } else {
       return (<OtherStuff />) 
}
 `

I get an error, because the function doesn't check if the element exists. This should be changed on line 4149 of the /dist version of wavesurfer.js
Instead of 
`      if (prevWidth != _this.drawer.wrapper.clientWidth && !_this.params.scrollParent) {
`

it should be something like 
    `  if (_this.drawer.wrapper && prevWidth != _this.drawer.wrapper.clientWidth && !_this.params.scrollParent) {`

miltonlaufer:
closes #2498

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 wavesurfer.js changelog
 =======================
 
+6.2.0 (unreleased)
+------------------
+- Fix `clientWidth` error in responsive mode (#2498)
+
 6.1.0 (31.03.2022)
 ------------------
 - Fix many calls to `setSinkId` resulting in no sound (#2481)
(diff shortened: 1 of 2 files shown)


thijstriemstra (changes requested):
thanks @miltonlaufer, can you also add a changelog entry (use `6.2.0 (unreleased)`)?

Issue #2502, opened 2022-04-18
Mute state is not respected after changing volume


othmar52:
changing volume during mute unmutes the audio  

### Wavesurfer.js version(s):
6.1.0

### Code needed to reproduce the issue:
  - open https://wavesurfer-js.org/example/mute/index.html  
  - mute the audio with the mute button  
  - ensure you can't hear any sound when pressing the play button
  - open JS console and type `window.wavesurfer.setVolume(0.5);`
  - now you can hear the audio but wavesurver claims to be still muted  
   
![image](https://user-images.githubusercontent.com/7056051/163789874-b36c3d5c-4466-4031-ba2f-36193badb7cb.png)

PR #2503, opened 2022-04-18
[FIX] respect mute state when changing volume during mute

othmar52:
fixes  https://github.com/katspaugh/wavesurfer.js/issues/2502


thijstriemstra (changes requested):
thanks @othmar52, can you also add a unit test for this? And use 6.2.0 in changelog instead.

othmar52:
> thanks @othmar52, can you also add a unit test for this?  

Unfortunately i don't know how to do this.  
Do you have any commit-hash that adds a unit test for a bugfix in the wavesurfer repo? Case yes i can take a look and give it a try...

othmar52:
@thijstriemstra  Thank you for the link!  
I will take a look and try to add a unit test.  But i am unsure which behaviour should be forced.  
In my opinion the current behaviour (without my changes) is **not correct**:  
image  
i think `getVolume()` should always return the actual or saved volume no matter if the audio is muted or not.  
Do you agree or do you think `getVolume()` should always return `0` when muted?

katspaugh (approved):
I suggest we merge this w/o tests for now since the code is pretty trivial.

PR #2512, opened 2022-05-11
cursor plugin: fix typeError if displayTime: null

### LINKED ISSUE #2501: [CURSOR PLUGIN]: TypeError: Cannot read properties of null (reading 'getBoundingClientRect') when showTime: false
Following line is throwing error, because if `showTime: false` then the `displayTime` is null:
https://github.com/katspaugh/wavesurfer.js/blob/master/src/plugin/cursor/index.js#L93

### Wavesurfer.js version(s):
`v6.1.0`

`Google Chrome Version 100.0.4896.75 (Official Build) (x86_64)`, system MacOs

```js
import WaveSurfer from "wavesurfer.js";
import CursorPlugin from "wavesurfer.js/dist/plugin/wavesurfer.cursor.min.js";
//...
const wave = WaveSurfer.create({
  container: "#waveform",
  cursorWidth: 0,
  height: 36,
  normalize: true,
        plugins: [
          CursorPlugin.create({ }),
        ],
});
```

Error in the browser: 

```
wavesurfer.cursor.min.js?2e66:6 Uncaught TypeError: Cannot read properties of null (reading 'getBoundingClientRect')
    at HTMLUnknownElement.eval (wavesurfer.cursor.min.js?2e66:6:1043)
```

Create a Cursor plugin without passing `showTime: true` and try to hover the waveform.

MikeyZat:
**Title:** Fix typeError if `displayTime: null` in Cursor plugin

Adding a simple check for `this.displayTime` before getting its `getBoundingClientRect()`. Previously it was giving an error:
```
Cannot read properties of null (reading 'getBoundingClientRect')
```

None

None

### Todos/Notes:
None

fixes #2501

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@ wavesurfer.js changelog
 6.2.0 (unreleased)
 ------------------
 - Fix `clientWidth` error in responsive mode (#2498)
+- Fix `TypeError` when `showTime: undefined` in cursor plugin (#2501)
 
 6.1.0 (31.03.2022)
 ------------------
--- a/src/plugin/cursor/index.js
+++ b/src/plugin/cursor/index.js
@@ -90,7 +90,8 @@ export default class CursorPlugin {
         const bbox = this.wrapper.getBoundingClientRect();
         let y = 0;
         let x = this.wrapper.scrollLeft + event.clientX - bbox.left;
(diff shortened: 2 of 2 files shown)


thijstriemstra (changes requested):
thanks @MikeyZat can you also add a changelog entry?

MikeyZat:
@thijstriemstra sure, it's done

PR #2523, opened 2022-05-29
Draw all bars shorter than barMinHeight with the specified minimum height

### LINKED ISSUE #2522: barMinHeight does not behave as expected
As far as I understand `barMinHeight` should set the absolute minimum bar height, i.e. no bar should have a height that is smaller than specified in the `barMinHeight` option. However, currently only absolute silence is getting raised to the `barMinHeight` value. That means if for example `barMinHeight` is 4 and the peak for a certain bar is 2 this bar is rendered with a height of 2px, but should be rendered with a height of 4px.

That means a waveform with near silence (for the config see below) is expected to look like this:
Bildschirmfoto 2022-05-29 um 16 29 19

But with the current logic looks like this:
Bildschirmfoto 2022-05-29 um 16 29 35

This is due to these lines in the MultiCanvas renderer. https://github.com/katspaugh/wavesurfer.js/blob/0edd6724b89165ea73c1d589d8f635d874a5ee9c/src/drawer.multicanvas.js#L328-L330
It checks if the calculated height is 0 and raises the height to `barMinHeight` if so, but leaves values between 0 and `barMinHeight` untouched.
I suggest using this instead:
```javascript
if (this.params.barMinHeight) { 
    h = Math.max(h, this.params.barMinHeight);
}
```
Which would clip the bar height to `barMinHeight` like you see in the first image.

**Wavesurfer.js version(s)**: 6.2.0 (and below)
**Browser and operating system version(s)**: Tested on Safari 15.5 on macOS 12.4, but applicable to all browsers

Any wavesurfer setup where there is silence in the loaded audio and `barMinHeight` is greater than 1.
The configuration for the examples:
```javascript
WaveSurfer.create({
    waveColor: '#B1B1B1',
    progressColor: '#B1B1B1',

    cursorColor: '#E50000',
    cursorWidth: 2,

    barWidth: 4,
    barGap: 4,
    barRadius: 5,
    barMinHeight: 4,

    height: 250,
    barHeight: 200 / 250,
    minPxPerSec: 71,
    fillParent: false,
    scrollParent: true,
    hideScrollbar: true,
})
```

drinking-code:
As described in the issue: previously, bar heights greater than 0, that are also smaller than `barMinHeight` were not changed. This for example resulted in a bar with a height of 1 to be smaller than `barMinHeight` if the option was set to anything higher than 1. Now, all bars are set to be at least `barMinHeight`.

### Breaking in the API:
No changes to any APIs.

fixes #2522


src/drawer.multicanvas.js
```diff
@@ -325,8 +325,8 @@ export default class MultiCanvas extends Drawer {
 
                     // in case of silences, allow the user to specify that we
                     // always draw *something* (normally a 1px high bar)
-                    if (h == 0 && this.params.barMinHeight) {
-                        h = this.params.barMinHeight;
+                    if (this.params.barMinHeight) {
+                        h = Math.max(h, this.params.barMinHeight);
```
thijstriemstra, line 329: can you add a comment describing why math.max needs to be used etc. The more details, the better.

Also, a changelog entry is needed.

thijstriemstra (changes requested):

thijstriemstra:
ping @drinking-code, see inline comment

thijstriemstra (approved):

PR #2546, opened 2022-07-05
markers plugin: add support for a context menu event on a marker

kpanditCtr:
This PR adds support for a context menu event on a marker in the markers plugin. It contains an option to disable the native context menu from being opened (`preventContextMenu`, defaults to false). This is almost identical to #1844 which implements the same functionality for the regions plugin.

I've also added an example of the event to example/markers

None

None

### Todos/Notes:
None

#1844


kpanditCtr:
@thijstriemstra Would appreciate your feedback!

thijstriemstra:
> Would appreciate your feedback!

Was on a holiday, sorry for the delay. Would be great if others could help me review PRs every now and then..

src/plugin/markers/index.js
```diff
@@ -8,6 +8,7 @@
  * @property {?position} string "top" or "bottom", defaults to "bottom"
  * @property {?markerElement} element An HTML element to display instead of the default marker image
  * @property {?draggable} boolean Set marker as draggable, defaults to false
+ * @property {?boolean} preventContextMenu Determines whether the context menu is prevented from being opened, defaults to false
```
thijstriemstra, line 11: can you keep this sentence on multiple lines (width of <= 80 chars)

src/plugin/markers/index.js
```diff
@@ -277,6 +279,13 @@ export default class MarkersPlugin {
             this.wavesurfer.fireEvent("marker-click", marker, e);
         });
 
+        labelDiv.addEventListener("contextmenu", (e) => {
```
thijstriemstra, line 282: when using `addEventListener`, make sure to `removeEventListener` as well (e.g. in `destroy()` for example).

thijstriemstra (changes requested):
thanks @kpanditCtr 

See inline comments.

Can you also add a changelog entry?

src/plugin/markers/index.js
kpanditCtr: Seems like the rest of the codebase assumes that removing an element from the DOM destroys its event listeners. The only event listeners currently being removed are ones on the document/body. If this is an oversight from earlier, let me know and I will also remove event listeners for mousedown, click etc.

thijstriemstra: >  If this is an oversight from earlier, let me know and I will also remove event listeners for mousedown, click etc.

try creating and `destroy()`-ing the plugin a few times and see if the event listeners get triggered multiple times and increase. I'm almost certain this would happen..

CHANGES.md
```diff
@@ -5,6 +5,7 @@ wavesurfer.js changelog
 ------------------
 - Fix `barMinHeight` option not setting the minimum bar height properly (#2522)
 - Regions plugin: restore support for one drag selection for all channels (#2529)
+- Add support for a context menu event on a marker (#2546)
```
thijstriemstra, line 8: can you use `Markers plugin:`

example/markers/app.js
```diff
@@ -60,6 +61,10 @@ document.addEventListener('DOMContentLoaded', function() {
         console.log("marker drop", marker.label);
     });
 
+    wavesurfer.on('marker-contextmenu', function(marker) {
+        console.log("marker context menu", marker.label);
```
thijstriemstra, line 65: `"marker context menu:"`. If you could add a `:` for the other console logs as well, great (eg. `console.log("marker drop:", marker.label);`)

CHANGES.md
```diff
@@ -5,7 +5,8 @@ wavesurfer.js changelog
 ------------------
 - Fix `barMinHeight` option not setting the minimum bar height properly (#2522)
 - Regions plugin: restore support for one drag selection for all channels (#2529)
-- Add support for a context menu event on a marker (#2546)
+- Markers plugin: 
+    - Add support for a context menu event on a marker (#2546)
```
thijstriemstra, line 9: 2 spaces

thijstriemstra (approved):
thanks!

Issue #1666, opened 2019-06-12
Spectrogram labels when scrolling


Brabudu:
It would be useful to be able to choose, in case of scrolling, if the labels of the spectrogram should move with the spectrogram or be fixed (css position: fixed).
Thank you

thijstriemstra:
@Brabudu if you have a fix for this, please open a pull request!

KerryTucker:
when fix this it goes away from spectrogram position fixed is not working

Iquebal:
When fix the position fixed/sticky label position sticky/fixed is not working.

---
Thanks,
Iquebal

musicog:
@Brabudu You ask for the possibility to choose between fixed and absolute, but to me the current 'fixed' behaviour just seems like a bug - I can't think of a use-case where this would be the intended behaviour. I've submitted PR #2558 that would just always make it absolute. 

Do you have a use-case in mind where 'fixed' would still be required?

For others finding this issue: a temporary work-around before the PR is merged can be achieved with the following CSS:

```
spectrogram .spec-labels {
  position:absolute !important;
}
```

PR #2558, opened 2022-07-21
 spectrogram plugin: fix issue where labels get stuck when scrolling

### LINKED ISSUE #2542: Frequency label display is fixed instead of absolute, which causes it to "float"
### Description
In version 6.2.0, frequency label display is fixed instead of absolute to enable consistent size on hi-dpi displays. As a result, the label "floats" as the page gets scrolled up and down

### Wavesurfer.js version(s):
6.2.0

Chrome version Version 100.0.4896.127

### Code and behaviour needed to reproduce the issue:
See example here: https://wavesurfer-js.org/example/spectrogram/index.html
Behaves similarly in the application I am building.
Screenshot 2022-06-30 at 3 41 17 PM

musicog:
Change display style property on spectrogram labels from 'fixed' to 'absolute' to correct their position when scrolling, to prevent the spectrogram from scrolling away while leaving the labels 'stuck' on their initial screen position.

To replicate the problem, go to https://wavesurfer-js.org/example/spectrogram/index.html and scroll down.

No

No

### Todos/Notes:
Tiny change to fix what appears to clearly be a bug / unintended behaviour

fixes #2542
#1666


thijstriemstra (changes requested):
thanks @musicog 

#1666 requests an option for this. Can you implement this instead? Having `absolute` as default for that new option is fine with me. Also, please add a changelog entry.

musicog:
Happy to try implementing an option for this if it's useful, but right now I can't see a use-case at all for position:fixed - it just seems like an outright bug to me. I'll ask @Brabudu for comment in the issue.

thijstriemstra:
thanks for adding that comment there @musicog. If and when @Brabudu responds, we can see what to do then but I'll merge this for now.

thijstriemstra:
oh, a changelog entry is missing, can you add that @musicog

musicog:
CHANGES.md updated!

thijstriemstra:
> CHANGES.md updated!

Not in this branch it seems..

musicog:
🤦 apologies - now hopefully on the correct branch.

thijstriemstra (approved):

MartinKlimes:
I guess I figured it out.
If you have one waveform + spectrogram, simply use:

```
wavesurfer.on('ready', function(){
     document.querySelector(".spec-labels").style.position = 'absolute'
})
```
In my case i had many waveforms and spectrograms saved into wavesurfer, so:

```
    wavesurfer[index].on('ready', function(){
        var labels = document.querySelectorAll(".spec-labels")
        labels.forEach(label => {
            label.style.position = 'absolute'
          });
    })

```

PR #2606, opened 2022-10-23
cursor plugin: fix crash when destroy triggered before ready

### LINKED ISSUE #2602: cursor plugin: crash when destroy triggered before ready
we don't provide support for React; take a look at https://www.npmjs.com/package/wavesurfer-react instead.

_Originally posted by @thijstriemstra in https://github.com/katspaugh/wavesurfer.js/issues/2563#issuecomment-1286889167_

SilvaQ:
fixes #2602

### DIFF
--- a/src/plugin/cursor/index.js
+++ b/src/plugin/cursor/index.js
@@ -17,7 +17,7 @@
  * the cursor follow the x and the y-position of the mouse. Use `false` to make the
  * it only follow the x-position of the mouse.
  * @property {function} formatTimeCallback Formats the timestamp on the cursor.
- * @property {boolean} isDestroyCalled  true if call destory before ready event
+ * @property {boolean} isDestroyCalled  true if called destroy before the ready event fired
  */
 
 /**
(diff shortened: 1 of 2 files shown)


src/plugin/cursor/index.js
```diff
@@ -322,8 +333,7 @@ export default class CursorPlugin {
      * @returns {string} Formatted timestamp
      */
     formatTime(cursorTime) {
-        cursorTime = isNaN(cursorTime) ? 0 : cursorTime;
-
+        cursorTime = isNaN( cursorTime ) ? 0 : cursorTime;
```
thijstriemstra, line 336: no need to add extra whitespace here

src/plugin/cursor/index.js
```diff
@@ -17,6 +17,7 @@
  * the cursor follow the x and the y-position of the mouse. Use `false` to make the
  * it only follow the x-position of the mouse.
  * @property {function} formatTimeCallback Formats the timestamp on the cursor.
+ * @property {boolean} isDestroyCalled  true if call destory before ready event
```
thijstriemstra, line 20: `true if called destroy before the ready event fired`

thijstriemstra (changes requested):
thanks, see inline comments. can you also add a changelog entry?

Issue #2239, opened 2021-04-05
Zoom performance


mikevercoelen:
The performance of zooming in is really bad, is there any stuff I could do to optimise it?

thijstriemstra:
See #1622 for an attempt to improve zooming performance. Could use some work.

entonbiba:
@mikevercoelen how big is the audio file that you are loading?

I added a debounce feature long time ago for the resize wave part of the code, I think that will help improve the zoom performance if we add it there as well.

https://github.com/katspaugh/wavesurfer.js/blob/2e49044b53680cdd30ec83e65d4fd49b36075e1d/src/wavesurfer.js#L491

Mladja:
First of all, thank you for this library @thijstriemstra. I managed to do a lot of good things. 
The problem I have at the moment is when I try to zoom in on a very large/long file and I see that other people have these problems as well.
Also, the problem is the same with the timeline plugin.

I think a good part of the performance problem is because position => left is used to move the canvas.
This approach is also used on the Timeline plugin.

[Pefrormance post](https://developer.mozilla.org/en-US/docs/Web/Performance/Fundamentals)
[Performance just shorter](https://www.sitepoint.com/atoz-css-translate-vs-position/)

If you were to move Canvas elements via transform instead of left property, do you see any problems with that?
How cursor plugin will behave, seek, and so on...? 

I would love to try tomorrow and do PR if I succeed.

![Screenshot 2021-11-26 at 12 11 16](https://user-images.githubusercontent.com/7488636/143572212-66b8cae2-5705-454e-8ce8-0e859d8bd2eb.png)

thijstriemstra:
try a pr, so we can take a look

Issue #2055, opened 2020-08-29
Performance for long audio files


msjaber:
Hey,

I was experimenting with Wavesrufer for long audio files (more than 60m), and I noticed that the average generated wave is around 300,000px wide. Those pixels are divided into 75 canvas, each of 4000px.

![Screen Shot 2020-08-24 at 4 52 44 PM](https://user-images.githubusercontent.com/1570921/91633989-75277180-e9f5-11ea-98e9-6f0fca2faaa7.png)

This was very heavy for the browser to handle the computation (especially when zooming in and out). 

I was wondering if this the right behavior for long audio files or am I doing something wrong?

Note that, other experiments with [bbc/peaks.js](https://github.com/bbc/peaks.js), I found they reuse the same 2000px wide canvas and only generate the required wave portion, and the performance was excellent.

Thanks

sundayz:
#1622 would fix this problem. Chrome also has a new css property that lets you tell the browser when things should not be rendered, but I dont know if it works on canvas elements, and it's only available in chrome right now: https://web.dev/content-visibility/

I don't know enough about the wavesurfer renderer to say why the same canvas isn't reused. You can also create your own waveform renderer and give it to wavesurfer to use instead of the default renderer, if you want to make a more efficient one (for example using one canvas, or using webgl, etc)

thijstriemstra:
> I don't know enough about the wavesurfer renderer to say why the same canvas isn't reused. 

I think one of the reasons was a dimensions limitation of a canvas element. But not sure if browsers fixed this yet and if it's still something wavesurfer needs to workaround. wavesurfer.js is a collection of a lot of ideas, workarounds and implementations by many people and sometimes these ideas clash and are incompatible, or even obsolete.

PR #2629, opened 2022-12-11
Added more performant way of zooming in (plus some additional wrapper events)

SenshiSentou:
The main reason for this PR is the changes I made to the zooming code, resulting in much better performance for large audio files, with minimal visual sacrifice.

**However**, I'm not sure if this would be the canonical way of doing this. For one, I merged horizontal and vertical zoom in one. Vertical zoom might also be better off as its own `param` option, and should probably (?) be returned by the `zoom` event as well, alongside the new `minPxPerSec`.

Also, as this is a result of tinkering for my own project, there is an additional change in her that adds a few more wrapper events and refactored that code a little.

As such, I don't expect this PR to be accepted as is, but my hope is someone more involved in the project can take it and run with it :)

Zoom-related:

- Added `maxPxPerSec` to `params`. This would be the resolution at which to render the original waveform to be used for scaling
- Updated `waveSurfer.zoom()` to take an extra parameter for the vertical scale
- `waveSurfer.zoom()` now defers to `drawer.updateZoom()`
- In the MultiCanvas renderer, zooming now scales using the canvas entries' `wave` and `progress`'s styles:
    - `left` and `width` are used for horizontal scaling
    - `transform` is used for vertical scaling

Event-related:

- Added dedicated `setupScrollbarAwareWrapperEvent()` function
- Use it to setup `click`, `mousedown`, `mousemove` and `mouseup` events

Only tested with the `regions` plugin, but should be good I think.

`waveSurfer.zoom()`

### Todos/Notes:
- Potentially make `verticalScale` a `params` option
- `zoom` event should return an object containing both the new `minPxPerSec` and `verticalScale`
- Actually draw at `maxPxPerSec` instead of `minPxPerSec` unless specified

#2239, #2055, #1622 (tangentially; tries to solve the same issue a different way)

Hope this can help some!

### DIFF
--- a/src/drawer.canvasentry.js
+++ b/src/drawer.canvasentry.js
@@ -86,10 +86,7 @@ export default class CanvasEntry {
      */
     initProgress(element) {
         this.progress = element;
-        this.progressCtx = this.progress.getContext(
-            '2d',
-            this.canvasContextAttributes
-        );
+        this.progressCtx = this.progress.getContext('2d', this.canvasContextAttributes);
     }
 
     /**
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "src/"
   ],
   "scripts": {
-    "start": "npm run build:plugins && npm run build:htmlinit && webpack serve --config ./build-config/webpack.dev.main.js",
+    "start": "npm run build:plugins && npm run build:htmlinit && nodemon --watch src --exec \"npm run build:normal && cp ./dist/wavesurfer.js ~/Projects/Secare/external/wavesurfer.js\"",
     "start:htmlinit": "npm run build:main && npm run build:plugins && webpack serve --config ./build-config/webpack.dev.htmlinit.js",
     "start:plugins": "npm run build:main && npm run build:htmlinit && webpack serve --config ./build-config/webpack.dev.plugins.js",
     "build": "npm run build:normal && npm run build:minified",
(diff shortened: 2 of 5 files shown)


src/drawer.canvasentry.js
```diff
@@ -86,10 +86,7 @@ export default class CanvasEntry {
      */
     initProgress(element) {
         this.progress = element;
-        this.progressCtx = this.progress.getContext(
-            '2d',
-            this.canvasContextAttributes
-        );
+        this.progressCtx = this.progress.getContext('2d', this.canvasContextAttributes);
```
thijstriemstra, line 89: no need for this formatting only change.

SenshiSentou: Like I said, feel free to ignore the parts that don't apply. I just wanted to throw this solution out there; I'm not expecting this PR to make it to master as-is

SenshiSentou:
Coming back to this today, it seems there's still an issue with the playback timing. It starts out correct, but shifts/ gets out of sync over time. I tried zeroing out the "overlap" but it still occurs and I'm not sure why... I'll keep playing with this, but any suggestions would be very welcome

entonbiba:
@SenshiSentou any update on this? What are some of the issues you running into, maybe we can help as well.

katspaugh:
Closing in favor of #2646.

Thanks @SenshiSentou!

PR #2641, opened 2023-01-09
regions plugin: improved delta calculation (resize end)

BQTH:
With the use case that I use this library for, users need to be able to resize the regions. When a region was resized and afterward set back to the min length (f.e. 3.1 sec or in HMSM, which is how we show it, 00:00:03:10) it would stay at 3.12sec (00:00:03:12) or 3.14sec (00:00:03:14) or something, but never go back down all the way to 3.1 (00:00:03:10). With these changes, the length of the region can be set back to 3.10 exactly

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,7 @@
 wavesurfer.js changelog
 =======================
 
-6.4.1 (12.01.2023)
+6.5.0 (unreleased) (12.01.2023)
 ------------------
 - Regions plugin:
   - Improved delta calculation (resize end) (#2641)
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 wavesurfer.js changelog
 =======================
 
+6.4.1 (12.01.2023)
+------------------
+- Regions plugin:
+  - Improved delta calculation (resize end) (#2641)
+
 6.4.0 (05.11.2022)
 ------------------
 - Markers plugin:
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,7 +1,7 @@
 wavesurfer.js changelog
 =======================
 
-6.5.0 (unreleased) (12.01.2023)
+6.5.0 (unreleased)
 ------------------
 - Regions plugin:
   - Improved delta calculation (resize end) (#2641)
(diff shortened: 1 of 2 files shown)


src/plugin/regions/region.js
```diff
@@ -586,6 +586,8 @@ export class Region {
         const onMove = (event) => {
             const duration = this.wavesurfer.getDuration();
             let orientedEvent = this.util.withOrientation(event, this.vertical);
+
```
thijstriemstra, line 589: no need for this newline

src/plugin/regions/region.js
```diff
@@ -641,9 +646,10 @@ export class Region {
                 }
             }
 
-            let delta = time - startTime;
+             if (!delta) delta = time - startTime;
```
thijstriemstra, line 649: put if statement on multiple lines

src/plugin/regions/region.js
```diff
@@ -641,9 +646,10 @@ export class Region {
...
             startTime = time;
 
+
```
thijstriemstra, line 652: no need for this newline

thijstriemstra (changes requested):
thanks, see inline comments. can you also add a changelog entry?

BQTH:
> thanks, see inline comments. can you also add a changelog entry?

Good morning, i've made the requested changes and added an entry to the changelog. This is my first time collaborating on a repo like this, hopefully i did it correctly. All feedback is welcome!

CHANGES.md
```diff
@@ -1,6 +1,11 @@
 wavesurfer.js changelog
 =======================
 
+6.4.1 (12.01.2023)
```
thijstriemstra, line 4: you can use 6.5.0 (unreleased) instead

thijstriemstra (changes requested):

CHANGES.md
```diff
@@ -1,7 +1,7 @@
 wavesurfer.js changelog
 =======================
 
-6.4.1 (12.01.2023)
+6.5.0 (unreleased) (12.01.2023)
```
thijstriemstra, line 4: and remove the date ;)

BQTH: Done!

thijstriemstra (approved):
thanks!

PR #2671, opened 2023-02-26
Docs: update the issue template and wavesurfer links

katspaugh:
Update the issue template to encourage people to use the Discussions forum instead of issues for questions and support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,12 @@
 # [wavesurfer.js](https://wavesurfer-js.org)
 
 [![npm version](https://img.shields.io/npm/v/wavesurfer.js.svg?style=flat)](https://www.npmjs.com/package/wavesurfer.js)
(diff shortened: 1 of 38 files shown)


README.md
```diff
@@ -10,25 +10,29 @@ Interactive navigable audio visualization using Web Audio and Canvas.
...
+# or
+
+yarn add wavesurfer.js
 
-### Can drawing be done as file loads?
-No. Web Audio needs the whole file to decode it in the browser. You can however load pre-decoded waveform data to draw the waveform immediately. See here: https://wavesurfer-js.org/example/audio-element/ (the "Pre-recoded Peaks" section).
+# or
 
-## API in examples
+<script src="https://cdnjs.cloudflare.com/ajax/libs/wavesurfer.js/6.4.0/wavesurfer.min.js" crossorigin="anonymous" referrerpolicy="no-referrer"></script>
```
thijstriemstra, line 27: - This means we need to update the version number everytime a new release is made
- We always used unpkg.com instead of cloudflare (which allows you to use the latest version instead of specifying a version nr)

thijstriemstra (changes requested):
Moving the project to a new URL also means any old https://github.com/katspaugh/wavesurfer.js links should be updated to https://github.com/wavesurfer-js/wavesurfer.js. There are many in the examples directory for example.

README.md
katspaugh: Sounds good, will change this.

thijstriemstra (approved):

PR #2686, opened 2023-02-27
Fix iphone silent switch webaudio mute

### LINKED ISSUE #2667: ignoreSilenceMode on iPhone - audio cuts out after a short time
Hi there!

I have noticed a bug on iPhone when `ignoreSilenceMode` is enabled and you play the audio with silent mode switched on then the audio plays for 2 or 3 seconds before the audio stops. I also reported this [here](https://github.com/katspaugh/wavesurfer.js/pull/2176#issuecomment-1398122664) but thought it needed an issue.

### Wavesurfer.js version(s):
I'm using wavesurfer.js v6.4.0

iPhone 13 Pro
iOS 16.2
Chrome 105.0.5195.147

https://codesandbox.io/s/reverent-dust-xtl33i?file=/src/AudioPlayer.js

Navigate to https://xtl33i.csb.app/ on an iPhone with silent mode switched on then click the play button when the audio has loaded and the waveform is visible. Observe that the audio cuts out after a short time.

entonbiba:
fixes #2667 

adds a 1 second webaudio context buffer with 44100 bit rate to allow audio to play even if iphone silent switch is turned on.

preview here using latest wavesurfer.js version 6.4.0
https://www.entonbiba.com/lab/wavesurfer/2667.html

closes #2667

### DIFF
--- a/src/util/silence-mode.js
+++ b/src/util/silence-mode.js
@@ -11,8 +11,8 @@
  */
 export default function ignoreSilenceMode() {
     // Set webaudio context with 1 second silent audio 44100 bit rate buffer to allow playing audio even if silent switch is on the device
-    let silentAC = new AudioContext();
-    let silentBS = silentAC.createBufferSource();
+    const silentAC = new AudioContext();
+    const silentBS = silentAC.createBufferSource();
     silentBS.buffer = silentAC.createBuffer(1, 1, 44100);
     silentBS.connect(silentAC.destination);
     silentBS.start();
--- a/src/util/silence-mode.js
+++ b/src/util/silence-mode.js
@@ -10,6 +10,13 @@
  * @since 5.2.0
  */
 export default function ignoreSilenceMode() {
+    // Set webaudio context with 1 second silent audio 44100 bit rate buffer to allow playing audio even if silent switch is on the device 
+    let silentAC = new AudioContext();
+    let silentBS = silentAC.createBufferSource();
+    silentBS.buffer = silentAC.createBuffer(1, 1, 44100);
+    silentBS.connect(silentAC.destination);
+    silentBS.start();
+
     // Set the src to a short bit of url encoded as a silent mp3
     // NOTE The silence MP3 must be high quality, when web audio sounds are played
     // in parallel the web audio sound is mixed to match the bitrate of the html sound
--- a/src/util/silence-mode.js
+++ b/src/util/silence-mode.js
@@ -10,7 +10,7 @@
  * @since 5.2.0
  */
 export default function ignoreSilenceMode() {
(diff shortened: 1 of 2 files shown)


entonbiba:
@thijstriemstra updated,

src/util/silence-mode.js
```diff
@@ -10,6 +10,13 @@
  * @since 5.2.0
  */
 export default function ignoreSilenceMode() {
+    // Set webaudio context with 1 second silent audio 44100 bit rate buffer to allow playing audio even if silent switch is on the device
+    let silentAC = new AudioContext();
+    let silentBS = silentAC.createBufferSource();
```
katspaugh, line 15: The `let` can be changed to `const` since you're not modifying these variables further on.

katspaugh (commented):
@entonbiba thanks for the fix!

Could you elaborate a little bit in what situations this fixes the problem? Do I get it right that it's only for the WebAudio backend? The function seems to have worked before, but I guess only for the MediaElement?

thijstriemstra (approved):

src/util/silence-mode.js
entonbiba: @katspaugh updated, yes the fix for mediaelement worked for webaudio as well to allow audio to play based on user interaction. However on latest updates of ios if the silent switch is on it will detect it's playing a mediaelement source so it blocks webaudio as it's not technically triggered by user interaction. To fix this a webaudio buffer needs to be played prior as well as mediaelement so it detects webaudio is played allowing the audio to continue.

katspaugh: Excellent, thank you!

WickeyZ:
This fix seems to be removed in the current version of wavesurfer.

But with the WebAudio being brought back now (https://github.com/katspaugh/wavesurfer.js/issues/3272), this issue seems to be occuring again in iOS

(I'm on iOS version 18.4, using `"wavesurfer.js": "^7.9.5"` and `"@wavesurfer/react": "^1.0.11"`)

---
My fix:

This seems to be a long standing reported bug of iOS - https://bugs.webkit.org/show_bug.cgi?id=237322
I can actually fix it with this snippet according to one of the comments in the post
```
    if (navigator.audioSession) {
      navigator.audioSession.type = "playback";
    }
```

Keep in mind to refer to docs of this API: [audioSession API](https://developer.mozilla.org/en-US/docs/Web/API/Navigator#:~:text=No-,audioSession,-Experimental), [audioSession type](https://www.w3.org/TR/audio-session/#audio-session-types)

katspaugh:
Happy to reapply the workaround. `navigator.audioSession.type` looks like a cleaner fix, though.

PR #2689, opened 2023-02-27
audio effects

entonbiba:
- create audio effects to allow for effects such as fade in when audio is starting to play.
- current pr allows for the following two options but can be enhanced to allow a group of effects. Other effects should be added prior to merging...

```
fadeInAudio: true,
fadeInLength: 10000,
fadeOutAudio: true,
fadeOutLength: 10000
```

todo
- [x] add fade in audio
- [ ] add fade out audio
- [ ] other audio effects?
- [ ] allow multiple effects to be applied in options?

discussion #2688

### DIFF
--- a/src/util/audioEffect.js
+++ b/src/util/audioEffect.js
@@ -1,0 +1,30 @@
+/**
+ * Apply audio effects
+ *
+ * @param {Object} effects The map of available properties per effect
+ *
+ */
+export default function audioEffect(effect) {
+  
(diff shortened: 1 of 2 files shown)


katspaugh:
@entonbiba do you think this could be something better handled in a plugin maybe?

katspaugh:
@entonbiba I suggest we close this PR then. Please feel free to make a new one with a plugin. Cheers!

PR #2699, opened 2023-03-03
Update minPxPerSec docs

katspaugh:
It wasn't trivial to figure out why this option didn't do anything on its own.


entonbiba (approved):
correct,

PR #2706, opened 2023-03-11
Refactor: remove deprecated ScriptProcessorNode

### LINKED ISSUE #1281: Replace the deprecated ScriptProcessorNode with AudioWorkletNode
`ScriptProcessorNode` is deprecated from the specification and replaced with `AudioWorklet` in Chrome 64. AudioWorklet is enabled by default for Chrome 66 or later. In Chrome 64 and 65, the feature is behind the experimental flag.

See https://developers.google.com/web/updates/2017/12/audio-worklet and even better: https://developers.google.com/web/updates/2018/06/audio-worklet-design-pattern

katspaugh:

Resolves #1281. Removes the deprecated [ScriptProcessorNode](https://developer.mozilla.org/en-US/docs/Web/API/ScriptProcessorNode).

The only purpose of this node in our code was to track time in the WebAudio backend. It's now being done via `requestAnimationFrame` instead. It's a little less precise, but much lighter on the CPU and good enough for rendering.

There are no breaking API changes. It's still possible to pass and external scriptNode in the options, but it won't be used for timing. It will be just added to the graph like before.

We should eventually deprecate the option, too.

### DIFF
--- a/src/webaudio.js
+++ b/src/webaudio.js
@@ -8,15 +8,11 @@ const FINISHED = 'finished';
 /**
  * WebAudio backend
  *
- * @extends {Observer}
+ * @extends {util.Observer}
  */
 export default class WebAudio extends util.Observer {
-    /** scriptBufferSize: size of the processing buffer */
-    static scriptBufferSize = 256;
     /** audioContext: allows to process audio with WebAudio API */
     audioContext = null;
-    /** @private */
-    offlineAudioContext = null;
     /** @private */
     stateBehaviors = {
         [PLAYING]: {
@@ -33,7 +29,6 @@ export default class WebAudio extends util.Observer {
         },
         [PAUSED]: {
             init() {
-                this.removeOnAudioProcess();
             },
             getPlayedPercents() {
                 const duration = this.getDuration();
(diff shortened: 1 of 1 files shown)


thijstriemstra:
> We should eventually deprecate the option, too.

Let's start with that, deprecating the functionality instead of removing it; why skip this step?

katspaugh:
@thijstriemstra sorry, I was probably not clear as to what I'm removing and what I'm suggesting to deprecate.

There were two scriptNode-related things:
* we used an internally created scriptNode to track time (aka audioprocess) – _this is removed_
* there's also an option to pass an externally created scriptNode – _this remains_ and I suggest to deprecate it (in another PR)

katspaugh:
@thijstriemstra do you have any comments on the code?

katspaugh:
Closing because no feedback from the requester.

katspaugh:
On another thought, I’ll go ahead and merge this. Having a script node just as a hi frequency timer makes very little sense in general, but now that it’s deprecated, it’s even worse.

thijstriemstra:
> On another thought, I’ll go ahead and merge this.

I am unable to make a commit without a review, why has this changed? And why does this not apply to you?

thijstriemstra:
> Because unfortunately things aren't moving forward otherwise.

The wavesurfer.js project changes in the last weeks have made me seriously reconsider spending anymore time on this project. I hope I'm just over-reacting.

katspaugh:
@thijstriemstra please shoot me an email at katspaugh@gmail.com if you want to talk about this.

PR #2724, opened 2023-03-15
Docs: add a video tutorial link to the readme

katspaugh:
There are quite a few video tutorials about wavesurfer.js on youtube.
I picked one from the first page that I think is very good. Kudos to the author!


entonbiba:
awesome, maybe I'll do a couple tutorials in the upcoming months as well something like short reels etc.

katspaugh:
That'd be amazing, @entonbiba!

PR #2734, opened 2023-03-18
Fix: avoid exit 1 in CI script

### LINKED ISSUE #2723: Fix the automatic publishing on NPM
The script that published the NPM package is broken after #2710.

https://github.com/wavesurfer-js/wavesurfer.js/actions/runs/4419688300/jobs/7748430370

Looks like this part was necessary after all:

```
      - uses: actions/setup-node@v3
         with:
           node-version: 16
           registry-url: https://registry.npmjs.org/
```

katspaugh:
Resolves #2723.

The release workflow intentionally fails when there's no new version to release, but this might look confusing because it looks like it failed due to an error.

Instead of `exit 1`, we're now checking for the new version on each step and exit w/o an error.

Also NPM publish actually wasn't working. @thijstriemstra as you noticed correctly in the dry-run, it was showing an error. Fixed it by adding a setup-node step.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -30,24 +30,30 @@ jobs:
             echo "Version $OLD_VERSION hasn't changed, skipping the release"
           fi
 
-      - name: Create a git tag
-        if: ${{ steps.version.outputs.version }}
-        run: git tag $NEW_VERSION && git push --tags
-        env:
-          NEW_VERSION: ${{ steps.version.outputs.version }}
+      # - name: Create a git tag
+      #   if: ${{ steps.version.outputs.version }}
+      #   run: git tag $NEW_VERSION && git push --tags
+      #   env:
+      #     NEW_VERSION: ${{ steps.version.outputs.version }}
 
-      - name: GitHub release
+      # - name: GitHub release
+      #   if: ${{ steps.version.outputs.version }}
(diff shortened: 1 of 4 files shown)


katspaugh:
@thijstriemstra it should be good now, I've tested both when there's a new version and when there's none. Made a 6.6.1 release from this branch.
Sorry about the back and forth on this.

thijstriemstra:
I wonder why the release script creates a git tag;

```
  - name: Create a git tag
        if: ${{ steps.version.outputs.version }}
        run: git tag $NEW_VERSION && git push --tags
        env:
          NEW_VERSION: ${{ steps.version.outputs.version }}
```

it would make more sense to me to let the maintainer create a tag, push it and then the publish script (that only runs on a certain tag name structure, e.g. release-1.2.3 which extracts the version nr from the tag name (if needed). This also adds a layer of permission control where only maintainers are allowed to create and push a tag.

katspaugh:
I like that in principle. It won't add anything in terms of permission control tho, as anyone who's able to approve a PR with a version update is also able to push a tag.
Also, I think a release PR would still be requited because you need to update the changelog and push it to main, _prior_ to tagging it. So I think it's easier to just run `yarn release`, make a release PR and have it approved by a maintainer.

thijstriemstra (approved):

PR #2741, opened 2023-03-24
Release 6.6.2

### LINKED ISSUE #2738: [Issue] Playhead out-of-sync with Waveform on long audio files
### Discussed in https://github.com/wavesurfer-js/wavesurfer.js/discussions/2737

<sup>Originally posted by **kai-shimada** March 21, 2023</sup>
Hello Community!

As I'm developing an Electron transcription app on Mac (WhisperScript), I am facing an issue when navigating the playhead via mouse click in long audio files (>15min).

I noticed that the playhead is not getting placed correctly where I position the cursor. The gap between the cursor position and the placed playhead gets bigger the more you seek into the audio file (see the video). If I set the seek point programmatically using SeekTo, the playhead gets placed correctly. I also noticed that the width of the < wave > or the container is getting calculated incorrectly as it's wider than the actual length of the audio (see the gap after it reaches the end of the audio).

Has anyone experienced the same issue?

https://user-images.githubusercontent.com/79719366/226669156-9b0ebd0d-3b72-41a8-b7f5-7720edfcc41b.mov

I am seeing this issue on WAV 48khz 24 bit as well as on MP3 44,1khz audio files... (it worked with none of the files I tested)

WaveSurfer Version: `6.6.1`
WaveSurfer Option:
```javascript
{
    ...
    backend: 'MediaElement',
    autoCenterImmediately: true,
    responsive: 80,
    normalize: true,
    minPxPerSec: 20,
    splitChannels: false,
    plugins: [
        RegionsPlugin.create({
            edgeScrollWidth: 3,
            dragSelection: true,
            minLength: 0.1,
            showTooltip: false,
            contentEditable: true,
            removeButton: true,
        })
    ]
}
```
I'd appreciate any help! thank you 🙏

</div>

katspaugh:
Resolves #2738.

As @kai-shimada found out, v6.6.0 introduced a regression in width calculation. The only likely cause is #2646, so I suggest we revert it for now.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 wavesurfer.js changelog
 =======================
 
+6.6.2 (24.03.2023)
+------------------
+- Revert "Zoom optimisation for Waves and matching implementation for Spectrograms (#2646)"
+- Fix: avoid exit 1 in CI script (#2734)
+
 6.6.1 (18.03.2023)
 ------------------
 - Fix: NPM publish in the CI job (#2727)
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "wavesurfer.js",
-  "version": "6.6.1",
+  "version": "6.6.2",
   "homepage": "https://wavesurfer-js.org",
   "authors": [
     "katspaugh <katspaugh@gmail.com>"
--- a/example/spectrogram/app.js
+++ b/example/spectrogram/app.js
@@ -22,6 +22,15 @@ function initAndLoadSpectrogram(colorMap) {
         ]
     };
 
+    if (location.search.match('scroll')) {
+        options.minPxPerSec = 100;
+        options.scrollParent = true;
+    }
+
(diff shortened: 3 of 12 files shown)


katspaugh:
If approved, I'll publish 6.6.2 from this branch which doesn't include #2706. It can go live in some other release.

thijstriemstra:
> If approved, I'll publish 6.6.2 from this branch which doesn't include https://github.com/wavesurfer-js/wavesurfer.js/pull/2706. It's currently on master, but it can go live in some other release.

Can you explain in detail? Why do releases have to be so complicated lately? Don't you agree?

katspaugh:
We have only one master branch that serves as both our development branch (people make PRs to it), and the branch we make releases from.
Since master now has a commit not intended for the release, I suggest to make a release from this branch instead, which is an earlier snapshot of master + the revert commit.

thijstriemstra (changes requested):
Please split releases and major code changes into separate pull requests, thanks.

PR #2743, opened 2023-03-25
feat(MarkersPlugin):  add getMarkers function

FalkF:
Add a function to get all current markers to the wavesurfer object.

None I guess.

None I guess.

### Todos/Notes:
Untested.
I also would like to add it to the regions plugin if this is ok.
This way I dont need to keep track of the state outside of wavesurfer.
Great tool btw, using it for https://gitlab.com/fraggen/fraggen


katspaugh (approved):
LGTM, thanks!

PR #2746, opened 2023-03-26
Docs: version the unpkg script in readme

katspaugh:
With the unpkg CDN, when the version is omitted, it will fetch the latest version published on NPM.

In our readme, we provide a CDN tag with the an unspecified version, as a way to keep users up-to-date.

However, if we release a major version update, e.g. 7.0.0, people who copied and pasted the unpkg tag from our readme, will have wavesurfer.js updated to that version, even though a major version update implies breaking changes.

Capping the version at `wavesurfer.js@6` would prevent users from unknowingly auto-upgrading with _intentional_ breaking changes.

Beaking changes or regressions can also occur _unintentionally_ (as was the case with 6.6.0), so pinning the minor version would be the safest (while still fetching bug fixes automatically).

The tag in the readme is obviously just a recommendation, and users can choose a different versioning strategy in their app.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ import WaveSurfer from 'wavesurfer.js'
 
 If you're not using a package manager, simply insert the script from a CDN:
 ```
-<script src="https://unpkg.com/wavesurfer.js"></script>
+<script src="https://unpkg.com/wavesurfer.js@6.6"></script>
 ```
 
 Create a container in your HTML:


thijstriemstra (changes requested):
Does this mean that whenever a new release is added, this needs to be updated? Is there a checklist for releases? I don't like these type of edits that have to be made with a release because 99% of the time these things are forgotten. I think 99% of the users want the latest release anyway.

katspaugh:
It doesn't have to be updated on every release, the whole point of it is to only update it when a version is deemed stable.
I had a user reporting a major breakage of their website because of an update they didn't expect. The least technical people who use the CDN tag are also the ones that probably want to be as stable as possible. More technical users will upgrade to the latest when needed.

Since you're requesting changes, can you please clarify what changes you're expecting to be made?

thijstriemstra (commented):
> Since you're requesting changes, can you please clarify what changes you're expecting to be made?

No changes requested, it's unfortunately another change we disagree about. Perhaps you can get a review from someone else, it's -1 from me.

katspaugh:
I get your point of you that you don't want to update the version in the readme.
The problem with always fetching the latest version is that it's done in runtime. You insert that script, and your app works. Then, a few months later, a new major version comes out, and suddenly your app breaks.
Compare to NPM: even if you install the latest version, it will do it only once. If you run install again, the default behavior is to only update the patch version, but not the minor version. With an un-versioned script tag, you're basically inserting a ticking bomb.
This PR prevents you from having unexpected changes in your app, even if it means we'll have to update the readme once in a while. But again, we don't have to do it very often.

PR #2760, opened 2023-04-13
Docs: add a note about wavesurfer.ts

katspaugh:
Encourage people to try out the alpha version. I'd like to gauge how compatible it is with the existing applications and whether it's missing some features folks need.


prakharpbuf (approved):

katspaugh:
Absolutely, in due time. For now it’s just an alpha, and the API can still change.

PR #2761, opened 2023-04-14
fix: relay dblclick event

prakharpbuf:
fixes #2759


katspaugh (approved):
Nice, thank you @prakharpbuf!

PR #2793, opened 2023-05-28
Use relative paths for dist exports in package.json

### LINKED ISSUE #2792: v7 beta usage with React – import error
I'm trying to use the new version in my react app. I am getting the following error.

```
Error: Invalid "exports" main target "dist/wavesurfer.js" defined in the package config/node_modules/wavesurfer.js/package.json imported from '/pages/home.js'; targets must start with "./"
```

this is how I am using it

```ts
import Wavesurfer from 'wavesurfer.js';

const Waveform = ({ url }: { url: string }) => {
  useEffect(() => {
    const wavesurfer = Wavesurfer.create({
      container: '#waveform',
      waveColor: '#4F4A85',
      progressColor: '#383351',
      url: '/audio.mp3',
    });
  }, []);

  return <div id="waveform"></div>;
};

export default Waveform;
```

malimccalla:
fixes #2792 

This fixed the issue I was having. As soon as I opened in my code editor I was prompted with the error

```
String does not match the pattern of "^\./".
```

This change resolved it and the error I was getting in my app

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,16 +20,16 @@
   "browser": "dist/wavesurfer.js",
   "exports": {
     ".": {
-      "import": "dist/wavesurfer.js",
-      "require": "dist/wavesurfer.min.js"
+      "import": "./dist/wavesurfer.js",
+      "require": "./dist/wavesurfer.min.js"
     },
     "./dist/plugins/*": {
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
Awesome, thank you!

PR #2797, opened 2023-05-28
Convert color values from array to lineargradient

jgentes:

I noticed someone had filed an issue on the Beta version to allow for an array of colors for the `waveColor` option. I was using this approach too, so added a helper here.

None, it continues to support a standard string color, a gradient, an empty array, an array with one color value, or multiple values, or `undefined`.

None

### Todos/Notes:

I didn't implement this for the `Record` plugin. I wasn't sure if you'd prefer to duplicate code or somehow reference it, but I didn't see other imports so I opted to just leave it out of this PR.

There was one on `wavesurfer.ts` but gone now :)


katspaugh:
Nice! Could you also update the type to `string | string[]`? Thank you!

jgentes:
done!

katspaugh (approved):
Thanks!

PR #2801, opened 2023-05-30
Multitrack+ allows separate containers to be defined for each track :)

jgentes:

The current multitrack feature requires a single container for all tracks. This change allows each track to specify a container and will sync audio operations across them, or if a single container is specified, it will use the current functionality of combining them into a single container.

None

I don't think there are any, although drag 'n drop isn't working (but this is consistent with main)

### Todos/Notes:

The iniital cursor shows up at zero time for each track, then adjusts when played. That's probably an easy fix, I haven't looked into it. I'm a few beers in at this point :)

### DIFF
--- a/examples/multitrack.js
+++ b/examples/multitrack.js
@@ -20,9 +20,11 @@ const multitrack = Multitrack.create(
   [
     {
       id: 0,
+      container: document.querySelector('#track0'),
     },
     {
       id: 1,
+      container: document.querySelector('#track1'),
       draggable: false,
       startPosition: 14, // start time relative to the entire multitrack
       url: '/examples/audio/librivox.mp3',
@@ -64,6 +66,7 @@ const multitrack = Multitrack.create(
     },
     {
       id: 2,
+      container: document.querySelector('#track2'),
       draggable: true,
       startPosition: 1,
       startCue: 2.1,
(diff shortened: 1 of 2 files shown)


jgentes:
oh, no tests.. :(

katspaugh:
Hey @jgentes, thanks for the PR!

Multitrack was a custom job and a pretty complex app as you surely noticed, so I wouldn't like to modify it at this point, unless absolutely required.

Syncing wavesurfers across a web page is a cool idea though. Perhaps worthy of a separate plugin?

One could pass the same Sync plugin instance to multiple wavesurfer instances like this:

```js
const sync = SyncPlugin.create({ ... })

const wavesurferA = WaveSurfer.create({
  container: '#containerA',
  url: urlA,
  plugins: [ sync ],
})

const wavesurferB = WaveSurfer.create({
  container: '#containerB',
  url: urlB,
  plugins: [ sync ],
})
```

Cheers!

jgentes:
Thanks, a plugin would be a good approach. Should I just follow the example of multitrack or is there a better example of how to build a plugin? And I assume people would import it from my repo which I publish to npm, or should I submit a PR to include it in the base wavesurfer?

katspaugh:
Multitrack isn't really a plugin, it's more of a separate app that manages multiple instances of wavesufer.

For a Sync plugin, the Timeline plugin would be a good place to start. You basically need to extends the BasePlugin class and implement a few methods like `constructor`, `onInit` and `destroy`. Please let me know if you have any questions down the road.

> And I assume people would import it from my repo which I publish to npm, or should I submit a PR to include it in the base wavesurfer?

Entirely up to you! If you choose to add it to the core package, please make sure to add your name on top of the plugin file. 🙂

PR #2819, opened 2023-06-01
Feat: custom render function

katspaugh:
I've added a new option `renderFunction: (peaks, ctx) => void` which allows swapping out the default rendering function for a custom one.

Screenshot 2023-06-01 at 08 58 12

Resolves https://github.com/katspaugh/wavesurfer.js/discussions/2811

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -225,6 +225,68 @@ class Renderer extends EventEmitter<RendererEvents> {
     return gradient
   }
 
+  private renderBars(
+    channelData: Array<Float32Array | number[]>,
+    options: WaveSurferOptions,
+    ctx: CanvasRenderingContext2D,
+  ) {
+    ctx.fillStyle = this.convertColorValues(options.waveColor)
+
+    // Custom rendering function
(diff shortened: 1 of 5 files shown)


src/renderer.ts
```diff
@@ -225,94 +225,97 @@ class Renderer extends EventEmitter<RendererEvents> {
...
-    const useNegative = isMono && rightChannel.some((v: number) => v < 0)
-    const length = leftChannel.length
+    ctx.fillStyle = this.convertColorValues(options.waveColor)
 
-    const barCount = Math.floor(width / (barWidth + barGap))
-    const barIndexScale = barCount / length
-    const halfHeight = height / 2
+    // Custom rendering function
+    if (options.renderFunction) {
+      options.renderFunction(channelData, ctx)
```
euvl, line 237: Amazing, exactly what i needed in https://github.com/katspaugh/wavesurfer.js/discussions/2815

katspaugh: Not sure I follow

PR #2841, opened 2023-06-05
Feat: use parent height if height not set

### LINKED ISSUE #2837: Issue: wavesurfer.js v7@beta - fillParent not working as expected
I'm using wavesurfer.js v7@beta and have encountered an issue where the `fillParent` property doesn't work as expected.
The problem is that the wavesurfer child elements have a default height of 128px instead of occupying 100% height when `fillParent` is set to `true`. This default height overrides the parent's 100% height setting.

To address this issue, we need to fix the height of the wavesurfer child element so that it properly fills the parent container when `fillParent` is enabled.

Here is a demo showcasing the issue: [Demo](https://codesandbox.io/s/festive-taussig-u815jh?file=/src/App.js).

katspaugh:
Resolves #2837

It will now use the container height if a height isn't passed in the options, otherwise fall back to 128 px.


src/renderer.ts
```diff
@@ -93,6 +95,11 @@ class Renderer extends EventEmitter<RendererEvents> {
     )
   }
 
+  private getHeight(): number {
+    const defaultHeight = 128
+    return this.options.height ?? (this.parent.clientHeight || defaultHeight)
```
IdoBouskila, line 100: In your code you are using the default height (128px) only if `height` doesn't provided in the options and parent.clientHeight is `null` or `undefined`
I think it should be: if `height` isn't provided AND `fillParent` is `true`, then don't use the default value

katspaugh: `fillParent` was meant only for width. I think it’s sufficient that you either pass a height or you have it fill the container. Otherwise the waveform will be just invisible, which is not very useful, is it?

IdoBouskila: I don't think so. I think its very useful for responsiveness in scenarios that the parent height is in relative units and you need the wave to be in the same height as the parent.

katspaugh: So I’ll leave as it is then. If someone does NOT want to it to fill the full height, they can pass a specific height.

IdoBouskila: Ok, your choice :)

IdoBouskila: Ok, your choice :)

PR #2843, opened 2023-06-05
Fix: webm -> wav for iOS Safari

### LINKED ISSUE #2836: Playback after recording not working on a Safari ios 16
After getting access rights and recording audio i can see a waveform, but there is no sound when playing.
In Desktop Chrome everything is ok.

### Wavesurfer.js version(s):
beta 7

Ios 16 Safari
Mac os Safari 16.5

Example have a same issue on Safari: 
https://wavesurfer-js.org/examples/#record.js

### Steps needed to reproduce the issue:
Go to https://wavesurfer-js.org/examples/#record.js -> start to record -> stop recording -> try to play the recording

katspaugh:
Resolves #2836.

### DIFF
--- a/src/plugins/record.ts
+++ b/src/plugins/record.ts
@@ -23,7 +23,7 @@ class RecordPlugin extends BasePlugin<RecordPluginEvents, RecordPluginOptions> {
   }
 
   private loadBlob(data: Blob[]) {
-    const blob = new Blob(data, { type: 'audio/webm' })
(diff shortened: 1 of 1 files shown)


iiKulakov:
Any thoughts why?

katspaugh:
It has to do with audio encoding probably. Or the fact that we load a blob. We need to experiment with different options to see what works on iOS.

Issue #2844, opened 2023-06-05
Dragging on timeline: interaction event handler no longer reports current time


brollin:
### Wavesurfer.js version(s):
`^7.0.0-beta.6`

### Browser and operating system version(s):
Firefox 113.0.2, macOS Monterey 12.6

### Code needed to reproduce the issue:
```javascript
wavesurfer.on("interaction", () => {
  console.log("interaction", wavesurfer.getCurrentTime());
});
```

### Use behaviour needed to reproduce the issue:
After attaching the above event listener, click and drag on the timeline. I'm noticing that the alpha version of wavesurfer will report the time that you have dragged to when `getCurrentTime()` is called, but the beta version of wavesurfer will report the time that you last clicked (mouseup'ed I suppose?) when `getCurrentTime()` is called.

brollin:
@katspaugh just looking into this a bit more, and it seems related to the debouncing of drag interaction events:
https://github.com/katspaugh/wavesurfer.js/blob/f3b6ac435d7189cb24a79bf07643bf0281e79eed/src/wavesurfer.ts#L224-L231

When editing locally, I'm finding that I'm only achieving my desired behavior with the complete removal of the debounce. For example, just manually setting the debounce time to 0 still gives undesired behavior, because of the asynchronous nature of `setTimeout`.

Would you be open to an option that completely ops out of the debouncing? Happy to implement such a thing if it would be helpful.

katspaugh:
I was asked to add the debouncing because otherwise it’s choppy with large files.

Since the interaction event is fired outside of the timeout, and thus reports the wrong time, maybe we can adjust that instead. I.e. pass a calculated new time to the event.

PR #2878, opened 2023-06-11
Provide a new time when emitting `interaction` events

brollin:
We now provide a new time when emitting the `interaction` event. This provides consumers with realtime information about drag events. Before, consumers would need to wait for the debounce. Tested behavior locally.

Additional parameter to `interaction` event handlers, but should be non-breaking.

None

### Todos/Notes:
None

https://github.com/katspaugh/wavesurfer.js/issues/2844


katspaugh (approved):
Thank you! 👍

jgentes:
This was great, thanks!

Issue #2845, opened 2023-06-05
The gap between canvases is too narrow.


letavocado:
### Summary
When wavesurfer.js generating multiple canvases, the barGap in canvases appears too narrow. 

### Possible solution:
This can be fixed by setting the barGap to be equal to the barWidth by default.

### Screenshots
![image](https://github.com/katspaugh/wavesurfer.js/assets/31704887/1129afb0-19ba-4eb8-91dd-8b4256e842a0)
![image](https://github.com/katspaugh/wavesurfer.js/assets/31704887/18ed2b71-fcca-45e1-b8ef-329c50cbf186)

katspaugh:
This happens at the edges of canvases when there’s scroll. It’s a bug indeed.

katspaugh:
If you’d like to try your hand at fixing this, the relevant code is here:

https://github.com/katspaugh/wavesurfer.js/blob/beta/src/renderer.ts#L356

It needs to take barWidth/barGap into account when calculating the canvas width. It has to be divisible by barWidth + barGap essentially.

letavocado:
Sure, I can take care of it during the upcoming weekend.

Cheers!

PR #2880, opened 2023-06-11
fix: narrow gaps between canvases

letavocado:
Resolves [#2845](https://github.com/katspaugh/wavesurfer.js/issues/2845)

A bug persists when barWidth or barGap is a floating number.

Todos/Notes:
requires e2e tests

Cheers!

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -204,6 +204,13 @@ class Renderer extends EventEmitter<RendererEvents> {
     }
   }
 
+  private closestInteger(a: number, b: number): number {
+    const c1: number = a - (a % b)
+    const c2: number = a + b - (a % b)
+
+    return Number.isInteger(c1) ? c1 : c2
+  }
+
   // Convert array of color values to linear gradient
   private convertColorValues(color?: WaveSurferOptions['waveColor']): string | CanvasGradient {
     if (!Array.isArray(color)) return color || ''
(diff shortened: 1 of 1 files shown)


src/renderer.ts
```diff
@@ -351,7 +358,16 @@ class Renderer extends EventEmitter<RendererEvents> {
...
+    let viewportWidth = Math.min(Renderer.MAX_CANVAS_WIDTH, clientWidth)
+
+    if (options.barWidth || options.barGap) {
+      const barWidth = options.barWidth ?? 1
+      const barGap = options.barGap ?? barWidth / 2
+      viewportWidth = this.closestInteger(
+        Math.min(Renderer.MAX_CANVAS_WIDTH, clientWidth) + barWidth / barGap,
+        barWidth + barGap,
+      )
+    }
```
katspaugh, line 370: How about something like this?

```suggestion
    if (options.barWidth || options.barGap) {
        const pixelRatio = window.devicePixelRatio || 1
        const barWidth = options.barWidth ? options.barWidth * pixelRatio : 1
        const barGap = options.barGap ? options.barGap * pixelRatio : options.barWidth ? barWidth / 2 : 0
        const remainder = viewportWidth % (barWidth + barGap)
        // viewportWidth must fit a whole number of bars and gaps
        if (remainder > 0) {
          viewportWidth += (barWidth + barGap - remainder)
        }
      )
    }
```

letavocado:
Thanks, @katspaugh! Awesome! :tada:

PR #2881, opened 2023-06-11
Auto height and plugin imports

katspaugh:
Modifies the auto height behavior introduced in #2841.

Sometimes the container might have a height due to paddings or some content. In this case filling that height by default would look weird and might be unexpected. A default height at least looks predictable.

To avoid these situations, I'm restoring the height behavior back to how it was prior to #2841, but it's now possible to set `height: 'auto` to fill the container vertically.

Also, I've made it easier to import plugins by making the `dist` part optional in the path and updated examples accordingly.


katspaugh:
FYI @IdoBouskila

src/renderer.ts
```diff
@@ -97,7 +97,10 @@ class Renderer extends EventEmitter<RendererEvents> {
 
   private getHeight(): number {
     const defaultHeight = 128
-    return this.options.height ?? (this.parent.clientHeight || defaultHeight)
+    if (this.options.height == null) return defaultHeight
+    if (!isNaN(Number(this.options.height))) return Number(this.options.height)
```
IdoBouskila, line 101: Why did you chose to use `! isNaN` instead of `isInteger`?

IdoBouskila:
> FYI @IdoBouskila

Thanks you 💯

IdoBouskila:
> FYI @IdoBouskila

Thank you! 💯

PR #2892, opened 2023-06-15
Check for supported mimeType in RecordPlugin

### LINKED ISSUE #2836: Playback after recording not working on a Safari ios 16

zetorama:

This MR improves `RecordPlugin`:

- allows providing `mimeType` and `audioBitsPerSecond` params when creating a `MediaRecorder` instance;
- if `mimeType` was not provided, automatically detects currently supported type via `MediaRecorder.isTypeSupported()`;
- preserves recorded `mimeType` when loads audio for playback.

None

- `Player#.setSrc()` for its second optional parameter now  instead of `ArrayBuffer` expects `Blob`, which preserves original audio mimeType within itself.

### Todos/Notes:

None

Resolves #2836

### DIFF
--- a/src/fetcher.ts
+++ b/src/fetcher.ts
@@ -2,8 +2,13 @@ async function fetchArrayBuffer(url: string): Promise<ArrayBuffer> {
   return fetch(url).then((response) => response.arrayBuffer())
 }
 
+async function fetchBlob(url: string): Promise<Blob> {
+  return fetch(url).then((response) => response.blob())
+}
+
 const Fetcher = {
   fetchArrayBuffer,
+  fetchBlob,
 }
 
 export default Fetcher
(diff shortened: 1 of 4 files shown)


katspaugh (approved):
That's a very smart solution, thank you so much!

PR #2907, opened 2023-06-17
All elements CSS editable 

TwanLuttik:
Allow the rest of the elements to be editable apart from the `cursor` element.

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -168,9 +168,9 @@ class Renderer extends EventEmitter<RendererEvents> {
       </style>
 
       <div class="scroll" part="scroll">
-        <div class="wrapper">
-          <div class="canvases"></div>
-          <div class="progress"></div>
+        <div class="wrapper" part="wrapper">
+          <div class="canvases" part="canvases"></div>
+          <div class="progress" part="progress"></div>
           <div class="cursor" part="cursor"></div>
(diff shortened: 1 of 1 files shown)


src/renderer.ts
```diff
@@ -168,9 +168,9 @@ class Renderer extends EventEmitter<RendererEvents> {
       </style>
 
       <div class="scroll" part="scroll">
-        <div class="wrapper">
-          <div class="canvases"></div>
-          <div class="progress"></div>
+        <div class="wrapper" part="wrapper">
+          <div class="canvases" part="canvases"></div>
```
katspaugh, line 172: Thanks for the PR!
I wouldn't like to encourage folks to customize these as a lot of rendering logic depends on the wrapper dimensions. Do you have a use case in mind?

TwanLuttik: I guess one of the ways we can do this is passing css styling when we initialize the wavesurfer
 
```
WaveSurfer.create({ 
  styles: { 
    cursor: {...},
    wrapper: {...} 
  } 
});
``` 

or initialize with the instance after creating it and use the reference like this 

```
instance.styles({ wrapper: {...} })
```

<br/>

besides this concept, I wonder how the performance difference between this method and what we have in place now.

PR #2911, opened 2023-06-18
Example: AmE vowels

katspaugh:
An example of generating mini-spectrograms for short sound clips.

Screenshot 2023-06-18 at 13 27 38


gartenfeld (commented):
🚀 Amazing!

gartenfeld (commented):
It sounds like 'er' has some artefacts -- drop of energy in the middle

Issue #2877, opened 2023-06-11
Beta: Typescript types aren't being resolved properly for imports


jgentes:
I thought there was a fix for this one, but it seems the types aren't resolving properly:
![image](https://github.com/katspaugh/wavesurfer.js/assets/1149470/592c1de5-a239-4297-a771-5d5936107062)
It's giving me:
```
Could not find a declaration file for module 'wavesurfer.js'. 'd:/mixpoint/node_modules/wavesurfer.js/dist/wavesurfer.min.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/wavesurfer.js` if it exists or add a new declaration (.d.ts) file containing `declare module 'wavesurfer.js';`ts(7016)
```
This also happens with the normal WaveSurfer import:
![image](https://github.com/katspaugh/wavesurfer.js/assets/1149470/5eedbd4c-3847-4b64-a4a0-adb2518ab881)

Thanks!

jgentes:
Here's my tsconfig.json for context:
```js
{
  "compilerOptions": {
    "lib": ["DOM", "DOM.Iterable", "ES2020"],
    "isolatedModules": true,
    "jsx": "react-jsx",
    "resolveJsonModule": true,
    "module": "NodeNext",
    "moduleResolution": "node16",
    "baseUrl": ".",
    "noEmit": true,
    "paths": {
      "~/*": ["./app/*"]
    },
    "allowJs": true,
    "forceConsistentCasingInFileNames": true,
    "strict": true,
    "target": "ESNext",
    "noImplicitAny": true,
    "strictNullChecks": true,
    "esModuleInterop": true
  },
  "include": ["**/*.ts", "**/*.tsx", "remix.env.d.ts", "./global.d.ts"]
}
```

katspaugh:
Thanks! I’ll try to reproduce it with your config. In the meanwhile, a workaround would be to import it by a full path, i.e. `wavesurfer.js/dist/wavesurfer.js`.

katspaugh:
Although actually I see it complains about full paths too…

katspaugh:
I've just tested with beta 9 and it now resolves types correctly. Tested in a TypeScript/Next.js/Webpack project.

katspaugh:
I've added explicit type exports, please see if beta.10 works for you. Cheers!

jgentes:
Well the error changed :)
[...]

PR #2937, opened 2023-06-25
experimental ts api

benallfree:
An idea for #2877.

This works well with build systems like Vite because it accesses the TS directly instead of the webpack output.

Related: https://github.com/microsoft/TypeScript/issues/48084

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,10 @@
   "types": "./dist/wavesurfer.d.ts",
   "browser": "./dist/wavesurfer.js",
   "exports": {
+    "./ts": {
+      "import": "./ts.ts",
+      "types": "./ts.ts"
+    },
     ".": {
       "import": "./dist/wavesurfer.js",
       "types": "./dist/wavesurfer.d.ts",
--- a/src/plugins/envelope.ts
+++ b/src/plugins/envelope.ts
@@ -169,7 +169,7 @@ class Polyline extends EventEmitter<{
   }
 }
 
-class EnvelopePlugin extends BasePlugin<EnvelopePluginEvents, EnvelopePluginOptions> {
+export class EnvelopePlugin extends BasePlugin<EnvelopePluginEvents, EnvelopePluginOptions> {
   protected options: Options
   private polyline: Polyline | null = null
   private audioContext: AudioContext | null = null
--- a/src/plugins/index.ts
+++ b/src/plugins/index.ts
@@ -1,0 +1,7 @@
+export * from './envelope'
+export * from './minimap'
+export * from './record'
+export * from './regions'
+export * from './spectrogram-fft'
+export * from './spectrogram'
+export * from './timeline'
--- a/src/plugins/minimap.ts
+++ b/src/plugins/minimap.ts
@@ -21,7 +21,7 @@ export type MinimapPluginEvents = {
   interaction: []
 }
 
-class MinimapPlugin extends BasePlugin<MinimapPluginEvents, MinimapPluginOptions> {
+export class MinimapPlugin extends BasePlugin<MinimapPluginEvents, MinimapPluginOptions> {
   protected options: MinimapPluginOptions & typeof defaultOptions
   private minimapWrapper: HTMLElement
   private miniWavesurfer: WaveSurfer | null = null
--- a/src/plugins/record.ts
+++ b/src/plugins/record.ts
@@ -19,7 +19,7 @@ export type RecordPluginEvents = {
 const MIME_TYPES = ['audio/webm', 'audio/wav', 'audio/mpeg', 'audio/mp4', 'audio/mp3']
 const findSupportedMimeType = () => MIME_TYPES.find((mimeType) => MediaRecorder.isTypeSupported(mimeType))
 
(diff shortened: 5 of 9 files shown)


package.json
```diff
@@ -24,6 +24,10 @@
   "types": "./dist/wavesurfer.d.ts",
   "browser": "./dist/wavesurfer.js",
   "exports": {
+    "./ts": {
+      "import": "./ts.ts",
+      "types": "./ts.ts"
```
katspaugh, line 29: Did you mean `./src/ts.ts`?

benallfree: I put it in the root so it looks like `import { WaveSurfer, RegionsPlugin } from 'wavesurfer.js/ts'`

katspaugh (commented):
It seems you intend to solve two different problems here:

## 1. Exporting everything from a single module
Exporting everything, i.e. the core wavesurfer lib and all of its plugins, from a single module is undoubtedly convenient in some situations. As a user, your import looks neat, just a single line. As a package maintainer, it's easier to document how to import the package.

I like that it's a separate export, so folks can choose to either import it from this new everything-module or the old way, separately.

The only gotcha is that when you import "raw" ES6 modules in the browser, you gotta be careful NOT to use this everything import if you're cautious about your download size, as it will inherently import everything. It wouldn't matter in bundlers like vite/rollup/webpack as they usually do tree-shaking.

So maybe, documenting-wise, this actually makes things a little bit more complicated. Maybe we shouldn't mention the separate module export explicitly?

## 2. Exporting a TypeScript source module instead of a compiled JS one
This is an interesting idea, and makes total sense. I haven't personally tried importing a TS file from an NPM module before. AFAIK, you typically need to configure tsconfig and/or webpack to also compile node_module imports for you. I can see this becoming problematic if the TypeScript version of the app is different than the one wavesurfer relies on.

Also with this approach, only people using TypeScript will benefit from the everything import. What about ES users?

benallfree:
> Exporting everything from a single module

Yes, this seems to be the trend I see happening, particularly in the Typescript community. Exports like `Config` become `RegionsConfig`, typically by just naming them that way in the first place. The source files are a little more verbose, but basically every exported symbol is named such that there are no collisions when everything is exported into a flat namespace. I wouldn't say this is a requirement, but I see it pretty frequently in popular TS packages and as you say it does make for a nice dx.

As far as I know, browsers aren't capable of importing TS directly, correct? I infer your ESM use case is for browsers that can natively import ESM? I've never used that. I'm not sure how a browser would accomplish tree shaking without first reading everything :) Is this a real use case? Are people doing ESM imports from the browser? I would imagine it would look like

```html
<script>
import { WaveSurfer } from 'https://.../wavesurfer.js'
import { RegionsPlugin } from 'https://.../wavesurfer.js/plugins/regions'
</script>
```
Firebase is a good example of a project that completely overhauled its import strategy to be more composable, perhaps for this exact use case. Rather than a hierarchical import tree, it's flattened so the dev imports all the services at the top level and then passes them around. Unlike [micro-dash](https://cdn.jsdelivr.net/npm/@s-libs/micro-dash@16.0.0/fesm2022/micro-dash.mjs) which does eventually just cram everything into one ESM, Firebase succeeds in decoupling the services. https://firebase.google.com/docs/web/modular-upgrade. I wonder if there is any wisdom or inspiration to be found there? They basically did away with classes altogether and instead use factory functions with an elegant IoC pattern.

> Exporting a TypeScript source module instead of a compiled JS one

This is for sure a better dx for TS devs. Nothing is lost in translation, bundlers have full access to the TS source, and IDEs like VSCode seem to have an easier time tracking type definitions. Not to mention, jumping to the type definition goes straight to the TS source, which can be immensely helpful in understanding what is going on.

katspaugh:
Those are fair points.

I suggest we do the following.
1. Make the main module an everything export. I.e. `import WaveSurfer, { RegionsPlugin } from 'wavesurfer.js'`. And in compiled JS, not TS. I would put the index file into `src/index.ts`, not in the root path.
2. Keep the existing `dist` exports as a way to import individual modules separately.
3. Export also the `src` path, so that folks can import TS files. Then you would do something like `import WaveSurfer, { RegionsPlugin, type WaveSurferOptions, type RegionsPluginOptions } from 'wavesurfer.js/src'` in TypeScript.
4. Document all of that in the readme.
5. Update all the examples.

benallfree:
Sounds good. Would you like me to do that?

How would you feel about exporting both as default and as an exported name, as I've done in this PR? `import { WaveSurfer }` is slightly easier than `import WaveSurfer` from a dx standpoint because VSCode's auto-import will index the name `WaveSurfer`.

katspaugh:
Absolutely, go for it!
I think it's a good idea to export both a default and a named export. 👍

benallfree:
#2939 is the first in this PR stack. I'm closing this PR in favor of a stacked approach because I think some of the ideas in here are not as wise as I thought. For example, pointing to the TS source code might be a good option for tools like Vite, but there are some tools which will not transpile TS found in `node_modules`, so I don't think it should be the only/default approach. It might be a good option though.

PR #2986, opened 2023-07-06
timeline plugin: restores formatTimeCallback

TS-Stefan:
## Short description
Restores the timeline plugin formatTimeCallback function from v6.

## Implementation details
copied over from v6

## Checklist

### DIFF
--- a/src/plugins/timeline.ts
+++ b/src/plugins/timeline.ts
@@ -27,7 +27,17 @@ export type TimelinePluginOptions = {
 
 const defaultOptions = {
   height: 20,
-  formatTimeCallback: Function(),
+  formatTimeCallback: (seconds: number) => {
+    if (seconds / 60 > 1) {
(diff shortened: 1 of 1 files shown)


src/plugins/timeline.ts
```diff
@@ -37,6 +40,7 @@ export class TimelinePlugin extends BasePlugin<TimelinePluginEvents, TimelinePlu
 
   constructor(options?: TimelinePluginOptions) {
     super(options || {})
+    defaultOptions.formatTimeCallback = this.defaultFormatTime
```
katspaugh, line 43: This is a little bit odd. I would move the default function out of the class directly to where the defaultOptions are specified. Otherwise instances of the plugin essentially overwrite a global object, affecting all other instances. Practically no difference but architecturally odd.

katspaugh (approved):
Thank you, LGTM!

PR #2999, opened 2023-07-11
Fix: [Regions] resizing when start is 0

### LINKED ISSUE #2998: [Regions] Resizing a region where {start: 0} doesn't work
### Discussed in https://github.com/katspaugh/wavesurfer.js/discussions/2996

<sup>Originally posted by **logesh2496** July 11, 2023</sup>
If you change the first region from the regions [example](https://wavesurfer-js.org/examples/#regions.js) to below

wsRegions.addRegion({
    **start: 0,**
    end: 2,
    content: 'Immovable',
    color: randomColor(),
  })
  
  The resizing of that region doesn't work. its not moving.</div>

katspaugh:
## Short description
Resolves #2998

## Implementation details
Resizing had a `start > 0` condition that should be `>=`.

## How to test it
Can be tested with the first region in the regions example.

## Screenshots
Screenshot 2023-07-11 at 08 47 42

### DIFF
--- a/src/plugins/regions.ts
+++ b/src/plugins/regions.ts
@@ -57,7 +57,7 @@ export type RegionParams = {
   maxLength?: number
 }
 
-export class Region extends EventEmitter<RegionEvents> {
+class SingleRegion extends EventEmitter<RegionEvents> {
   public element: HTMLElement
   public id: string
   public start: number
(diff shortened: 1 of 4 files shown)


src/plugins/regions.ts
```diff
@@ -57,7 +57,7 @@ export type RegionParams = {
   maxLength?: number
 }
 
-class Region extends EventEmitter<RegionEvents> {
+export class Region extends EventEmitter<RegionEvents> {
```
katspaugh, line 60: Exported this so that it's visible in the generated docs.

PR #3019, opened 2023-07-14
Feat: [Record] decouple real-time waveform and recorded audio

### LINKED ISSUE #3017: Record plugin: decouple mic access and recording
* Add a function to start the mic independently from recording
* If the mic isn't started when recording starts, it should start the mic automatically
* Keep the mic stream open until explicitly closed

katspaugh:
## Short description
Resolves #3017

## Implementation details
This PR refactors and introduces breaking changes in the Record plugin API. It will be released in 7.1.0.

The Record plugin will now continuously monitor microphone audio and render a real-time waveform for it.
It can be stopped by calling the `stopMic()` method.

Events renamed:
* `startRecording` -> `record-start`
* `stopRecording` -> `record-end`

Recorded audio will *not* be rendered in the same wavesurfer instance anymore. Instead, the `record-end` event will return a blob that can be rendered in another wavesurfer instance.

## Screenshots
Screenshot 2023-07-14 at 10 38 07

### DIFF
--- a/examples/record.js
+++ b/examples/record.js
@@ -5,7 +5,7 @@ import RecordPlugin from 'https://unpkg.com/wavesurfer.js@7/dist/plugins/record.
 
 // Create an instance of WaveSurfer
 const wavesurfer = WaveSurfer.create({
-  container: '#waveform',
+  container: '#mic',
   waveColor: 'rgb(200, 0, 200)',
   progressColor: 'rgb(100, 0, 100)',
 })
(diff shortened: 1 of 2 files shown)


hotmeteor:
Any idea when this will get in? I need it to fix the Firefox issue with:
```
Connecting AudioNodes from AudioContexts with different sample-rate is currently not supported.
```

katspaugh:
@hotmeteor I'm releasing 7.1.0 today

PR #3090, opened 2023-08-08
Chore: simplify plugin import syntax

ffxsam:
## Short description

~~TypeScript was complaining about not being able to find the declarations, so I fixed the exports map and tested it. Now I get full TS + autocompletion for plugin constructors.~~

This might've been an issue with VS Code misbehaving, so this PR is now just a bit of polish rather than a fix. I added a shorter way to import plugins, e.g.: `import SomePlugin from 'wavesurfer.js/plugins/some-plugin'`

## Checklist
* [X] This PR is covered by e2e tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,11 @@
       "types": "./dist/wavesurfer.d.ts",
       "require": "./dist/wavesurfer.cjs"
     },
+    "./dist/plugins/*.js": {
+      "import": "./dist/plugins/*.esm.js",
+      "types": "./dist/plugins/*.d.ts",
+      "require": "./dist/plugins/*.cjs"
+    },
     "./plugins/*": {
       "import": "./dist/plugins/*.esm.js",
       "types": "./dist/plugins/*.d.ts",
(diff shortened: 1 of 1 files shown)


katspaugh:
Hey Sam, thanks for the PR!

Unfortunately, this change will most certainly break people's code.

In the readme and on the website, there are examples of how to correctly import plugins, namely:

```js
import Regions from 'wavesurfer.js/dist/plugins/regions.js'
```

As you can see, it should have a `.js` extension. With your changes, this type of import will likely result in a double extension (e.g. `regions.js.esm.js`), which won't work.

If you prefer to import regions w/o an extension and with a shorter path, which I fully agree is nicer, we should add an additional mapping to the package.json, not replace an existing, working, mapping.

ffxsam:
Ah, right! Maybe this is something we can phase out in a future major version, then. I just pushed an update.

I feel like this PR started from what I *thought* was a bug. I swear the type declarations were not working when I tried the `dist/plugins/*.js` format, but now I'm thinking VS Code was experiencing a hiccup.

I'll edit the PR to reflect that this is _not_ a fix.

PR #3098, opened 2023-08-11
Allow timeline label intervals to be specified in notch count as well as seconds

rewbs:
## Short description

If the timeline displays units other than seconds – such as beats, or video frame position – notch and label intervals are likely to be precise float when converted to seconds. Precision is especially important when zoomed in at a high minPxPerSec. 

Currently `primaryLabelInterval` and `secondaryLabelInterval` are specified in seconds, and the computation to assess whether a given notch position is a label rounds to 2 decimal places. This is presumably desirable in some cases but not in the aforementioned use case, where the loss of precision causes labels to be skipped altogether.

I'd therefore like to suggest an alternative way to specify when labels should occur, simply based on the number of notches. For example:
* `primaryLabelSpacing=10` means every 10th notch is a primary label.
* `secondaryLabelSpacing=5` means every 5th notch is a secondary label.

## Implementation details

Add `primaryLabelSpacing` and `secondaryLabelSpacing` as additional ways to specify the label interval. 

If these options are present alongside the original `primaryLabelInterval` and `secondaryLabelSpacing`, both are used to assess whether a given notch should be a label.

## Checklist

(I can add tests if you agree this is useful)

### DIFF
--- a/src/plugins/timeline.ts
+++ b/src/plugins/timeline.ts
@@ -15,10 +15,14 @@ export type TimelinePluginOptions = {
   duration?: number
   /** Interval between ticks in seconds */
   timeInterval?: number
-  /** Interval between numeric labels */
+  /** Interval between numeric labels in seconds */
   primaryLabelInterval?: number
-  /** Interval between secondary numeric labels */
+  /** Interval between secondary numeric labels in seconds */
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
Great addition, thank you!

PR #3031, opened 2023-07-20
Fix: [Regions] support overlapping regions for region-in / region-out events

japostoles:
## Short description
This PR enables overlapping regions to receive in/out events

## Implementation details
A region array tracks diffs going in and out, emitting appropriate events as they mutate

## How to test it
Overlap two regions and monitor for region-in / region-out events

## Checklist
* [ ] This PR is covered by e2e tests
* [X] It introduces no breaking API changes

PR #3100, opened 2023-08-11
Fix: [Regions] Overlpaeed Regions region-out event error

gyeongseokKang:
- When regions are overlapped, region-out event triggered strangely

## Short description
Previously, there was only one active Region.

However, if there are multiple regions, there can be multiple active regions.

I have improved the logic for this case.

## Implementation details

- change activeRegion to actionRegions(array)
- find all played regions
- trigger event by  activeRegion & played regions
- update activeRegions

## How to test it
Check the triggered event(region-out) after overlapping different regions

## Screenshots
```
//reg region
{
    "listeners": {
        "update-end": {},
        "play": {},
        "click": {},
        "dblclick": {},
        "remove": {}
    },
    "totalDuration": 103.944,
    "minLength": 0,
    "maxLength": null,
    "id": "red",
    "start": 26.9,
    "end": 40.6,
    "drag": true,
    "resize": false,
    "color": "rgba(242, 85, 75, 0.3)",
    "element": null
}

// green region
{
    "listeners": {
        "update-end": {},
        "play": {},
        "click": {},
        "dblclick": {},
        "remove": {}
    },
    "totalDuration": 103.944,
    "minLength": 0,
    "maxLength": null,
    "id": "green",
    "start": 29.6,
    "end": 43.3,
    "drag": false,
    "resize": false,
    "color": "rgba(0, 179, 130, 0.3)",
    "element": null
}
```
image

before
image
after
image

### DIFF
--- a/src/plugins/regions.ts
+++ b/src/plugins/regions.ts
@@ -322,7 +322,6 @@ class RegionsPlugin extends BasePlugin<RegionsPluginEvents, RegionsPluginOptions
         // Trigger region-in when activeRegions doesn't include a played regions
         playedRegions.forEach((region) => {
           if (!activeRegions.includes(region)) {
-            activeRegions = [...activeRegions, region]
             this.emit('region-in', region)
           }
         })
@@ -335,7 +334,7 @@ class RegionsPlugin extends BasePlugin<RegionsPluginEvents, RegionsPluginOptions
         })
 
         // Update activeRegions only played regions
-        activeRegions = playedRegions.filter((region) => playedRegions.includes(region))
+        activeRegions = playedRegions
       }),
     )
   }
--- a/src/plugins/regions.ts
+++ b/src/plugins/regions.ts
@@ -313,22 +313,29 @@ class RegionsPlugin extends BasePlugin<RegionsPluginEvents, RegionsPluginOptions
     }
     this.wavesurfer.getWrapper().appendChild(this.regionsContainer)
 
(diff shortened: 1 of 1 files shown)


src/plugins/regions.ts
```diff
@@ -313,22 +313,29 @@ class RegionsPlugin extends BasePlugin<RegionsPluginEvents, RegionsPluginOptions
...
-          activeRegion = playedRegion
-          this.emit('region-in', playedRegion)
-        }
+        // Detect when regions are being played
+        const playedRegions = this.regions.filter((region) => region.start <= currentTime && region.end >= currentTime)
+
+        // Trigger region-in when activeRegions doesn't include a played regions
+        playedRegions.forEach((region) => {
+          if (!activeRegions.includes(region)) {
+            activeRegions = [...activeRegions, region]
```
katspaugh, line 325: Here, too, you can just assign `activeRegions = playedRegions`.

katspaugh (changes requested):

katspaugh:
Thanks for the PR, @gyeongseokKang!

It looks cleaner than #3031 but I still have a couple suggestions.

src/plugins/regions.ts
gyeongseokKang: Oh, it's my mistake!!
that are same

katspaugh (approved):
Thanks!

PR #3110, opened 2023-08-13
Fix spectrogram if it is hidden when Wavesurfer is initialised

rewbs:
## Short description

My use case is that I want my parent element with all audio UI components, including Wavesurfer, to be hidden until an audio file is loaded. At that point it is unhidden. 

Currently, if Wavesurfer includes a spectrogram, this will fail with:
```
IndexSizeError: Failed to construct 'ImageData': The source width is zero or not a number.
```

The reason is that the spectrogram width is retrieved when Wavesurfer is initialised, so is 0 if Wavesurfer is hidden at that time.

In this PR, we delay the retrieval of the width value until the spectrogram is drawn.

## Implementation details

`Spectrogram.width` is assigned from `wavesurfer.getWrapper().offsetWidth` in `drawSpectrogram()` instead of `onInit()`.

## How to test it

See attached testcase.

## Screenshots

Included in PR as part of testcase.

## Checklist
* [X] This PR is covered by e2e tests


katspaugh:
Thank you!

PR #3124, opened 2023-08-17
added 'id' and 'content' to setOptions(), multiple id values

rioharper:
## Short description
Added the ability to change the value of both id and content both in the region class using setOptions(), and also update the values that show up in the DOM. I have also added the ability to assign multiple ids to one region.

Resolves #
The ability to dynamically update the id and region content in the DOM after the region is created. This allows for things like dynamic CSS styling based on event triggers updating region id, or edited content based on user input. Multiple id's allow for tiered regions. An implementation, for instance, is that a region with multiple id's incorporate several different css style rules (priorities can be also used to create tiered systems)

## Implementation details
For updating id's and content values, It's pretty staightforward, I added the possible values to setOptions() in regions.ts, and the necessary code to update both the region class values, but also the element attributes. One thing to mention is that content can be passed through as either a HTMLElement or a string. 

For multiple id's, one can either pass through a string (`id = "id1"`) or as a string array (`id= ["id1", "id2", "id3"]`). This was done with some basic modifications to the SingleRegion() Constructor and initElement.

## Checklist
- [ ] This PR is covered by e2e tests
- [x] It introduces no breaking API changes

## How to test it
```javascript 
//html for the sake of visualization of the changes
/*
  <html>
    <style>
      #waveforms ::part(active) {
        border-radius: 40px !important;
        background-color: rgba(194, 78, 53, 0.5) !important;
      }

      #waveforms ::part(test) {
        height: 50% !important;
        background-color: rgba(34, 159, 49, 0.5);
      }

    #waveforms ::part(inactive) {
    border-radius: 20px !important;
    }
*/

// Create an instance of WaveSurfer
    container: '#waveform',
    waveColor: 'violet',
    progressColor: 'purple'
});

// Load an audio file
wavesurfer.load('path/to/audio/file');

// Create two regions
wavesurfer.on('ready', function () {
    //region with both id and content specified, will update both values
    wavesurfer.addRegion({
        start: 1, 
        end: 3, 
        color: 'hsla(100, 100%, 30%, 0.1)',
        id: "inactive",
        content: "inactive"
    });

    //region with neither id nor content specified, will create both values
    wavesurfer.addRegion({
        start: 5, 
        end: 7, 
        color: 'hsla(200, 50%, 70%, 0.4)'
    });

    //region with multiple id's specified, will update id value to string and generate content value
    wavesurfer.addRegion({
        start: 11, 
        end: 13,
        color: 'hsla(200, 50%, 70%, 0.4)',
        id: ["active", "highlighted"]
    });
});

// Add event listeners for region-in and region-out
wavesurfer.on('region-in', function(region) {
    region.setOptions({ id: "active", content: "active"});
    console.log('Region in');
    console.log(region);
});

wavesurfer.on('region-out', function(region) {
    region.setOptions({ id: "inactive", content: "inactive"});
    console.log('Region out');
    console.log(region);
});

### CODE REVIEW

File: src/plugins/regions.ts [hunk 1]
```diff
@@ -61,7 +61,7 @@ export type RegionParams = {
 
 class SingleRegion extends EventEmitter<RegionEvents> {
   public element: HTMLElement
-  public id: string
+  public id: string | string[]
```
katspaugh, line 64: I would keep this as a string because you're always joining it into a string if it's an array, so it's never actually an array.

src/plugins/regions.ts
```diff
@@ -73,8 +73,10 @@ class SingleRegion extends EventEmitter<RegionEvents> {
 
   constructor(params: RegionParams, private totalDuration: number) {
     super()
-
     this.id = params.id || `region-${Math.random().toString(32).slice(2)}`
+    if (Array.isArray(this.id)) {
+      this.id = this.id.join(' ')
+    }
```
katspaugh, line 79: Subsequently, here I would assign it only as a string. I.e.
```suggestion
    if (typeof params.id === 'string) {
      this.id = params.id
    } else if (Array.isArray(params.id)) {
      this.id = params.id.join(' ')
    } else {
      this.id = `region-${Math.random().toString(32).slice(2)}`
    }
```

src/plugins/regions.ts
```diff
@@ -90,8 +92,9 @@ class SingleRegion extends EventEmitter<RegionEvents> {
   private initElement(content?: string | HTMLElement) {
     const element = document.createElement('div')
     const isMarker = this.start === this.end
-
-    element.setAttribute('part', `${isMarker ? 'marker' : 'region'} ${this.id}`)
+  
+    element.setAttribute('part', `${isMarker ? 'marker' : 'region'}`)
+    element.setAttribute('id', Array.isArray(this.id) ? this.id.join(' ') : this.id)
```
katspaugh, line 97: Adding an id won't be useful, as it's part of a Shadow DOM tree. Only the part attribute is usable for CSS styling. So please keep it as it was.

src/plugins/regions.ts
```diff
@@ -106,7 +109,7 @@ class SingleRegion extends EventEmitter<RegionEvents> {
       pointer-events: all;
     `,
     )
-
+
```
katspaugh, line 112: Please run `yarn prettier` next time you commit, it should take care of these spaces.

src/plugins/regions.ts
```diff
@@ -279,6 +282,23 @@ class SingleRegion extends EventEmitter<RegionEvents> {
       this.end = options.end ?? (isMarker ? this.start : this.end)
       this.renderPosition()
     }
+    if (options.id !== undefined && this.start !== undefined && this.end !== undefined) {
```
katspaugh, line 285: I would make a separate `if` for the id and not mix it with start/end.

src/plugins/regions.ts
```diff
@@ -279,6 +282,23 @@ class SingleRegion extends EventEmitter<RegionEvents> {
...
+      const isMarker = this.start === this.end
+      if (Array.isArray(this.id)) {
+        this.element.setAttribute('part', `${isMarker ? 'marker' : 'region'} ${this.id.join(' ')}`)
+      } else {
+        this.element.setAttribute('part', `${isMarker ? 'marker' : 'region'} ${this.id}`)
+      }
+    }
+    if (options.content) {
+      if (typeof options.content === 'string') {
+        this.element.innerHTML = options.content
```
katspaugh, line 296: Not sure about `innerHTML` tbh. On one hand, it's convenient to pass an HTML string, on the other hand it's a security hole as you can pass any script like this creating an XSS vulnerability.

I would stick to either passing an HTML element or a pure text string using `content.textContent = options.content`.

katspaugh (changes requested):
Thanks for the PR!
A few suggestions.

src/plugins/regions.ts
rioharper: This is needed to parse an input array, so this will have to be kept in unless I'm missing something (which I very well might!)

src/plugins/regions.ts
rioharper: Im just going to do string, good point on the security flaws! This is my first js/ts project, so I may do some stupid stuff (I normally do python)

src/plugins/regions.ts
rioharper: Update: I just committed a change that makes region.id to be an string array, rather than a string with multiple id's separated by a space. I thought it would be more intuitive on the front end than splitting a string. So this should make these changes necessary

src/plugins/regions.ts
katspaugh: Please don't resolve my comments, I will do so when they are addressed.

src/plugins/regions.ts
katspaugh: IMHO it's better to stick to one type of value, especially since it's called `id` and not `ids`.
I would prefer to avoid all the `if`s down the line. I suggest that `id` property stays a string.

src/plugins/regions.ts
```diff
@@ -279,6 +291,22 @@ class SingleRegion extends EventEmitter<RegionEvents> {
...
+    if (this.start !== undefined && this.end !== undefined) {
+      if (options.id !== undefined) {
+        this.id = options.id
+        const isMarker = this.start === this.end
+        if (Array.isArray(this.id)) {
+          this.element.setAttribute('part', `${isMarker ? 'marker' : 'region'} ${this.id.join(' ')}`)
+        } else {
+          this.element.setAttribute('part', `${isMarker ? 'marker' : 'region'} ${this.id}`)
+        }
+      }
```
katspaugh, line 303: Please move this out of the parent `if`. Start/end props have nothing to do with the id, do they?

katspaugh:
Hey @rioharper, thanks for the PR!
I've implemented some of your changes in another PR (linked above), plus some extra code that was necessary for this to work.

PR #3128, opened 2023-08-20
Spectrogram plugin: make the container parameter optional

rewbs:
## Short description
Currently, if you try to create a spectrogram plugin without specifying a container explicitly in the plugin options, the typescript compiler will complain. I'm pretty sure this is a mistake in the type definition, because the implementation seems to cope fine if the container is not specified, and the example does not specify a container. Assuming that's right, this change corrects the type def.

## Implementation details
I just made `SpectrogramPluginOptions.container` optional, and added a test for the case where it _is_ specified (as the case where it wasn't is already covered by existing tests, which did not run into the issue because they are js rather than ts).

## How to test it
In Typescript code, try to create a Spectrogram plugin that does not specify a container explicitly. 

## Screenshots
n/a

## Checklist


katspaugh (approved):
Nice, thank you!
Could you run `yarn prettier` and commit again? Cheers

rewbs:
> Nice, thank you! Could you run `yarn prettier` and commit again? Cheers

Done, sorry about that. I also moved the expected output screenshots for my spectrogram tests to where I think they are expected to be when running `yarn test`. I'm a little confused because:
- It seems like `yarn cypress` and `yarn test` create/check screenshots in different locations
- As far as I can see, the other tests don't have snapshots under version control, so unless I'm missing something, the test run will just take snapshots without comparing?
- I've also noticed a little non-determinism with the spectrogram where sometimes there's a few pixels differences between the expected and new spectrogram, resulting in flaky tests.

As a result of the above doubts, this PR probably isn't ready for merge. Let me know if you have any thoughts on the above, else I'll keep digging. :)

Thanks!

katspaugh:
Oops. Good catch! looks like I accidentally deleted the snapshots in #3105. I'll restore them in a separate PR.

And yes, I need to figure out why `yarn cypress` and `yarn test` create them in different folders.

katspaugh:
Hey @rewbs, I've pushed your change here: https://github.com/katspaugh/wavesurfer.js/commit/994a6af3d108b2b8effdc4eebd61c6c11947e1bf
Without the test because snapshotting the entire body seems to be giving inconsistent results.

PR #3171, opened 2023-09-08
Regions plugin: Only add resize handles if resize prop is true

yyao84:
## Short description
Currently, adding a new region also creates left and right resize handles regardless of the resize param being set. For small regions, the resize handles makes it so you can't see the actual background color.

## Implementation details
Add the resize param to the resize handle creation conditional.

## How to test it
```
regionsPlugin.addRegion({
...,
resize: false
});
```
Added an e2e test as well

## Screenshots
Before: Before
After: After

## Checklist
* [✅] This PR is covered by e2e tests
* [✅] It introduces no breaking API changes

### DIFF
--- a/cypress/e2e/regions.cy.js
+++ b/cypress/e2e/regions.cy.js
@@ -308,4 +308,23 @@ describe('WaveSurfer Regions plugin tests', () => {
       win.wavesurfer.destroy()
     })
   })
+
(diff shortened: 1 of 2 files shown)


katspaugh (approved):
Thanks!

larsrgit:
The current implementation only checks during creation of the region. The resize handles should be added when the resize prop is set to true with changeOptions

PR #3192, opened 2023-09-14
Update player.ts

kapishv86:
## Short description
Resolves #

## Checklist


katspaugh:
I don't know what error you're getting in Angular, but the import is correct as it is. `GeneralEventTypes` are used only as a type.

PR #3208, opened 2023-09-23
Feat: update the container via setOptions

anton7r:
## Short description
Allows the user of the library to remount the wavesurfer instance to a new container.

Resolves #3197

## Implementation details
This feature allows changing the container via setOptions and adds a new event type 'containerRemount' which is only emitted when the container is remounted i.e. updated via setOptions.

If only the container is updated, it wont try to rerender the waveform.

In the change we introduced a new file `equals.ts` which has common utility functions for comparing 
Perhaps the `equals.ts` needs it's own unit testing or to be changed to `lodash`'s utility functions but from my experience it has some quirkiness in some scenarios.

## How to test it
As usual you can test the via the `yarn test` command.

You can also test the feature via the new example `remounting-container.js`.

## Checklist

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -31,15 +31,7 @@ class Renderer extends EventEmitter<RendererEvents> {
 
     this.options = options
 
-    let parent
-    if (typeof options.container === 'string') {
-      parent = document.querySelector(options.container) as HTMLElement | null
-    } else if (options.container instanceof HTMLElement) {
-      parent = options.container
-    }
-    if (!parent) {
-      throw new Error('Container not found')
-    }
+    const parent = this.parentFromOptionsContainer(options.container)
     this.parent = parent
 
     const [div, shadow] = this.initHtml()
@@ -56,6 +48,21 @@ class Renderer extends EventEmitter<RendererEvents> {
     }
 
     this.initEvents()
+  }
+
+  private parentFromOptionsContainer(container: WaveSurferOptions['container']) {
+    let parent
+    if (typeof container === 'string') {
+      parent = document.querySelector(container) satisfies HTMLElement | null
+    } else if (container instanceof HTMLElement) {
+      parent = container
+    }
+
+    if (!parent) {
+      throw new Error('Container not found')
+    }
+
+    return parent
   }
 
   private initEvents() {
@@ -204,8 +211,17 @@ class Renderer extends EventEmitter<RendererEvents> {
     return [div, shadow]
   }
 
+  /** Wavesurfer itself calls this method. Do not call it manually. */
   setOptions(options: WaveSurferOptions) {
+    if (this.options.container !== options.container) {
+      const newParent = this.parentFromOptionsContainer(options.container)
+      newParent.appendChild(this.container)
+
+      this.parent = newParent
+    }
+
     this.options = options
+
     // Re-render the waveform
     this.reRender()
   }
--- a/cypress/e2e/index.html
+++ b/cypress/e2e/index.html
@@ -7,6 +7,7 @@
   </head>
   <body>
     <div id="waveform" style="width: 600px"></div>
+    <div id="otherWaveform" style="width: 600px"></div>
 
     <script type="module">
       import WaveSurfer from '../../dist/wavesurfer.js'
--- a/cypress/e2e/options.cy.js
+++ b/cypress/e2e/options.cy.js
@@ -1,4 +1,5 @@
 const id = '#waveform'
+const otherId = `#otherWaveform`
 
 describe('WaveSurfer options tests', () => {
   beforeEach(() => {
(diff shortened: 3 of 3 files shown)


index.html
```diff
@@ -162,6 +162,7 @@ <h1>wavesurfer.js examples</h1>
           <li><a href="#multitrack.js">Multi-track</a></li>
           <li><a href="#vowels.js">Vowels</a></li>
           <li><a href="#fm-synth.js">FM synth</a></li>
+          <li><a href="#remounting-container.js">Remounting container</a></li>
```
katspaugh, line 165: This link might be too long for the sidebar. I’d prefer not to add a whole example for this, since it’s quite an edge case.

src/equals.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+export function objectEquals<T extends UnknownProperties>(
+  a: T,
+  b: T,
+  ignoredFields?: (keyof T)[],
+  isPartial = false,
+): boolean {
+  return internalObjectEquals(a, b, '', ignoredFields, isPartial)
+}
```
katspaugh, line 73: Strongly opposed to adding another file to the source tree.

src/renderer.ts
```diff
@@ -204,10 +213,24 @@ class Renderer extends EventEmitter<RendererEvents> {
     return [div, shadow]
   }
 
+  /** Wavesurfer itself calls this method. Do not call it manually. */
   setOptions(options: WaveSurferOptions) {
+    if (this.options.container !== options.container) {
+      const newParent = this.parentFromOptionsContainer(options.container)
+      this.parent.removeChild(this.container)
```
katspaugh, line 220: You don’t need to remove it before appending. AppendChild will move the element.

src/renderer.ts
```diff
@@ -204,10 +213,24 @@ class Renderer extends EventEmitter<RendererEvents> {
...
+      newParent.appendChild(this.container)
+
+      this.parent = newParent
+      this.emit('containerRemount')
```
katspaugh, line 224: The name of the event isn’t consistent with other events. None of them are in camelCase.

src/renderer.ts
```diff
@@ -204,10 +213,24 @@ class Renderer extends EventEmitter<RendererEvents> {
...
+    }
+
+    const renderOptionsChanged = !objectEquals(this.options, options, ['container'], true)
```
katspaugh, line 227: This could be simplified to something like this:
```js
if (!options.container || Object.keys(options).length > 1) {
  // re-render
}
```

src/wavesurfer.ts
```diff
@@ -120,6 +120,8 @@ export type WaveSurferEvents = {
   zoom: [minPxPerSec: number]
   /** Just before the waveform is destroyed so you can clean up your events */
   destroy: []
+  /** When the `container` is updated and wavesurfer remounts to it */
+  containerRemount: []
```
katspaugh, line 124: A new event seems like an overkill for this. The setOptions operation is synchronous so the caller can just react to it immediately after the call.

katspaugh (changes requested):
Hey, thanks for the PR!
I’ve left a few comments mostly related to minimizing the necessary changes.

src/renderer.ts
anton7r: Done

src/renderer.ts
anton7r: Removed the event.

index.html
anton7r: I mostly used it for testing purposes, removing it.

src/wavesurfer.ts
anton7r: Removed the event.

src/renderer.ts
anton7r: Originally wanted to add the `objectEquals` method so that it could be later used for eliminating the need for rerendering when options not related to rendering change such as `dragToSeek`, `autoplay`, `interact` (as far as i know those wont affect the rendering of the waveform themselves but i could be mistaken). If you don't plan on adding those then i'll just remove the utility functions.

Just now realised that there might be an issue with when the `fillParent` option is set to true and the `container` is updated to an element which dimensions does not match the original container's dimensions. This is not problematic for my use case right now but eventually it is very likely that atleast someone else finds it problematic.

src/equals.ts
anton7r: Removed the file

katspaugh (approved):
Looks good, thank you!

src/renderer.ts
katspaugh: True, as the only required param, I didn't consider container as a settable option, so some places might break.

katspaugh:
The `should remount the container when set via setOptions` test is failing.

src/renderer.ts
```diff
@@ -204,10 +211,21 @@ class Renderer extends EventEmitter<RendererEvents> {
...
+
+      this.parent = newParent
+    }
+
     this.options = options
-    // Re-render the waveform
-    this.reRender()
+
+    // Re-render the waveform if render related options have been changed
+    if (!options.container || Object.keys(options).length > 1) {
```
anton7r, line 226: The problem was caused by `Object.keys(options).length` returning 17 instead of 1 as we earlier expected. 

In the `wavesurfer.ts` file the `setOptions` method does an `Object.assing(...)` before it passes the value to the renderer.
Do i just add the equals.ts back in and if so where would you like the logic to reside in?

katspaugh: Good point. Then I would just remove this if and let it re-render every time. It would be actially good because then it will also have a chance to adjust the width to that of the new container.

I understand your particular use case was aiming to avoid re-renderings, and that was the whole point. I would suggest for those use cases to move (appendChild) the entire outer container and not the wavesurfer container.

anton7r: Tests now pass. Im continuing with the way you are suggesting therefore i don't really need this feature anymore.

It is up to you if you want to merge it or not.

katspaugh (approved):
LGTM, thank you!

PR #3223, opened 2023-09-30
Feat: add public method to set media element

jthomas-81:
## Short description
Allow user to set `media` after creation of Wavesurfer instance.

Resolves #3190

## Implementation details
Currently, we can specify a `media` element only during creation of Wavesurfer instance. In some advanced use cases where there are multiple players on a screen and we want to sync play/pause/seek between two players having the ability to share the `media` element during runtime would be advantageous, especially if one player is always persistent and survives screen navigation.

To enable this feature we added a public `setMediaElement` method to `Player` class.

## How to test it
As usual you can test with the `yarn cypress` command.

## Checklist

### DIFF
--- a/cypress/e2e/basic.cy.js
+++ b/cypress/e2e/basic.cy.js
@@ -176,4 +176,13 @@ describe('WaveSurfer basic tests', () => {
       win.wavesurfer.destroy()
     })
   })
+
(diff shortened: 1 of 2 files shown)


katspaugh (approved):
Thank you!

PR #3234, opened 2023-10-05
Feat: Add channelIdx parameter to have the ability to create a region just for a specific channel

### LINKED ISSUE #3220: Regions with split channels
Restore the `channelIdx` property in `RegionParams`

### Discussed in https://github.com/katspaugh/wavesurfer.js/discussions/3219

<sup>Originally posted by **jenga5** September 29, 2023</sup>
What happened to the `channelIdx` property in `RegionParams`? Now when you create a region, it is spread across all channels if you're using `splitChannels`</div>

bdriguesdev:
## Short description
Resolves #3220 

## Implementation details

Add channelIdx parameter to have the ability to create a region just for a specific channel

image

## How to test it

You can use this code to test it:

<details>
  <summary>code</summary>

```js
import WaveSurfer from 'https://unpkg.com/wavesurfer.js@7/dist/wavesurfer.esm.js'
import RegionsPlugin from 'https://unpkg.com/wavesurfer.js@7/dist/plugins/regions.esm.js'

  container: document.body,
  url: '/examples/audio/stereo.mp3',
  splitChannels: true,
})

// Initialize the Regions plugin
const wsRegions = wavesurfer.registerPlugin(RegionsPlugin.create())

wavesurfer.on('decode', () => {
  wsRegions.addRegion({
    start: 0,
    end: 8,
    content: 'Resize me',
    color: 'red',
    drag: false,
    resize: true,
    channelIdx: 1,
  })
  wsRegions.addRegion({
    start: 9,
    end: 39,
    content: 'Resize me',
    color: 'blue',
    drag: true,
    resize: true,
    channelIdx: 0,
  })
  wsRegions.addRegion({
    start: 100,
    end: 110,
    color: 'green',
    drag: true,
    resize: true,
  })
})

wavesurfer.on('interaction', () => {
  wavesurfer.play()
})
```

</details>

## Checklist


src/plugins/regions.ts
```diff
@@ -57,6 +58,8 @@ export type RegionParams = {
   minLength?: number
   /** Max length when resizing (in seconds) */
   maxLength?: number
+  /** The id of the channel */
+  channelId?: number
```
katspaugh, line 62: In v6, this param was called `channelIdx` which was technically more correct because it's a [channel's index](https://developer.mozilla.org/en-US/docs/Web/API/AudioBuffer/getChannelData), not an id.
Should we call it `channelIndex` maybe?

src/plugins/regions.ts
```diff
@@ -70,8 +73,10 @@ class SingleRegion extends EventEmitter<RegionEvents> {
   public content?: HTMLElement
   public minLength = 0
   public maxLength = Infinity
+  public channelId: number
+  public wavesurfer: WaveSurfer
 
-  constructor(params: RegionParams, private totalDuration: number) {
+  constructor(params: RegionParams, private totalDuration: number, wavesurfer: WaveSurfer) {
```
katspaugh, line 79: I intentionally didn't pass the wavesurfer instance to the Region class to limit its responsibilities.
The only place where you're using it is to get the numberOfChannels, so I'd suggest to pass that instead.

katspaugh (commented):
Hey, thanks for the PR! Looks good overall, just left a couple comments. Cheers!

src/plugins/regions.ts
bdriguesdev: I agree with you, for some reason I didn't realize it was an index instead of an id 🙈

src/plugins/regions.ts
bdriguesdev: I see it, thanks for explaining! I just pushed this change

katspaugh:
Thanks for addressing my comments!
Looks like the Regions tests are failing.

katspaugh (approved):
Thank you!

PR #3246, opened 2023-10-11
Show the correct way to import with type definitions enabled

AmitJoki:
## Short description
I tried the documentation in the existing README and that was not picking up the types and VSCode complained:

```
Cannot find module 'wavesurfer.js/plugins/record.js' or its corresponding type declarations
```

I then chanced upon https://github.com/katspaugh/wavesurfer.js/issues/2877#issuecomment-1606044454 and tried that and everything worked correctly, with types picked up. This PR updates the documentation.

## Checklist


katspaugh (approved):
Thanks!

PR #3249, opened 2023-10-12
feat: add pause/resume functionality to RecordPlugin

AmitJoki:
## Fixes
https://github.com/katspaugh/wavesurfer.js/discussions/3248

## Short description
Adds pause/resume functionality to RecordPlugin so users can pause and resume the recording instead of only being able to start and stop the recording.

## Implementation details
It uses [MediaRecorder.pause](https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder/pause) and [MediaRecorder.resume](https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder/resume)

## How to test it

The record.js example has been updated to include this functionality.

## Screenshots

image

## Checklist

### DIFF
--- a/examples/record.js
+++ b/examples/record.js
@@ -41,7 +41,17 @@ record.on('record-end', (blob) => {
     textContent: 'Download recording',
   })
 })
+const pauseButton = document.querySelector('#pause');
+pauseButton.onclick = () => {
+  if (record.isPaused()) {
+    record.resumeRecording()
+    pauseButton.textContent = 'Pause'
+    return
+  }
 
+  record.pauseRecording();
+  pauseButton.textContent = 'Resume'
+}
 {
   // Record button
   const recButton = document.querySelector('#record')
@@ -50,6 +60,7 @@ record.on('record-end', (blob) => {
     if (record.isRecording()) {
       record.stopRecording()
       recButton.textContent = 'Record'
+      pauseButton.style.display = 'none';
       return
     }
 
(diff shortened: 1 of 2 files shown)


AmitJoki:
@katspaugh based on the discussion https://github.com/katspaugh/wavesurfer.js/discussions/3247 added the PR. There's still one thing left to do however. When we pause currently, the waveform still picks the audio and generates the animations. I tried `wavesurfer.pause()` and `wavesurfer.empty()` but it did not work.

The canvas needs to stop animating when `pauseRecording()` and resume animating when `resumeRecording()` and it is probably trivial but I don't have experience working with canvases a lot so it'd be good if you took a quick look.

katspaugh (approved):
Awesome, thank you!

AmitJoki:
@katspaugh any thoughts on the following?

>There's still one thing left to do however. When we pause currently, the waveform still picks the audio and generates the animations. I tried wavesurfer.pause() and wavesurfer.empty() but it did not work.
>
>The canvas needs to stop animating when pauseRecording() and resume animating when resumeRecording() and it is probably trivial but I don't have experience working with canvases a lot so it'd be good if you took a quick look.

katspaugh:
The animation serves as a monitor for the mic, it will animate even if you're not recording at all once the mic is started. So that's by design.

So if you'd like to pause the animation, we'd have to add a separate method for it.

AmitJoki:
@katspaugh okay fair enough. Would you be publishing another release soon? I tried installing directly from 

`npm install https://github.com/katspaugh/wavesurfer.js` but that only works if the `package.json` has `prepare` command. If you'd be creating a release, I will use it directly otherwise I'd have to build the dist folder in my fork and add it to source control and use it from there. Do let me know, thanks!

katspaugh:
Sure, I've just pushed a 7.3.5 release.

Issue #3252, opened 2023-10-12
Error when repeatedly recording


katspaugh:
### Discussed in https://github.com/katspaugh/wavesurfer.js/discussions/3251

<div type='discussions-op-text'>

<sup>Originally posted by **nguyenhuyhoang241298** October 12, 2023</sup>
https://wavesurfer.xyz/examples/?record.js

When I record 2 or more times, I have warning like this 
![image](https://github.com/katspaugh/wavesurfer.js/assets/113531382/06c91c2f-f0e1-4900-b1c6-f99d4a15d08f)

</div>

katspaugh:
@AmitJoki looks like a regression after the pause feature (#3249).
Version 7.3.4 works fine.

AmitJoki:
The line that's causing the error is this https://github.com/katspaugh/wavesurfer.js/blob/main/src/plugins/record.ts#L97 which is not related to pause feature but rather the issue of the tabs showing recording is still on even when stopRecording has been called. I'll see if it can be solved in another way or that line could be removed.

PR #3253, opened 2023-10-12
Fix/multiple records

AmitJoki:
## Short description
Resolves https://github.com/katspaugh/wavesurfer.js/issues/3252

## Implementation details
Since the tracks were discarded, I guess the old references to it won't work. So I have set both `stream` and `mediaRecorder` to `null` so they can be set afresh.

## How to test it

Try doing multiple recordings. It should work. And when you stop recording the recording indicator on the tab should also correctly disappear.

## Checklist

### DIFF
--- a/src/plugins/record.ts
+++ b/src/plugins/record.ts
@@ -148,6 +148,8 @@ class RecordPlugin extends BasePlugin<RecordPluginEvents, RecordPluginOptions> {
     if (this.isRecording()) {
       this.mediaRecorder?.stop()
       this.stream?.getTracks().forEach((track) => track.stop())
+      this.stream = null
+      this.mediaRecorder = null
     }
   }
 
--- a/src/plugins/record.ts
+++ b/src/plugins/record.ts
@@ -96,6 +96,7 @@ class RecordPlugin extends BasePlugin<RecordPluginEvents, RecordPluginOptions> {
     if (!this.stream) return
     this.stream.getTracks().forEach((track) => track.stop())
     this.stream = null
+    this.mediaRecorder = null
   }
 
   /** Start recording audio from the microphone */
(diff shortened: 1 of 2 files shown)


src/plugins/record.ts
```diff
@@ -148,6 +148,8 @@ class RecordPlugin extends BasePlugin<RecordPluginEvents, RecordPluginOptions> {
     if (this.isRecording()) {
       this.mediaRecorder?.stop()
       this.stream?.getTracks().forEach((track) => track.stop())
+      this.stream = null
+      this.mediaRecorder = null
```
katspaugh, line 152: Actually, this whole thing is incorrect. When recording is stopped, it shouldn't stop the stream. The stream should be stopped only when the mic is stopped (the `stopMic` method is called).

The idea is that you can monitor the microphone across recordings. You can stop the mic explicitely from your app.

katspaugh: That's because `this.mediaRecorder` is cached and is never re-created. So it keeps the old stream.

katspaugh: We need to recreate the MediaRecorder object when the mic stream changes. https://github.com/katspaugh/wavesurfer.js/blob/main/src/plugins/record.ts#L105-L111

AmitJoki: We should then set `this.mediaRecorder = null` inside `stopMic` because then, in the `startRecording` function, `this.mediaRecorder` will be `null` and 
```
      new MediaRecorder(stream, {
        mimeType: this.options.mimeType || findSupportedMimeType(),
        audioBitsPerSecond: this.options.audioBitsPerSecond,
      })
```

gets executed thereby recreating the MediaRecorder object when the mic stream changes.

katspaugh: Yes, sounds right

AmitJoki: @katspaugh made the change.

katspaugh (approved):
Thank you, looks good!

PR #3256, opened 2023-10-13
Feat: Record Plugin - Choose Mic

tscritch:
## Short description
Resolves #3245
Open to feedback!

## Implementation details
- Adds options to `startMic` and `startRecording` to use a specific mic (via deviceId)
- Add function to get a list of available audio devices `getAvailableAudioDevices`

## How to test it
1. Open the Record example. 
1. If you haven't allowed mic permissions on the site before, for this example you will need to hit the record button first, allow permissions and refresh the page for the list to be updated.
1. You should be able to select a mic from the list.
1. Hit record and test the mic.

In Chrome, you can reset the "Allow Mic" permissions by going to Settings > Privacy and Security > Site settings > Microphone > Click trash icon on `wavesurfer.xyz` or `localhost:9090`.

## Checklist

### DIFF
--- a/examples/record.js
+++ b/examples/record.js
@@ -41,7 +41,7 @@ record.on('record-end', (blob) => {
     textContent: 'Download recording',
   })
 })
-const pauseButton = document.querySelector('#pause');
+const pauseButton = document.querySelector('#pause')
 pauseButton.onclick = () => {
   if (record.isPaused()) {
     record.resumeRecording()
@@ -49,8 +49,21 @@ pauseButton.onclick = () => {
     return
   }
 
-  record.pauseRecording();
+  record.pauseRecording()
   pauseButton.textContent = 'Resume'
+}
+
+const micSelect = document.querySelector('#mic-select')
+{
+  // Mic selection
+  record.getAvailableAudioDevices().then((devices) => {
+    devices.forEach((device) => {
+      const option = document.createElement('option')
+      option.value = device.deviceId
+      option.text = device.label || device.deviceId
+      micSelect.appendChild(option)
+    })
+  })
 }
 {
   // Record button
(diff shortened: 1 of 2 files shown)


src/plugins/record.ts
```diff
@@ -167,6 +174,25 @@ class RecordPlugin extends BasePlugin<RecordPluginEvents, RecordPluginOptions> {
...
+    let stream: MediaStream
+    try {
+      stream = await navigator.mediaDevices.getUserMedia({
+        audio: true,
+      })
+    } catch (err) {
+      throw new Error('Error getting audio devices: ' + (err as Error).message)
+    }
+    this.stream = stream
+    this.stopMic()
```
katspaugh, line 190: Looks good overall, thank you!

These lines are almost the same as in `startMic`, should we just call it here? Or maybe throw an error like `Please start the microphone first`.

tscritch: Possibly. I was concerned about how these lines from `startMic` would affect it? It doesn't need to render the stream when calling the function. What do you think?
```
    const onDestroy = this.renderMicStream(stream)
    this.subscriptions.push(this.once('destroy', onDestroy))
```

tscritch: I could also use this: https://developer.mozilla.org/en-US/docs/Web/API/Permissions/query
Then if no permission - call startMic

tscritch: Sorry for all the suggestions here. It would be nice if this was actually a static function and didn't touch the main class properties. It would keep similar code here but just use the local variable and use similar code to `stopMic` to ensure the steam gets deallocated.

tscritch: Landed on removing starting a stream and just adding better docs. 
The permissions api I shared isn't doesn't support a microphone query in Firefox 🤷🏻‍♂️

tscritch:
@katspaugh updated. Let me know your thoughts 🙏🏻

katspaugh (approved):
Thanks!

examples/record.js
```diff
@@ -41,17 +41,30 @@ record.on('record-end', (blob) => {
...
 
-  record.pauseRecording();
+  record.pauseRecording()
   pauseButton.textContent = 'Resume'
 }
+
+const micSelect = document.querySelector('#mic-select')
+{
+  // Mic selection
+  record.getAvailableAudioDevices().then((devices) => {
```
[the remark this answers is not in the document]
katspaugh: Good catch! Could you make a PR?

PR #3276, opened 2023-10-17
feat add ZoomPlugin

HoodyHuo:
## Short description
Add zoom plugin  ,which implements waveform zoom operation via mouse wheel.

## Implementation details
By calculating the scaled ratio and scroll position based on the audio time at the mouse's location, the current zoom level, and the width of the div container, we can then adjust the zoom level of wavesurfer and control scrolling

## Screenshots
Peek 2023-10-17 17-03

## Checklist

### DIFF
--- a/src/plugins/zoom.ts
+++ b/src/plugins/zoom.ts
@@ -55,6 +55,9 @@ class ZoomPlugin extends BasePlugin<ZoomPluginEvents, ZoomPluginOptions> {
     if (!this.wavesurfer?.options.minPxPerSec || !this.container) {
       return
     }
+    // prevent scrolling the sidebar while zooming
+    e.preventDefault()
+
     const duration = this.wavesurfer.getDuration()
     const oldMinPxPerSec = this.wavesurfer.options.minPxPerSec
     const x = e.clientX
--- a/src/plugins/zoom.ts
+++ b/src/plugins/zoom.ts
@@ -1,0 +1,93 @@
+/**
+ * Zoom plugin
+ *
+ * zoom player when mouse wheel
+ *
+ * @author HoodyHuo (https://github.com/HoodyHuo)
+ *
+ * @example
+ * // ... initialising wavesurfer with the plugin
+ * var wavesurfer = WaveSurfer.create({
+ *   // wavesurfer options ...
+ *   plugins: [
+ *     ZoomPlugin.create({
+ *       // plugin options ...
+ *     })
+ *   ]
+ * });
+ */
+
+import { BasePlugin, BasePluginEvents } from '../base-plugin.js'
+
+export type ZoomPluginOptions = {
+  scale?: number // every wheel step zoom  0.1 means 10%
+}
+const defaultOptions = {
+  scale: 0.2,
+}
+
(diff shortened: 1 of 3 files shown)


src/plugins/zoom.ts
```diff
@@ -0,0 +1,93 @@
+/**
+ * Zoom plugin
+ *
+ * zoom player when mouse wheel
```
katspaugh, line 4:
```suggestion
 * Zoom in our out on the waveform when scrolling the mouse wheel
```

src/plugins/zoom.ts
```diff
@@ -0,0 +1,93 @@
...
+ *       // plugin options ...
+ *     })
+ *   ]
+ * });
+ */
+
+import { BasePlugin, BasePluginEvents } from '../base-plugin.js'
+
+export type ZoomPluginOptions = {
+  scale?: number // every wheel step zoom  0.1 means 10%
```
katspaugh, line 23:
```suggestion
  scale?: number // the amount of zoom per wheel step, e.g. 0.1 means a 10% magnification per scroll
```

src/plugins/zoom.ts
```diff
@@ -0,0 +1,93 @@
...
+    if (this.wavesurfer?.options.container) {
+      const containerParam = this.wavesurfer?.options.container
+      if (typeof containerParam === 'string') {
+        this.container = document.querySelector(containerParam) as HTMLElement
+      } else if (containerParam instanceof HTMLElement) {
+        this.container = containerParam
+      } else {
+        this.container = this.wavesurfer?.getWrapper().parentElement as HTMLElement
+      }
+    }
```
katspaugh, line 59:
```suggestion
    this.container = this.wrapper.parentElement as HTMLElement
```

The container width is not always the width of the resulting waveform, so it's better to use the parent in all cases.

src/plugins/zoom.ts
```diff
@@ -0,0 +1,93 @@
...
+    this.wrapper.addEventListener('wheel', this.onWheel)
+  }
+
+  private onWheel = (e: WheelEvent) => {
+    if (!this.wavesurfer?.options.minPxPerSec || !this.container) {
+      return
+    }
+    const scrollEl = this.wrapper?.parentElement as HTMLElement
```
katspaugh, line 67: Here you could just use `this.container`

katspaugh (commented):
A nice plugin with a clear purpose! ❤️

katspaugh:
Yes, I'd like the comments I left addressed.

katspaugh:
The code looks good!
It would be nice to also add an example in the examples folder.

HoodyHuo:
> The code looks good! It would be nice to also add an example in the examples folder.

nice, I will add one tomorrow.

Serek16:
@katspaugh would be nice to have that example on the website.

katspaugh:
Feel free to add it

Serek16:
I meant on wavesurfer.xyz/examples/. And if possible, I haven't found a way to do it.

katspaugh:
They are sourced from https://github.com/katspaugh/wavesurfer.js/tree/main/examples

PR #3278, opened 2023-10-17
Fix: setMediaElement to properly subscribe/unsubscribe media events

jahangiranwari:
## Short description
When we change the underlying `media` element using `setMediaElement` the progress bar would not update as the track was playing. This was because of various media events subscriptions was missing. 
 
Resolves #3266

## Implementation details
To ensure that we are properly handling setting of a new media element `setMediaElement` has been tweaked as follows:

1. Unsubscribe from all event listeners for current `media` element
2. Attach the new `media` element
3. Subscribe to media (player) events for newly attached `media`

## How to test it

1. Create a Wavesurfer instance
2. Attached another `media` element using `setMediaElement`
3. Play

The progress bar should be updated as the track play progresses.

## Checklist

### DIFF
--- a/cypress/e2e/basic.cy.js
+++ b/cypress/e2e/basic.cy.js
@@ -177,13 +177,80 @@ describe('WaveSurfer basic tests', () => {
     })
   })
 
-  it('should set media without errors', () => {
-    cy.window().then((win) => {
-      const media = document.createElement('audio')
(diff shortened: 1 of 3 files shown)


katspaugh (approved):
Looks great, thank you sir!

PR #3281, opened 2023-10-18
Feat: restore the "backend: 'WebAudio'" option

### LINKED ISSUE #3272: Bring back Web Audio playback support
v7 switched away from Web Audio for the reasons described here: https://github.com/katspaugh/wavesurfer.js/discussions/2762#discussioncomment-5669347

However, we've seen a few use cases where Web Audio is preferrable over HTML audio:

* Precise control over current time and scheduling – impossible with HTML audio
* Problems with Variable Bit Rate audio – see #2890
* Simulataneous playback of several audio tracks on iOS Safari – slow with HTML audio, fast with Web Audio
* Volume control on iOS Safari – disabled for MediaElement
* [Certain audio encodings in Safari don't work](https://github.com/katspaugh/wavesurfer.js/discussions/3271)

I have a [prototype](https://github.com/katspaugh/wavesurfer-multitrack/blob/main/src/webaudio.ts) implemented that emulates the MediaElement API using a Web Audio BufferSourceNode. This could be used a shim for cases where MediaElement doesn't work.

katspaugh:
## Short description
Resolves #3272

I've restored the `backend: 'WebAudio'` option.

## Implementation details

There's a new class WebAudioPlayer that can be used as a drop-in replacement for a HTMLMediaElement. It emulates an audio element via a Web Audio buffer source node.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,16 +95,15 @@ See [this example](https://wavesurfer.xyz/examples/?styling.js) for play around 
 Most options, events, and methods are similar to those in previous versions.
 
 ### Notable differences
- * The `backend` option is removed – [HTML5 audio (or video) is the only playback mechanism](https://github.com/katspaugh/wavesurfer.js/discussions/2762#discussioncomment-5669347). However, you can still connect wavesurfer to Web Audio via `MediaElementSourceNode`. See this [example](https://wavesurfer.xyz/examples/?webaudio.js).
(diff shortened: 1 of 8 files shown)


katspaugh:
In what sense? Nothing is changed in this regard, Web Audio has always been used for decoding. This PR is only about the playback mechanism.

PR #3301, opened 2023-10-26
set public renderMicStream of Record plugin

cesxhin:
## Implementation details
You can now use a real-time MediaStream object to generate waves with Record plugin

## How to test it
Example get stream by tag audio

## Checklist


katspaugh (approved):
Thanks!

PR #3363, opened 2023-11-29
Ensure compatability between TimelinePlugin and RegionsPlugin

### LINKED ISSUE #3293: Can't add new regions by drag selection when timeline plugin is created with insertPosition 'beforebegin'
## Bug description
Title.

## Environment
 - Browser: Chrome

## Minimal code snippet
Use the [official regions example](https://wavesurfer.xyz/examples/?regions.js): 
Import the timeline plugin, and add it to wavesurfer, with insertPosition 'beforebegin'.

For copying:
import TimelinePlugin from 'https://unpkg.com/wavesurfer.js@7/dist/plugins/timeline.esm.js'
ws.registerPlugin(TimelinePlugin.create({insertPosition: 'beforebegin'}));

## Expected result
You can add regions when dragging on waveform.

## Obtained result
On drag nothing happens.

alexgdiaz98:
A few sensible changes to ensure safe interaction between the TimelinePlugin and RegionsPlugin in the DOM.

## Short description
Resolves #3293

## Implementation details
TimelinePlugin wrapper element has `pointer-events: none;` to ensure mouse interaction with other elements above or below the timeline. The RegionsPlugin targets the `canvases` div in `this.wavesurfer.getWrapper()` as the draggable target to avoid unexpectedly targeting an incorrect div.

## How to test it
```typescript
  container: '#wavesurfer-container',
})
const timeline = TimelinePlugin.create({
  container: wavesurfer.getWrapper(),
  insertPosition: 'beforebegin'
})
wavesurfer.registerPlugin(timeline)
const regions = RegionsPlugin.create()
wavesurfer.registerPlugin(regions)
regions.enableDragSelection()
```
Dragging anywhere on the canvas, including over the timeline, should create a region.

## Checklist

### DIFF
--- a/src/plugins/regions.ts
+++ b/src/plugins/regions.ts
@@ -402,6 +402,7 @@ class RegionsPlugin extends BasePlugin<RegionsPluginEvents, RegionsPluginOptions
 
   private initRegionsContainer(): HTMLElement {
     const div = document.createElement('div')
+    div.setAttribute('part', 'regions')
     div.setAttribute(
       'style',
       `
@@ -509,8 +510,8 @@ class RegionsPlugin extends BasePlugin<RegionsPluginEvents, RegionsPluginOptions
    * Returns a function to disable the drag selection.
    */
   public enableDragSelection(options: Omit<RegionParams, 'start' | 'end'>): () => void {
-    const wrapper = this.wavesurfer?.getWrapper()?.querySelector('div')
-    if (!wrapper) return () => undefined
+    const wrapper = this.wavesurfer?.getWrapper()?.querySelector('div.canvases')
+    if (!wrapper || !(wrapper instanceof HTMLElement)) return () => undefined
 
     const initialSize = 5
     let region: Region | null = null
--- a/src/plugins/timeline.ts
+++ b/src/plugins/timeline.ts
@@ -49,7 +49,7 @@ export type TimelinePluginEvents = BasePluginEvents & {
 }
 
 class TimelinePlugin extends BasePlugin<TimelinePluginEvents, TimelinePluginOptions> {
-  private timelineWrapper: HTMLElement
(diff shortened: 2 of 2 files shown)


src/plugins/timeline.ts
```diff
@@ -49,7 +49,7 @@ export type TimelinePluginEvents = BasePluginEvents & {
 }
 
 class TimelinePlugin extends BasePlugin<TimelinePluginEvents, TimelinePluginOptions> {
-  private timelineWrapper: HTMLElement
+  public timelineWrapper: HTMLElement
```
katspaugh, line 52: Every time we make something internal like this public, it increases the support surface area and makes upgrading to newer versions harder for users.

E.g. `this.wavesurfer.getWrapper` was never meant to be used by anything other than internally by plugins, and it says so in the doc comment.

So I would suggest reverting this.

src/plugins/regions.ts
```diff
@@ -402,6 +402,7 @@ class RegionsPlugin extends BasePlugin<RegionsPluginEvents, RegionsPluginOptions
 
   private initRegionsContainer(): HTMLElement {
     const div = document.createElement('div')
+    div.setAttribute('part', 'regions')
```
katspaugh, line 405: What use case do you have in mind for this?

src/plugins/timeline.ts
alexgdiaz98: Reverted 👍🏽

src/plugins/regions.ts
katspaugh: I wouldn’t add code for “just in case”. It’s all rendered in a shadow DOM for a reason. Other elements selectively have the part attribute for external styling.

katspaugh (approved):
Thanks!

PR #3392, opened 2023-12-10
Feat: [Timeline] Can now specify secondary label opacity

ffxsam:
## Short description
We wanted a way to have all time indicators & ticks presented with equal visual prominence.

## Implementation details
Added a new option, `secondaryLabelOpacity`, which defaults to 0.25.

## Checklist

Opening this in draft state so we can discuss. Open to any feedback/changes!

### DIFF
--- a/src/plugins/timeline.ts
+++ b/src/plugins/timeline.ts
@@ -27,6 +27,8 @@ export type TimelinePluginOptions = {
   style?: Partial<CSSStyleDeclaration> | string
   /** Turn the time into a suitable label for the time. */
   formatTimeCallback?: (seconds: number) => string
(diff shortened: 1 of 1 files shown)


katspaugh:
This addition makes sense to me! An alternative approach would be to override the opacity via CSS like this:
```css
::part(timeline-notch) {
  opacity: 1 !important;
}
```

katspaugh (approved):

ffxsam:
Wow, today I learned about `part`! I can't keep up. 😅

Now I'm on the fence about this PR. I'll leave it to your discretion. It obviously didn't take much time to add this option, so it's cool if we toss it!

PR #3396, opened 2023-12-11
Fix: [Zoom] Performance improvements (especially w/ trackpad zoom)

ffxsam:
## Short description
Zooming with a trackpad was causing huge CPU spikes due to constant redraws. This PR imposes rate limiting via a delta threshold before actually redrawing the waveform.

## Implementation details
Debounce & throttle were not appropriate here, as we don't necessarily want to throttle the events, we just want to ignore the "in-between" delta values and have a more notched zoom (a la Adobe Audition).

## How to test it
Test using any trackpad with smooth scrolling (i.e. lots of event firing when scrolling).

## Checklist
* [?] It introduces no breaking API changes

### DIFF
--- a/src/plugins/zoom.ts
+++ b/src/plugins/zoom.ts
@@ -33,13 +33,13 @@ export type ZoomPluginOptions = {
    * The amount the wheel or trackpad needs to be moved before zooming the waveform. Set this value to 0 to have totally
    * fluid zooming (this has a high CPU cost).
    *
-   * @default 10
+   * @default 5
    */
   deltaThreshold?: number
 }
 const defaultOptions = {
   scale: 0.5,
-  deltaThreshold: 10,
+  deltaThreshold: 5,
 }
 
 export type ZoomPluginEvents = BasePluginEvents
--- a/src/plugins/zoom.ts
+++ b/src/plugins/zoom.ts
@@ -86,10 +86,7 @@ class ZoomPlugin extends BasePlugin<ZoomPluginEvents, ZoomPluginOptions> {
       const width = this.container.clientWidth
       const scrollX = this.wavesurfer.getScroll()
       const pointerTime = (scrollX + x) / oldMinPxPerSec
-      const newMinPxPerSec = this.calculateNewZoom(
-        oldMinPxPerSec,
-        this.options.deltaThreshold > 0 ? this.accumulatedDelta : -e.deltaY,
-      )
+      const newMinPxPerSec = this.calculateNewZoom(oldMinPxPerSec, this.accumulatedDelta)
       const newLeftSec = (width / newMinPxPerSec) * (x / width)
 
       if (newMinPxPerSec * duration < width) {
--- a/src/plugins/zoom.ts
+++ b/src/plugins/zoom.ts
@@ -4,6 +4,8 @@
  * Zoom in or out on the waveform when scrolling the mouse wheel
  *
  * @author HoodyHuo (https://github.com/HoodyHuo)
+ * @author Chris Morbitzer (https://github.com/cmorbitzer)
+ * @author Sam Hulick (https://github.com/ffxsam)
  *
  * @example
  * // ... initialising wavesurfer with the plugin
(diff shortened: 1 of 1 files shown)


ffxsam:
Still a work-in-progress. Standard mouse wheeling is a bit wonky now (doesn't always zoom). Delta accumulation logic needs tweaks.

Also, unsure if this is a breaking API change. IMO, the default behavior should be the zoom detents (`deltaThreshold` of 10). Users can have smooth zooming by setting this to 0.

Open to any ideas, changes, etc. This is just a start!

ffxsam:
This might help explain why throttling wasn't the right approach:

In Audition, you can scroll as quickly as you want, but it's a stepped zoom rather than a fluid one. 

https://github.com/katspaugh/wavesurfer.js/assets/12532733/ab792ceb-72dd-4f5c-9cf5-72cf665ae870

src/plugins/zoom.ts
```diff
@@ -59,20 +75,33 @@ class ZoomPlugin extends BasePlugin<ZoomPluginEvents, ZoomPluginOptions> {
...
+      if (newMinPxPerSec * duration < width) {
+        this.wavesurfer.zoom(width / duration)
+        this.container.scrollLeft = 0
+      } else {
+        this.wavesurfer.zoom(newMinPxPerSec)
+        this.container.scrollLeft = (pointerTime - newLeftSec) * newMinPxPerSec
+      }
+
+      // Reset the accumulated delta
+      this.accumulatedDelta %= this.options.deltaThreshold
```
katspaugh, line 104: Here it can probably be just set to 0?

ffxsam: I'm gonna rework some of this. I'll mark the PR as ready for review once I think it's in a good spot.

ffxsam:
Ok, just setting it to 0 seems to have fixed ignored mouse wheel movements.

The thing is, all mice are different, and some of them may have smaller delta jumps with each wheel notch. Tested on a Logitech Lift and a Logitech MX Master 3S (with smooth wheel enabled). Both had similar deltas (see below), but that's likely because they're both Logitech. 😉 

CleanShot 2023-12-11 at 11 47 50

src/plugins/zoom.ts
```diff
@@ -59,20 +75,33 @@ class ZoomPlugin extends BasePlugin<ZoomPluginEvents, ZoomPluginOptions> {
...
+      const duration = this.wavesurfer.getDuration()
+      const oldMinPxPerSec = this.wavesurfer.options.minPxPerSec
+      const x = e.clientX
+      const width = this.container.clientWidth
+      const scrollX = this.wavesurfer.getScroll()
+      const pointerTime = (scrollX + x) / oldMinPxPerSec
+      const newMinPxPerSec = this.calculateNewZoom(
+        oldMinPxPerSec,
+        this.options.deltaThreshold > 0 ? this.accumulatedDelta : -e.deltaY,
+      )
```
katspaugh, line 92: I think this can be simplified to 
```suggestion
      const newMinPxPerSec = this.calculateNewZoom(oldMinPxPerSec, this.accumulatedDelta)
```

If the threshold is 0, `this.accumulatedDelta` would be just equal to the delta because it's reset every time.

katspaugh (approved):
Thanks! 👍

katspaugh:
So to clarify, you ideally want the waveform to re-render every time the mouse wheel clicks, but we cannot count clicks, only a (pixel?) delta.

Looking at [this MDN article](https://developer.mozilla.org/en-US/docs/Web/API/Element/wheel_event), it looks like we can only see what mode the mouse wheel is in (scrolling per page, per line or per pixel).

And of course touchpads and trackpads don't have notches at all...

So I think 10 is a sensible default, especially if your two mice are around that ballbark.

@cmorbitzer good with you?

ffxsam:
Right. Too many re-renders = bad news for those on lighter-weight computers. Step Zoom is just the safer route.

Even when I compare Adobe Audition vs. iZotope RX with trackpad zoom, RX has smooth zoom, and my CPU usage jumps to 150-200+. With Audition (step zoom), my computer doesn't even get its heart rate going. 😁

cmorbitzer:
LGTM

I don't have an opinion on performance, but 10 felt choppy to me as a user, where 5 felt seamless enough. Do we get much of the same resource benefit by setting the threshold to 5?

ffxsam:
A setting of 5 had more spikes for me (on a Mac Studio), but still better than totally smooth zoom. Maybe we default delta threshold to 5? I can test in a bit, and I'd like to update the docs a bit to explain this setting better.

ffxsam:
Yep, 5 is good. Not as good as 10, but a good middle ground. Just updated the code, and I think the description for this setting will do.

katspaugh:
Thanks!

bernardwiesner:
@ffxsam Even setting the `deltaThreshold` to 50, when I scroll quickly on a big waveform, page suddenly crashes. 

Could we not use a throttling instead? I mean user scrolls very quickly, but we only re-draw once after the scroll is over. It would not be as fluid but more stable.

ffxsam:
@bernardwiesner Hey Bernard! I'm honestly not sure. If we used throttle, that would mean that scroll events would have no visible effect until the user stopped scrolling, which is not ideal.

If you're experiencing any kind of bug or performance issue, please feel free to open a new issue.

Thanks!

PR #3422, opened 2023-12-18
Feat: [Region] Add region-removed event

jason-whitted:
## Short description
Add `region-removed` event to `RegionPluginEvents`, which emits when a Region's `remove` event is dispatched.

Resolves #

## Implementation details
This `region-removed` event fires after the Region has been removed and is not cancelable.

## Checklist


katspaugh (approved):
Nice, thank you!

PR #3471, opened 2024-01-07
Fix: don't use a blob URL for unknown MIME types

katspaugh:
## Short description
Resolves #3380 

As per [this comment](https://github.com/katspaugh/wavesurfer.js/discussions/3380#discussioncomment-8034069), if an audio file is fetched from a server that doesn't send the correct MIME type, the downloaded blob might fail to play in an audio element.

## Implementation details

I've added a check when setting an audio src to a blob URL: if the MIME type is not supported, it will fallback to the original audio URL.

### DIFF
--- a/src/player.ts
+++ b/src/player.ts
@@ -63,11 +63,15 @@ class Player<T extends GeneralEventTypes> extends EventEmitter<T> {
     }
   }
 
+  private canPlayType(type: string): boolean {
+    return this.media.canPlayType(type) !== ''
+  }
+
   protected setSrc(url: string, blob?: Blob) {
     const src = this.getSrc()
     if (src === url) return
     this.revokeSrc()
-    const newSrc = blob instanceof Blob ? URL.createObjectURL(blob) : url
+    const newSrc = blob instanceof Blob && this.canPlayType(blob.type) ? URL.createObjectURL(blob) : url
     this.media.src = newSrc
   }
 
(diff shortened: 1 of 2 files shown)


katspaugh:
@719media could you test this branch against the problematic server?

719media:
I have confirmed that the https://blob-type.wavesurfer-js.pages.dev/ page works for loading "bad mime-type" file examples as discussed in the linked issue. Vs current master where it continues to not work (no waveform/nothing plays)

katspaugh:
Awesome, thank you! 🙏

wfk007:
@katspaugh Is it possible to add a config for `canPlayType` check? It is tough to upgrade `wavesurfer` version in my system because of the server side.

katspaugh:
You mean to optionally skip this check?

wfk007:
> You mean to optionally skip this check?

Yes. we can choose whether to use this check.

katspaugh:
I would suggest an alternative to that: specify a MIME type in the options. E.g.

```
WaveSurfer.create({
  container: '#container',
  url: 'music.mp3',
  mimeType: 'audio/mpeg',
})
```

It would then use this MIME type when [fetching a blob](https://github.com/katspaugh/wavesurfer.js/blob/main/src/fetcher.ts#L50):

```
fetch('url-to-your-data')
  .then(response => response.blob()) // Get the blob from the response
  .then(blob => {
    if (options.mimeType) {
      const newType = options.mimeType
      const dataBlob = new Blob([blob], { type: newType }); // Override the MIME type
      return dataBlob;
    }
    return blob
  })
```

wfk007:
This seems to work for me, and this design is consistent with the record plugin(`mediaRecorder.mimeType`). let me have a try. Thanks for your great suggestion！

PR #3472, opened 2024-01-08
Feat: `record-progress` event and `getDuration` method

aantusahaa:
## Short description
Introduces a new event `record-progress` and a new method `getDuration`.
Resolves #3431 

### `record-progress` event
Internal implementation keeps track of the time with `Clock` class. The clock will emit `tick` with `requestAnimationFrame`. It is heavily inspired by the already existing `Timer` class. Only difference is it keeps track of the time along with ticking. 

The clock will start when the media recording has begun. It will pause/resume accordingly as well. The clock will not reset until the next `record-start` event fires. 

### `getDuration` method
Always returns the `currentTime` from the `Clock`. Since the clock won't reset until the next start event, the consumer will be able to get the duration even after the recording ends.

## How to test it
There is a progress timer added in the `record.js` example.

## Checklist


src/plugins/record.ts
```diff
@@ -46,6 +49,8 @@ class RecordPlugin extends BasePlugin<RecordPluginEvents, RecordPluginOptions> {
   private dataWindow: Float32Array | null = null
   private isWaveformPaused = false
   private originalOptions: { cursorWidth: number; interact: boolean } | undefined
+  private clock: Clock
+  protected subscriptions: Array<() => void> = []
```
katspaugh, line 53: The BasePlugin already has this defined, so you don't have to.

src/plugins/record.ts
```diff
@@ -248,6 +270,7 @@ class RecordPlugin extends BasePlugin<RecordPluginEvents, RecordPluginOptions> {
 
   /** Destroy the plugin */
   public destroy() {
+    this.subscriptions.forEach((unsubscribe) => unsubscribe())
```
katspaugh, line 273: This can be replaced by `super.destroy()` (the BasePlugin unsubscribes from all events).

src/clock.ts
```diff
@@ -0,0 +1,49 @@
...
+    }
+  }
+
+  start(): void {
+    this.startTime = performance.now()
+    this.currentTime = 0
+    this.isPaused = false
+
+    this.unsubscribe = this.on('tick', () => {
+      requestAnimationFrame(this.updateTime.bind(this))
```
katspaugh, line 27: We don't really have to update the time on each tick. You could just return `performance.now() - startTime` from `getCurrentTime`. On pause, you could save the intermediate duration into a variable. This is both much more accurate and efficient.

Check out thow time is tracked in this class: https://github.com/katspaugh/wavesurfer.js/blob/main/src/webaudio.ts#L148-L150

With this in mind, we could get away with just the Timer class + tracking the recording duration in the Record plugin itself, w/o creating a new Clock class.

katspaugh (changes requested):

katspaugh:
Thanks for the PR! 🙏 Made a couple comments.

aantusahaa:
Thanks for the pointer @katspaugh! Made some changes.

katspaugh (approved):
Nice, thank you, looks good to me!

PR #3480, opened 2024-01-10
Feat: add redrawcomplete event

brian-byassee:
## Short description
Currently the `redraw` event firers before `renderHead` and `renderTail` have completed. This adds a new event `redrawcomplete` that fires once both head and tail have completed rendering.

In longer files `redrawcomplete` can take awhile to fire so we left `redraw` and `redrawcomplete` as two separate events

Resolves #

## How to test it
1. Navigate to [zoom example](http://127.0.0.1:9090/#zoom-plugin.js)
2. Add: `['zoom', 'redraw', 'redrawcomplete'].forEach((e) => wavesurfer.on(e, () => console.log(Date.now(), e)))`
3. view console

## Screenshots
image

## Checklist

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -8,6 +8,7 @@ type RendererEvents = {
   drag: [relativeX: number]
   scroll: [relativeStart: number, relativeEnd: number]
   render: []
+  rendered: []
 }
 
 class Renderer extends EventEmitter<RendererEvents> {
@@ -454,7 +455,7 @@ class Renderer extends EventEmitter<RendererEvents> {
     }
   }
 
-  private renderChannel(channelData: Array<Float32Array | number[]>, options: WaveSurferOptions, width: number) {
+  private renderChannel(channelData: Array<Float32Array | number[]>, options: WaveSurferOptions, width: number, done: () => void) {
     // A container for canvases
     const canvasContainer = document.createElement('div')
     const height = this.getHeight(options.height)
@@ -501,6 +502,14 @@ class Renderer extends EventEmitter<RendererEvents> {
       )
     }
 
+    const status: { [k:string]: boolean } = { head: false, tail: end >= len }
+    const complete = (type: string) => {
+      status[type] = true
+      if (status.head && status.tail) {
+        done()
+      }
+    }
+
     // Draw the waveform in viewport chunks, each with a delay
     const headDelay = this.createDelay()
     const tailDelay = this.createDelay()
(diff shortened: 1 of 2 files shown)


brian-byassee:
Not certain how you feel about the new event names but happy to update them if you want.

src/renderer.ts
```diff
@@ -518,6 +529,8 @@ class Renderer extends EventEmitter<RendererEvents> {
         tailDelay(() => {
           renderTail(fromIndex + viewportLen, toIndex + viewportLen)
         })
+      } else {
+        complete('tail')
```
katspaugh, line 533: I'm thinking, maybe this could be refactored as a promise, smth like:

```js

    // Draw the waveform in viewport chunks, each with a delay
    return Promise.all([
      new Promise((resolve) => {
        const headDelay = this.createDelay()
        const renderHead = (fromIndex: number, toIndex: number) => {
          draw(fromIndex, toIndex)
          if (fromIndex > 0) {
            headDelay(() => {
              renderHead(fromIndex - viewportLen, toIndex - viewportLen)
            })
          } else {
            resolve(true)
          }
        }
        renderHead(start, end)
      }),

      new Promise((resolve) => {
        if (end >= len) {
          resolve(true)
          return
        }
        const tailDelay = this.createDelay()
        const renderTail = (fromIndex: number, toIndex: number) => {
          draw(fromIndex, toIndex)
          if (toIndex < len) {
            tailDelay(() => {
              renderTail(fromIndex + viewportLen, toIndex + viewportLen)
            })
          } else {
            resolve(true)
          }
        }
        renderTail(end, end + viewportLen)
      }),
    ])
```

src/wavesurfer.ts
```diff
@@ -99,8 +99,10 @@ export type WaveSurferEvents = {
   decode: [duration: number]
   /** When the audio is both decoded and can play */
   ready: [duration: number]
-  /** When a waveform is drawn */
+  /** When visible waveform is drawn */
   redraw: []
+  /** When all audio channel chunks of the waveform have drawn */
+  redrawcomplete: []
```
katspaugh, line 105: I personally like the event name, it's clear!

src/renderer.ts
jasonwhitted-socialive: @katspaugh, we thought about that, too; however, one downside to this is that createDelay could end up canceling the setTimeout, which would mean that `resolve` never gets called and ends up adding unresolved promises as memory leaks.

jason-whitted: @katspaugh We thought about that, too; however, one downside to this is that `createDelay` could end up canceling the setTimeout, which would mean that resolve never gets called and ends up adding unresolved promises as memory leaks.

This is especially problematic if the rendercomplete gets interrupted by a new render.

katspaugh: True, good point. I’ll approve the PR as is then.

katspaugh (approved):

dukeeagle:
Sick!!!

PR #3481, opened 2024-01-10
Feat: add part attribute to timeline

brian-byassee:
## Short description
Add part attributes to the other remaining elements rendered in the wavesurfer

Resolves #
Makes writing css easier outside the shadow root

## Implementation details
- timeline
- canvases
- canvas 
- canvas-wave
- canvas-progress
- canvas-container
- progress-container 

## Screenshots
image

## Checklist

### DIFF
--- a/src/plugins/timeline.ts
+++ b/src/plugins/timeline.ts
@@ -173,6 +173,8 @@ class TimelinePlugin extends BasePlugin<TimelinePluginEvents, TimelinePluginOpti
       },
     })
 
+    timeline.setAttribute('part', 'timeline')
+
     if (typeof this.options.style === 'string') {
       timeline.setAttribute('style', timeline.getAttribute('style') + this.options.style)
     } else if (typeof this.options.style === 'object') {
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -212,7 +212,7 @@ class Renderer extends EventEmitter<RendererEvents> {
 
       <div class="scroll" part="scroll">
         <div class="wrapper" part="wrapper">
-          <div class="canvases" part="canvases"></div>
+          <div class="canvases"></div>
           <div class="progress" part="progress"></div>
           <div class="cursor" part="cursor"></div>
         </div>
@@ -442,7 +442,6 @@ class Renderer extends EventEmitter<RendererEvents> {
   ) {
     const pixelRatio = window.devicePixelRatio || 1
     const canvas = document.createElement('canvas')
-    canvas.setAttribute('part', 'canvas canvas-wave')
     const length = channelData[0].length
     canvas.width = Math.round((width * (end - start)) / length)
     canvas.height = height * pixelRatio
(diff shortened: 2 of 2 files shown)


brian-byassee:
Completely understand and maybe our approach isn't correct which is why we are having to jump through some hoops.

We're using WaveSurfer to display a waveform with additional content before and after the `canvases` and `progress`, within the `wrapper`. Initially, elements like the `timeline-wrapper` and our custom `transcript-wrapper` appeared over the waveform, which caused overlap issues with audio spikes.

image

To accommodate these elements, we've been using JavaScript to adjust the CSS post-render. This approach, however, leads to an initial render followed by a repaint once our styles are applied. We're looking for a solution that allows CSS styling directly, avoiding this two-step rendering process.

This is our current hacked in solution because without the parts we can't use css selectors

querySelector
```js
  instance.getWrapper().style.height = `${PLAYER_HEIGHT}px`;
  instance.getWrapper().querySelector(':scope [part=timeline-wrapper] > div').style.top = `${TRIANGLE_HEIGHT}px`;
  instance.getWrapper().querySelector(':scope .canvases').style.minHeight = `${WAVE_HEIGHT}px`;
  instance.getWrapper().querySelector(':scope .progress').style.minHeight = `${WAVE_HEIGHT}px`;
  instance.getWrapper().querySelector(':scope .canvases > div').style.height = `${
    TIMELINE_HEIGHT + GAP_HEIGHT + WAVE_HEIGHT + GAP_HEIGHT
  }px`;
 ```

html
image

**Comment by katspaugh** [comment]
> Initially, elements like the timeline-wrapper and our custom transcript-wrapper appeared over the waveform, which caused overlap issues with audio spikes.

If you just want to space out the waveform a bit, you can simply set a fractional `barHeight`, e.g. 0.6 and a bigger `height`.
https://wavesurfer.xyz/examples/?609afb7f9ca8a11e40ea9eb0fea9b4e2

E.g. if you desired waveform height is 100px and you want to have 10 empty pixels above and below:

```js
WaveSurfer.create({
  height: 120,
  barHeight: 0.8, // stretch only to 80% of the height
})
```

brian-byassee:
We would like timeline and transcript plugins on their own row so that regions and waveforms will never overlap them.

image

katspaugh:
I see. Yes, in the current setup all plugins are dumped into the same wrapper div, and I can see how this can be problematic.

Regions in particular stretch to 100% of the wrapper, and AFAIR some users did prefer that. So perhaps you could reduce the height of Regions via CSS, as they do have a `part` selector.

I'll think about plugin isolation a bit more, perhaps there's a neat way to stack them instead of overlapping even if they use `position: absolute` like Regions.

brian-byassee:
@katspaugh Any thoughts or updates on this PR?

katspaugh:
I suggest that we keep the timeline part but not add the other ones as they are internal elements, not public API.

katspaugh:
@brian-byassee btw, I've invited you to be a collaborator in this repository, so that it's easier for you to make PRs. You can now push directly to this repo instead of a fork. Cheers!

brian-byassee:
@katspaugh I noticed and thank you! 🎊

katspaugh (approved):
👍

PR #3484, opened 2024-01-11
Refactor: promise-based chunked rendering

katspaugh:
## Short description

As outlined [in a comment](https://github.com/katspaugh/wavesurfer.js/pull/3480/files#r1447869479) in #3480, I've refactored the code that was awaiting all waveform chunks to be rendered into sequenced promises.

## Implementation details

The rendering timeouts are now also promise-based, and it a render timeout is cancelled due to another rendering (e.g. when zooming using a slider), it will also reject the correspondng promise. So it's possible to track the entire chain of promises to determine where `rendercomplete` should be called.

@brian-byassee @jason-whitted please review

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -107,7 +107,7 @@ class Renderer extends EventEmitter<RendererEvents> {
     this.resizeObserver = new ResizeObserver(() => {
       delay()
         .then(() => this.onContainerResize())
-        .catch(() => {})
+        .catch(() => undefined)
     })
     this.resizeObserver.observe(this.scrollContainer)
   }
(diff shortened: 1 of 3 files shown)


src/renderer.ts
```diff
@@ -575,31 +591,32 @@ class Renderer extends EventEmitter<RendererEvents> {
...
+    // Render the waveform
+    try {
+      if (this.options.splitChannels) {
+        // Render a waveform for each channel
+        await Promise.allSettled(
+          Array.from({ length: audioData.numberOfChannels }).map((_, i) => {
+            const options = { ...this.options, ...this.options.splitChannels?.[i] }
+            return this.renderChannel([audioData.getChannelData(i)], options, width)
+          }),
+        )
```
brian-byassee, line 607: `allSettled` will prevent the catch from being invoked in subsequent renders.

This seems like more render calls would be happing than one might expect.

I think `Promise.all` might be the better option here. But I may be miss understanding something here.

katspaugh: True, thank you, we want to reject the entire promise when one of them fail.

brian-byassee (approved):

PR #3496, opened 2024-01-16
Fix: Region update event

### LINKED ISSUE #3494: Adjusting region size while zoomed on part of region cause major render issues
## Bug description
When the wavesurfer is zoomed in on a region so part of the region is off screen. If the user grabs the visible end of the wavesurfer and starts to drag there is major rendering issues. Please see attached video for example. (It can be reproduced here [wavesrufer regions example](https://wavesurfer.xyz/examples/?regions.js)

https://github.com/katspaugh/wavesurfer.js/assets/47955954/dbb257cb-f449-4d17-b52a-f1d6395184dd

The bug was introduced with this [pull request](https://github.com/katspaugh/wavesurfer.js/pull/3424)
`adjustScroll` is being called for any region update not just when the region is being dragged.

One possible solution might be to add the `side` to the update event and only run `adjustScroll` when it is the region being dragged (`undefinded`) not the `start` or `end`

## Environment
 - Browser: Chrome

## Expected result
The region size is able to be adjusted with out render issues.

## Obtained result
It is impossible to adjust the size of the region in this case.

brian-byassee:
## Short description
Resolves #3494

## Implementation details
Pass the `side` variable to the update event. We only call `adjustScroll` when it is the region being dragged.

## How to test it
Use [localhost regions example](http://127.0.0.1:9090/#regions.js)
See that `adjustScroll` only fires for region drag events

*IMPORTANT NOTE: this stops a broken experience but it fails to fully implement what a user would expect to happen when resizing/dragging a region with a scrollbar.  

## Checklist

### DIFF
--- a/src/plugins/regions.ts
+++ b/src/plugins/regions.ts
@@ -25,7 +25,7 @@ export type RegionEvents = {
   /** Before the region is removed */
   remove: []
   /** When the region's parameters are being updated */
-  update: []
+  update: [side?: 'start' | 'end']
   /** When dragging or resizing is finished */
   'update-end': []
   /** On play */
@@ -270,7 +270,7 @@ class SingleRegion extends EventEmitter<RegionEvents> {
       this.end = newEnd
 
       this.renderPosition()
-      this.emit('update')
+      this.emit('update', side)
     }
   }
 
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
Thank you! 🙏

NTCarlos:
Works like a charm, awesome!

PR #3522, opened 2024-01-26
Do not render the waveform if viewport is 0

kuguarpwnz:
## Short description
Fixes memory leak that is caused by calling `draw()` in an infinite loop

## Implementation details
Check if `viewportLen` is greater than 0 before drawing (and adding canvases).
Without this check, the tail cycle will run infinitely, because `i += viewportLen` will never change the `i` value.

## How to test it
https://codesandbox.io/p/sandbox/fancy-lake-vddsjh
Inspect the DOM and see how many canvases are being added

## Screenshots
image

## Checklist

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -530,30 +530,32 @@ class Renderer extends EventEmitter<RendererEvents> {
     const end = Math.floor(start + viewportWidth * scale)
     const viewportLen = end - start
 
-    // Draw the visible part of the waveform
-    draw(start, end)
+    if (viewportLen > 0) {
+      // Draw the visible part of the waveform
+      draw(start, end)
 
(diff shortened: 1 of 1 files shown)


src/renderer.ts
```diff
@@ -530,30 +530,32 @@ class Renderer extends EventEmitter<RendererEvents> {
...
-      (async () => {
-        if (end === dataLength) return
-        const delay = this.createDelay()
-        for (let i = end; i < dataLength; i += viewportLen) {
-          await delay()
-          draw(i, Math.min(dataLength, i + viewportLen))
-        }
-      })(),
-    ])
+    if (viewportLen > 0) {
```
katspaugh, line 533: Good catch, thank you! Could you turn this into an early return instead of an if?

katspaugh (approved):
Thanks! 👍

PR #3532, opened 2024-01-29
Feat: add error event

zachkrall:
## Short description

Adds support for `wavesurfer.on('error')` event. This is helpful when building components where the file source is unknown at build time and you might want to hook into that event to display an error message.

## Implementation details

It catches errors thrown by "Fetcher.fetchBlob" and emits it as an event named `error`

## How to test it

```
yarn test -- --spec ./cypress/e2e/error.cy.js
```

## Checklist
      I think it's likely that this introduces a breaking API change if users previously relied on `.create` throwing an exception which is now caught and reported by `.on('error')`.

### DIFF
--- a/src/wavesurfer.ts
+++ b/src/wavesurfer.ts
@@ -130,6 +130,8 @@ export type WaveSurferEvents = {
   zoom: [minPxPerSec: number]
   /** Just before the waveform is destroyed so you can clean up your events */
   destroy: []
+  /** When browser is unable to fetch media file */
+  error: []
 }
 
 class WaveSurfer extends Player<WaveSurferEvents> {
@@ -236,6 +238,10 @@ class WaveSurfer extends Player<WaveSurferEvents> {
 
       this.onMediaEvent('seeking', () => {
         this.emit('seeking', this.getCurrentTime())
+      }),
+
+      this.onMediaEvent('error', () => {
+        this.emit('error')
       }),
     )
   }
@@ -366,6 +372,15 @@ class WaveSurfer extends Player<WaveSurferEvents> {
     if (!blob && !channelData) {
       const onProgress = (percentage: number) => this.emit('loading', percentage)
       blob = await Fetcher.fetchBlob(url, onProgress, this.options.fetchParams)
+        .then((blob) => blob)
+        .catch(() => {
+          this.emit('error')
+          return undefined
+        })
+    }
+
+    if (!blob) {
+      return
     }
 
     // Set the mediaelement source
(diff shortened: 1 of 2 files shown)


src/wavesurfer.ts
```diff
@@ -366,6 +372,15 @@ class WaveSurfer extends Player<WaveSurferEvents> {
     if (!blob && !channelData) {
       const onProgress = (percentage: number) => this.emit('loading', percentage)
       blob = await Fetcher.fetchBlob(url, onProgress, this.options.fetchParams)
+        .then((blob) => blob)
+        .catch(() => {
+          this.emit('error')
+          return undefined
```
katspaugh, line 378: This would be a breaking change since it's swallowing the error that previously rejecting the promise returned from `wavesurfer.load`. Should be fine to release in 7.8.0 though.

src/wavesurfer.ts
```diff
@@ -130,6 +130,8 @@ export type WaveSurferEvents = {
   zoom: [minPxPerSec: number]
   /** Just before the waveform is destroyed so you can clean up your events */
   destroy: []
+  /** When browser is unable to fetch media file */
```
katspaugh, line 133: What about media errors and decoding errors?

zachkrall:
```suggestion
  /** When media files is unable to be fetched, decoded, or an error is thrown by media element */
```

does this look better?

src/wavesurfer.ts
zachkrall: > The load function can also throw if decoding the audio has failed.

ah good to know

src/wavesurfer.ts
katspaugh: Yes, perfect!

src/wavesurfer.ts
katspaugh: Can you catch that error too?

zachkrall: @katspaugh i think it is on line 243 

https://github.com/katspaugh/wavesurfer.js/pull/3532/files#diff-3e036de76e58315f9cb343d5cc0c2a80f8299050b8f6aff522b95e52429344c1R243

katspaugh: I meant the peaks decoding [on line 402](https://github.com/katspaugh/wavesurfer.js/blob/e87542289d23d4cf243bff16ba828caf0a44ad50/src/wavesurfer.ts#L402).

zachkrall: oooh i see, yeah i can add a `.catch` there too :)

katspaugh:
One of the tests seems to be failing on this branch.

Issue #3411, opened 2023-12-13
Add new setScroll method


ffxsam:
Per https://github.com/katspaugh/wavesurfer.js/discussions/3405, it would be great to have a `setScroll` method (counterpart to `getScroll`), to be able to programmatically set the scroll position.

I'll likely work on this. Feel free to mark this issue as an enhancement.

tapeboy7:
I need this too.
What about having `setScroll` forward the command to the `private scrollIntoView` method in renderer.ts?

katspaugh:
`scrollIntoView` won't set the scroll to an arbitrary point, it only makes sure the playhead is visible.

It should be a new renderer method that simply sets `this.scrollContainer.scrollLeft`.

PR #3547, opened 2024-02-11
Feat: Add setScrollTime method

jagthedrummer:
## Short description

This adds a `setScrollTime` method to `wavesurfer` that takes a time in seconds. It also adds `setScroll` and `setScrollPercentage` methods to `renderer` which are used by `setScrollTime`.

This allows the viwing window to be set indepenently of the play head.

Resolves: https://github.com/katspaugh/wavesurfer.js/issues/3411
Related: https://github.com/katspaugh/wavesurfer.js/discussions/3361

## Implementation details

`setScrollTime` uses the current duration to calculate the percentage within the file. It forwards that percentage to `setScrollPercentage` in the `renderer`, which then uses the width of the scroll container to calculate an exact offset to pass to `setScroll`.

## How to test it

```
yarn cypress
```

## Checklist

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -249,7 +249,7 @@ class Renderer extends EventEmitter<RendererEvents> {
     return this.scrollContainer.scrollLeft
   }
 
-  setScroll(pixels: number) {
+  private setScroll(pixels: number) {
     this.scrollContainer.scrollLeft = pixels
   }
 
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -249,6 +249,16 @@ class Renderer extends EventEmitter<RendererEvents> {
     return this.scrollContainer.scrollLeft
   }
 
+  setScroll(pixels: number) {
+    this.scrollContainer.scrollLeft = pixels
+  }
+
+  setScrollPercentage(percent: number) {
+    const { scrollWidth } = this.scrollContainer
+    const scrollStart =  scrollWidth * percent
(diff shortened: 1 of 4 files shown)


src/renderer.ts
```diff
@@ -249,6 +249,16 @@ class Renderer extends EventEmitter<RendererEvents> {
     return this.scrollContainer.scrollLeft
   }
 
+  setScroll(pixels: number) {
```
katspaugh, line 252:
```suggestion
  private setScroll(pixels: number) {
```

src/wavesurfer.ts
```diff
@@ -350,6 +350,12 @@ class WaveSurfer extends Player<WaveSurferEvents> {
     return this.renderer.getScroll()
   }
 
+  /** Move the start of the viewing window to a specific time in the audio (in seconds) */
+  public setScrollTime(time: number) {
+    let percentage = time / this.getDuration()
```
katspaugh, line 355:
```suggestion
    const percentage = time / this.getDuration()
```

katspaugh (commented):
Thanks, looks great! Left two small comments.

katspaugh (approved):
🚀

katspaugh:
Thank you too!

We can release anytime. You can create a PR that just bumps the version, and CI will do the rest.

jagthedrummer:
Ah, good to know!

PR #3555, opened 2024-02-15
Feat: Add preload option

### LINKED ISSUE #3435: Passing peaks and duration (predecoded) still downloads content in chrome
## Bug description
When passing peaks and duration audio is still downloaded at least partially in chrome. Firefox does not seem to do it.

How to reproduce:
1-Open devtools and check networking tab in chrome
2-Access: https://wavesurfer.xyz/examples/?predecoded.js

You can observer that demo.wav is partially downloaded. Around 200KB is downloaded and when I click play the rest (~400KB) is downloaded.

## Environment
 - Browser: Chrome

## Expected result
Should not download any content, and only on play should start streaming.

bernardwiesner:
## Short description
Resolves #3435. Add the ability to set preload option. https://developer.mozilla.org/docs/Web/API/HTMLMediaElement/preload

## Implementation details
```js
  container: document.body,
  preload: 'none'
})
```

### DIFF
--- a/src/player.ts
+++ b/src/player.ts
@@ -5,6 +5,7 @@ type PlayerOptions = {
   mediaControls?: boolean
   autoplay?: boolean
   playbackRate?: number
+  preload?: 'none' | 'metadata' | 'auto' | ''
 }
 
 class Player<T extends GeneralEventTypes> extends EventEmitter<T> {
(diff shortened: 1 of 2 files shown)


katspaugh:
I wrote in #3435 that I'm open to adding this option but I'm having second thoughts now.

Consider the following cases:

1. Wavesurfer is initialized with basic options and an audio URL:
 * Wavesurfer will download the audio as a blob
 * Sets its media element src to that blob – `preload` won't do anything

2. Wavesurefer is initialized with an audio URL and peaks:
* The media element src is set to the URL directly
* Preload works

3. Wavesurefer is initialized with a custom media element:
* Preload also works, but could as well be set directly on the media element

---

Case 1 is probably 80% of people, and it might be confusing to them that this parameter has no effect.

So we should probably at least mention it in the docs.

katspaugh:
As I mentioned above, I think this option would confuse most of the users who don't pre-decode their audio, as it won't have any effect. So I'm closing the PR, sorry. Thank you for the contribution!

PR #3590, opened 2024-03-11
add dragDebounceTime as an option

### LINKED ISSUE #3589: add debounce time as an option
It would be nice to override the default debounce time so that scrubbing the waveform sends a time update event as fast as it can. Specifically, I would use this in my application that relies on real-time updates based on the scrub head position.

Pull request to follow.

[Original issue](https://github.com/katspaugh/wavesurfer.js/issues/61)

phill2mj:
## adds dragDebounceTime as an option and uses it as the debounce time in ms when dragging the timeline and `dragToSeek` is true. A default is set at 200ms.
Resolves #3589

## Checklist

### DIFF
--- a/src/wavesurfer.ts
+++ b/src/wavesurfer.ts
@@ -74,6 +74,8 @@ export type WaveSurferOptions = {
   fetchParams?: RequestInit
   /** Playback "backend" to use, defaults to MediaElement */
   backend?: 'WebAudio' | 'MediaElement'
+  /** Number of ms to debounce dragging when `dragToSeek` is true; defaults to 200ms */
+  dragDebounceTime?: number
 }
 
 const defaultOptions = {
@@ -87,6 +89,7 @@ const defaultOptions = {
   autoScroll: true,
   autoCenter: true,
   sampleRate: 8000,
+  dragDebounceTime: 200,
 }
 
 export type WaveSurferEvents = {
@@ -311,7 +314,7 @@ class WaveSurfer extends Player<WaveSurferEvents> {
             () => {
               this.seekTo(relativeX)
             },
-            this.isPlaying() ? 0 : 200,
+            this.isPlaying() ? 0 : this.options.dragDebounceTime,
           )
 
           this.emit('interaction', relativeX * this.getDuration())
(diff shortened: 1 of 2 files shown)


src/wavesurfer.ts
```diff
@@ -74,6 +74,8 @@ export type WaveSurferOptions = {
   fetchParams?: RequestInit
   /** Playback "backend" to use, defaults to MediaElement */
   backend?: 'WebAudio' | 'MediaElement'
+  /** Number of ms to debounce dragging when `dragToSeek` is true; defaults to 200ms */
+  dragDebounceTime?: number
```
katspaugh, line 78: Should we maybe combine it with the `dragToSeek`? If false it should be disabled, if a number, it’s a debounce value and it’s enabled. It true, then 200.

phill2mj: IMO that kind of overload could be confusing. Maybe if the type was `boolean | { debounceTime: number }`, but so far, that kind of layering doesn't match the rest of options.

The only other options that accept radically different types are the css passthroughs like `width?: number | string` but in that case I think it makes sense.

But it's your library, so I'll defer :) Either way is usable, and I'd love to use this feature!

katspaugh: My thinking is that since drag-to-seek is already itself a togglable option, adding another dependent option that won’t do anything if dragToSeek isn’t enabled would also be somewhat confusing. Perhaps if it was called `dragToSeekDebounce` it would be more clear.
Your suggestion to make it a nested object makes sense to me tho, I would go with that.

phill2mj: > "adding another dependent option that won’t do anything if dragToSeek isn’t enabled would also be somewhat confusing"

Very true, I didn't think of that. And with adding the nested object as a union to the boolean enable/disable type, it stays backwards compatible. I'll update.

src/wavesurfer.ts
```diff
@@ -307,11 +313,21 @@ class WaveSurfer extends Player<WaveSurferEvents> {
...
           // Set the audio position with a debounce
           clearTimeout(debounce)
+          let debounceTime;
+
+          if (this.isPlaying()) {
+            debounceTime = 0;
+          } else if (this.options.dragToSeek === true) {
+            debounceTime = 200;
+          } else if (typeof this.options.dragToSeek === 'object' && this.options.dragToSeek !== undefined) {
+            debounceTime = this.options.dragToSeek['debounceTime'];
```
phill2mj, line 323: I don't like accessing by field name but the type system wouldn't have it any other way that I could think of. The rest of the added code tries to ensure that this is safe.

PR #3614, opened 2024-03-29
Fix: potential memory leak

wfk007:
## Short description
After wavesurfer destroyed, the `ready` event will still emit, which will cause a memory leak in some cases such as invoking wavesurfer apis in setInterval callback.

## Implementation details
Add abortController and signal, when wavesurfer destroyed, cancel the blob Fetcher and stop ready event emit.

## How to test it
1. git clone https://github.com/wfk007/test-wavesurfer
2. pnpm install
3. pnpm run dev
4. toggle between `Home` and `Docs` and observe the performance in Chrome devtools
5. a memory leak occurred

## Screenshots
image

image

## Checklist


katspaugh (approved):
Great catch! 🙏

PR #3624, opened 2024-04-03
Fix: makeDraggable memory leak

wfk007:
## Short description
`makeDraggable` has memory leak in `renderer` and `plugins`(envelope and regions)

## Implementation details
Using subscriptions to collect the return function of `makeDraggable` and unsubscribe it when destroyed or moved.

## How to test it
1. git clone https://github.com/wfk007/test-wavesurfer
2. pnpm install
3. pnpm run dev
4. visit http://localhost:8000/docs
5. open performance in devtools
6. GC manually and start a new record
7. toggle between `Docs` and `test dragToSeek option` many times
8. stopped at `Docs` page, GC manually and stop the record

## Screenshots

before
image

after
image

## Checklist

### DIFF
--- a/src/plugins/envelope.ts
+++ b/src/plugins/envelope.ts
@@ -55,10 +55,12 @@ class Polyline extends EventEmitter<{
       circle: SVGEllipseElement
     }
   >
+  private subscriptions: (() => void)[] = []
 
   constructor(options: Options, wrapper: HTMLElement) {
     super()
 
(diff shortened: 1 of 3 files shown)


katspaugh (approved):
Oh, thank you, good catch!
It's a very recent regression introduced in #3616 where makeDraggable started subscribing to document events, before it was done inside a pointerdown event on the draggable element itself.

PR #3628, opened 2024-04-03
feat(hover): add formattimecallback option

Adrii77:
## Short description

Add `formatTimeCallback` option for Hover plugin

## Implementation details

Reuse previous `formatTime` method as a new option named `formatTimeCallback`.

## How to test it

- yarn start
- Go to Hover plugin's page on the app
- Hover the waveform to see plugin's working
- In Hover.create object in params, add 
```
formatTimeCallback(time) {
            const hours = Math.floor(time / 3600);
            const minutes = Math.floor((time % 3600) / 60);
            const seconds = Math.floor(time % 60);

            const formattedTime = [hours, minutes, seconds]
              .map(value => (value < 10 ? `0${value}` : value))
              .join(':');

            return formattedTime;
          },
```
- Hover the waveform to see time label as "hh:mm:ss"

## Checklist

### DIFF
--- a/src/plugins/hover.ts
+++ b/src/plugins/hover.ts
@@ -11,11 +11,19 @@ export type HoverPluginOptions = {
   labelColor?: string
   labelSize?: string | number
   labelBackground?: string
+  /* Allows to customize time label */
+  formatTimeCallback?: (seconds: number) => string
 }
 
 const defaultOptions = {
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
Perfect, thank you! 🙏

PR #3660, opened 2024-04-15
update: expose setScroll api from render

leyiang:
## Short description
Resolves #3659

I didn't check pixels in `setScroll` because it's redundant, setting invalid value for scrollLeft will be set as left or right edge.


leyiang:
cypress test passed

katspaugh (approved):
Thanks!🙏

PR #3685, opened 2024-05-01
fix: when width is set high, now multiple canvases are rendered

### LINKED ISSUE #3682: If the wavesurfer container has a very large width, only one canvas will be rendered.
I replicated this on a forked version of wavesurfer where I ran `npm run start` to start the local examples.

If the body, i.e., the wavesurfer's container, is given a very large width (e.g., 70000px), then the wavesurfer will only create one single canvas that is 70000px large. The expected behavior is that is creates lots of smaller canvases, such as 70 1000px canvases.

It also seems that this bug replicates when `fillParent` is both true and false.

I am currently fixing this on my fork and could contribute to wavesurfer by providing some code when I figure out what the fix might be.

Johnrobmiller:
## Short description
Resolves #3682

## Implementation details
Please see #3682

## How to test it
Please see #3682

## Screenshots
Please see #3682

## Checklist

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -532,7 +532,9 @@ class Renderer extends EventEmitter<RendererEvents> {
     }
 
     // Draw the entire waveform
-    if (!this.isScrollable) {
+    // Note, when the waveform container's width is set to a very large value, then the waveform will not be scrollable.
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
Awesome, thank you!

PR #3705, opened 2024-05-14
Fix: Split channel overflowing when height:auto

htc36:
## Short description
Resolves https://github.com/katspaugh/wavesurfer.js/discussions/3690

## Implementation details

Alter getHeight function to divide the parentHeight by number of channels if the height=auto of the wavesurfer instance

## How to test it
- Give the parent container a height
- Set `height : auto` & `splitChannel : []` of Wavesurfer instance
- Observe the waveforms not overflowing outside the parent container as their heights are 

## Screenshots (Observe split channels is half the size of the parent)

splitChannel=[] | splitChannel=undefined
--- | --- 
image |  image

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -147,11 +147,19 @@ class Renderer extends EventEmitter<RendererEvents> {
     )
   }
 
-  private getHeight(optionsHeight?: WaveSurferOptions['height']): number {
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
🙏

PR #3747, opened 2024-06-16
Make yarn build command cross-platform

### LINKED ISSUE #3746: Build: yarn build fails on Windows
## Bug description
**Not a bug in the library - only impacts local development**
`yarn build` fails on Windows because of `rm -rf dist`

## Environment
 - OS: Windows

alexanderleecii:
## Short description
The `build` npm script currently fails on non-unix platforms where `rm` is not defined. This PR makes the script cross-platform.
Resolves #3746

## Implementation details
1. Created a `scripts/clean.cjs` script to recursively delete `dist` folder using nodejs APIs
2. Added a `clean` npm script which calls `scripts/clean.cjs`
3. Call `npm run clean` instead of `rm -rf dist` in `build` npm script

## How to test it
Run `yarn build`. Any existing `dist` folder should be removed and the build process should execute correctly.

## Checklist

### DIFF
--- a/scripts/clean.cjs
+++ b/scripts/clean.cjs
@@ -1,9 +1,9 @@
-const path = require("path")
-const fs = require("fs")
+const path = require('path')
+const fs = require('fs')
 
 const run = () => {
-    const distPath = path.join(__dirname, '../dist')
-    fs.rmSync(distPath, { recursive: true, force: true })
+  const distPath = path.join(__dirname, '../dist')
+  fs.rmSync(distPath, { recursive: true, force: true })
 }
 
 run()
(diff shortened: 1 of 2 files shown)


katspaugh (approved):
Thanks!

scripts/clean.cjs
```diff
@@ -0,0 +1,9 @@
+const path = require("path")
+const fs = require("fs")
+
+const run = () => {
+    const distPath = path.join(__dirname, '../dist')
+    fs.rmSync(distPath, { recursive: true, force: true })
+}
+
+run()
```
katspaugh, line 9:
```suggestion
const path = require('path')
const fs = require('fs')

const run = () => {
  const distPath = path.join(__dirname, '../dist')
  fs.rmSync(distPath, { recursive: true, force: true })
}

run()
```

katspaugh (approved):
👍

PR #3749, opened 2024-06-16
Regions: allow overriding of css styles of region div and handle divs.

rewbs:
## Short description

Allows for general CSS overrides for regions and their handles.

While this allows for a wide range of customisations, my immediate use-cases for adding this were:
* Creating wider handles 
* Setting the z-offset to ensure some regions are always rendered "above" others.

It's very possible I missed a way to do this with existing code. I'm also conscious that you might not want to allow arbitrary CSS overrides, in favour of more controlled options or custom HTML elements (as is done for the region content).  

## Implementation details

Added additional fields to region params which are overlayed onto CSS options for the relevant divs on creation.

## How to test it

Specify custom overrides as part of region params and ensure they render.

## Screenshots

https://github.com/katspaugh/wavesurfer.js/assets/74455/ad85ad8f-7d4d-427b-889d-5cae86698824

## Checklist

Sorry, **no tests yet** – if there's consensus this is a good approach I'll happily help put tests together. :)

### DIFF
--- a/src/plugins/regions.ts
+++ b/src/plugins/regions.ts
@@ -63,6 +63,14 @@ export type RegionParams = {
   channelIdx?: number
   /** Allow/Disallow contenteditable property for content */
   contentEditable?: boolean
+  /** Additional CSS styles to be applied to the region div. */
+  styleOverrides?: { [key: string]: string }
+  /** Additional CSS styles to be applied to the region handle divs. */
+  handleStyleOverrides?: { [key: string]: string }
+  /** Additional CSS styles to be applied to the left region handle div. */
+  leftHandleStyleOverrides?: { [key: string]: string }
+  /** Additional CSS styles to be applied to the right region handle div. */
+  rightHandleStyleOverrides?: { [key: string]: string }
 }
 
 class SingleRegion extends EventEmitter<RegionEvents> {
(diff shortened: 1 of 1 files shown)


katspaugh:
Hey @rewbs, thanks for the PR!

Technically regions are already styleable via CSS as demonstrated here: https://wavesurfer.xyz/examples/?styling.js

Although it's not entirely the same as passing styles in JS, it's still pretty flexible. You can, for example, give a region an id, and use that to set a custom z-index. In the example I linked, the region with the id `region-green` is given custom styles this way.

rewbs:
For anyone else coming across this in the future: the challenge I was facing was attempting to style subsets of regions in different ways. In my usecase, users can create 2 types of markers which I want to style differently. I also need the markers to have unique IDs.

Wavesurfer lets you style regions and markers using their [CSS Part ](https://developer.mozilla.org/en-US/docs/Web/CSS/::part) name, which is taken [verbatim from the region ID](https://github.com/katspaugh/wavesurfer.js/blob/main/src/plugins/regions.ts#L120). However, CSS does not support wildcards in Part selectors. So how can you style a group of regions which are related only by some substring in their region ID?

The solution is to **use spaces in the region ID**: each space delimited substring will be treated as a different Part identifier. So if you create regions with the following IDs, you style all 4 in two groups by declaring 2 CSS entries that select "manual" and "detected" respectively:

```
region1.id='manual 123'
region2.id='manual 456'
region3.id='detected 789'
region4.id='detected 098'
```

PR #3760, opened 2024-06-21
Add mediaRecorderTimeslice option to RecordPluginOptions

nizar:
## Short description
On Safari on both Mac and iOs, the audio generated by the record plugin cannot be properly processed by the OpenAI whisper model. This is a known issue that is addressed by passing a timeslice param of 1000 to the mediaRecorder.start() call.

## Implementation details
If the optional `mediaRecorderTimeslice` value is provided to the RecordPlugin initializer, the value is passed to the `mediaRecorder.start()` call.

## How to test it
Pass a value of 1000 when running on any browser on iOS or in Safari on MacOS, recorded audio will be processed by OpenAI whisper. Without this value, only the first second or so of audio is correctly processed.

## Screenshots
N/A

## Checklist

### DIFF
--- a/src/plugins/record.ts
+++ b/src/plugins/record.ts
@@ -16,6 +16,8 @@ export type RecordPluginOptions = {
   scrollingWaveform?: boolean
   /** The duration of the scrolling waveform window, defaults to 5 seconds */
   scrollingWaveformWindow?: number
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
Thanks!

PR #3788, opened 2024-07-19
refactor FFT

ndunks:
Just refactor FFT. and planed to add MFCC spectrogram


katspaugh:
Hi @ndunks!

Thanks for the PR!
Can I you elaborate on what you’re aiming to achieve with this refactoring?

ndunks:
I Fixing some typescript issue in spectrogram FFT..
And restructuring the spectrogram plugin and FFT class,  to support other method like MFCC instead of only FFT, I plan to implement it.
If you didn't like, just ignore this PR, I just want to share it.

katspaugh:
It's not that I don't like it, I just lack the context on why you're proposing it. Plus I see it touches multiple files unrelated to the spectrogram.

ndunks:
I see another PR with mel scale, that was I mean. I will use that PR after merged.. and I think this project has some issue with typescript typings.. I will do another PR and spitting out changes. I know Typescript but new for audio processing. thanks.

Issue #3140, opened 2023-08-25
Spectrogram's canvas does not full fill the container of it


KrazyPhish:
## Bug description
I was doing a Vue2 to Vue3 project, found that spectrogram's canvas cannot full fill its parent element.
But it works well in my Vue2 project.
Infomation of the versions: 
 my Vue2 project: wavesurfer.js 6.2.0 (2022-05-16)
 my Vue3 project: wavesurfer.js latest

## Environment
 - Dev: VSCode, TypeScript, Vue3
 - Browser: Chrome - version 116.0.5845.97
 - OS: windows 11

## Minimal code snippet
In Vue2:
 plugins.push(
  spectrogram.create({
    container: this.$refs.waveSpectrogramRef,
    labels: true,
    height: height * this.verticalZoom.value,
    colorMap
  })
)

In Vue3: 
 const spectrogram: SpectrogramPlugin = SpectrogramPlugin.create({
  container: waveSpectrogramRef.value as HTMLElement,
  labels: true,
  height: height * state.verticalZoom,
  colorMap
})

The value height equals its parent's clientHeight, it's a constant value, so is the verticalZoom value.

## Expected result
Like wavesurfer works in my Vue2 project, spectrogram's canvas fully fill its parent element.

## Obtained result
Same as Expected result

## Screenshots
 The parent element div.
![image](https://github.com/katspaugh/wavesurfer.js/assets/52484425/da0e40df-a0bd-4587-8b21-9478153105ec)
 The left side label canvas, this means spectrogram tried to fill the element, I guess.
![image](https://github.com/katspaugh/wavesurfer.js/assets/52484425/f679577c-d8c1-474b-aa1f-a5cb08d660dc)
The style attributes show that the canvas has fully filled its parent element, there must be some problems with canvas while the plugin working with canvas.
![image](https://github.com/katspaugh/wavesurfer.js/assets/52484425/4fca79dc-2110-4eba-a8da-20549354d6d3)

katspaugh:
You might need to increase the sampleRate for it to have the data to fill the chart. Try passing 22050 or 44100 like here: https://wavesurfer-js.org/examples/?spectrogram.js

KrazyPhish:
Attribute sampleRate 22050 works the same as my last try, 44100 makes a little diffrence but still didn't solve the problem, here's the pic in 44100 samepleRate, it seems that only expanded the label values:
[...]

PR #3796, opened 2024-07-22
Fix: [Spectrogram] fix cropping and scaling issues

### LINKED ISSUE #3663: fftSamples shouldn't scale the spectrogram image
## Bug description
On the page [https://wavesurfer.xyz/examples/?spectrogram.js](https://wavesurfer.xyz/examples/?spectrogram.js), set the **Spectrogram** plugin option **fftSamples** from 1024 to 128 (or others), the height of the spectrogram image will change.

## Environment
 - Browser: Chrome

## Expected result
The **fftSamples** parameter only affects the resolution of the image, not its height.
I think the final image should be stretched to cover the whole range.
If the sample rate is 48000Hz, the image should always cover 0 - 24000Hz, independent of **fftSamples**.
The lower **fftSamples** means the highlights in the image will be more blur, more like a mosaic effect.

## Screenshots
image
image

jonom:
## Short description

Resolves #3663, #3140

## Implementation details

I've attempted to address some issues with the spectrogram rendering creating an image of the wrong height for the target container, causing it to be too tall or short to fill the area, and meaning the frequency labels do not line up with data in the image. There are some magic numbers in here (E.g. division by 2 or 4) I don't understand so appreciate if someone else can vet this!

(Marked as draft for now as I'm running tests locally but it's taking quite a while.)

Separate to this I've implemented some caching of the result of `getFrequencies()` as well as the rendered bitmaps as this can be quite slow, and for my use case we allow zooming in and out so I wanted to speed up the rendering. If this is something you'd like included I can add to this PR or open another one later.

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Enhanced audio playback quality with an updated sample rate.
	- Added new configuration options for the Spectrogram, including frequency range and scale settings.
- **Bug Fixes**
	- Improved error handling for color map validation.
- **Refactor**
	- Streamlined frequency scaling and rendering logic in the Spectrogram plugin.

### DIFF
--- a/examples/spectrogram.js
+++ b/examples/spectrogram.js
@@ -9,7 +9,7 @@ const ws = WaveSurfer.create({
   waveColor: 'rgb(200, 0, 200)',
   progressColor: 'rgb(100, 0, 100)',
   url: '/examples/audio/audio.wav',
-  sampleRate: 22050,
+  sampleRate: 44100,
 })
 
 // Initialize the Spectrogram plugin
@@ -18,6 +18,11 @@ ws.registerPlugin(
     labels: true,
     height: 200,
     splitChannels: true,
+    scale: 'mel', // or 'linear'
+    frequencyMax: 8000,
+    frequencyMin: 0,
+    fftSamples: 1024,
+    labelsBackground: 'rgba(0, 0, 0, 0.1)',
   }),
 )
 
--- a/src/plugins/spectrogram.ts
+++ b/src/plugins/spectrogram.ts
@@ -247,7 +247,7 @@ export type SpectrogramPluginOptions = {
   alpha?: number
   /** Min frequency to scale spectrogram. */
   frequencyMin?: number
-  /** Max frequency to scale spectrogram. Set this to samplerate/2 to draw whole range of spectrogram. */
+  /** Max frequency to scale spectrogram. Set this to samplerate/4 to draw whole range of spectrogram. */
   frequencyMax?: number
   /**
    * Based on: https://manual.audacityteam.org/man/spectrogram_settings.html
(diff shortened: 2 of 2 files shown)


katspaugh:
@jonom thanks for the PR! How did your tests go, did it solve those issues for you?

jonom:
I'm not exactly sure how the e2e tests are supposed to be ran - after a while I worked out I could run them manually, but I didn't see snapshots to compare with expected results.

One thing I have found since is that setting sampleRate on the ws instance doesn't seem to affect rendering, except that if it's too small, the upper range of the spectrogram may be missing. So I've been setting it to 88200 and then it seems to always work 🤷‍♂️

katspaugh:
I've commented out Spectrogram tests for now, they mysteriously always fail on GitHub Actions but work locally.

abbott:
Howdy, the current state of the spectrogram height is broken. The current v7 library renders inconsistent height values in the DOM, which is a struggle to produce a basic spectrogram visual. 

```
<div id="spectrogram" style="width: 100%;">
    <div style="display: block; position: relative; user-select: none; width: 100%; overflow: hidden; height: 64px;">
        <canvas part="spec-labels" style="position: absolute; z-index: 9; width: 55px; height: 100%;" height="128" width="110"></canvas>
        <canvas style="position: absolute; left: 0px; top: 0px; width: 100%; height: 100%; z-index: 4;" width="537" height="64"></canvas>
    </div>
</div>
```

Does the spectrogram have other dependencies? I have followed conversations on sample rate, but it's unclear if those are hacks?

```
SpectrogramPlugin.create({
          container: "#spectrogram",
          labels: true,
          height: 80,
          ...
```

jonom:
@katspaugh sorry for the slow reply, I didn't realise the Mel scale had been added and the fixes I supplied previously only worked for the Linear scale. I've worked with @lucastheis to update the fixes and the Mel scale should work a lot better now, being rendered at the appropriate height and also fixing a few bugs with the scale labels.

If you could let me know what I would need to do to get this over the line that would be great. No rush - I'll be off until the new year now. If you're taking some time off as well, happy holidays!

image

katspaugh (approved):
Amazing, thank you! Happy holidays to you as well!

PR #3818, opened 2024-08-10
Feat: [Record] continuous waveform mode

katspaugh:
## Short description

A new preview option for the Record plugin: `continuousWaveform`. It shows the audio recorded so far as a continuously growing waveform.

## Implementation details

Along with the new `continuousWaveform` parameter, there's an optional `continuousWaveformDuration` that allows setting the expected duration of a recording.

## How to test it
Open the [record.js example](https://continuous.wavesurfer-js.pages.dev/#record.js) and record something.

## Screenshots
Screenshot 2024-08-10 at 18 37 26


lofcz:
@katspaugh this is great, can this work without making the container scroll once the recording duration > `continuousWaveformDuration`, similar to how `scrollingWaveform` works?

image

PR #3848, opened 2024-08-30
Optimized Method for Calculating Overlap Size in Spectrogram Resampling

shellRaining:
## Short description
Resolves #3847

### DIFF
--- a/src/plugins/spectrogram.ts
+++ b/src/plugins/spectrogram.ts
@@ -728,7 +728,7 @@ class SpectrogramPlugin extends BasePlugin<SpectrogramPluginEvents, SpectrogramP
         const oldEnd = oldStart + oldPiece
         const newStart = i * newPiece
         const newEnd = newStart + newPiece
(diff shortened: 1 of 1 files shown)


src/plugins/spectrogram.ts
```diff
@@ -728,12 +728,8 @@ class SpectrogramPlugin extends BasePlugin<SpectrogramPluginEvents, SpectrogramP
         const oldEnd = oldStart + oldPiece
         const newStart = i * newPiece
         const newEnd = newStart + newPiece
+        const overlap = Math.max(0, Math.min(oldEnd, newEnd) - Math.max(oldStart, newStart));
```
katspaugh, line 731:
```suggestion
        const overlap = Math.max(0, Math.min(oldEnd, newEnd) - Math.max(oldStart, newStart))
```

Lint will complain otherwise (`yarn prettier`).

katspaugh (approved):
Thanks!

PR #3850, opened 2024-09-02
fix: #3707

### LINKED ISSUE #3707: the total duration wavesurfer is giving me is wrong
I have a audio file that is exactly 6.84 seconds long. I use js to and my OS to confirm that it exactly that duration. However, when I load it into Wavesurfer and call `ws.getDuration()`, wavesurfer is wrongly reporting that it is 6.81795 seconds long. This was causing a bug in my app that was difficult to troubleshoot.

The duration that wavesurfer is outputting ought to be exactly the same as the actual duration of the audio file.

Note: wavesurfer gets the duration wrong even if it is manually set in the options, as noted here https://github.com/katspaugh/wavesurfer.js/discussions/3656

shellRaining:
## Short description
Resolves #3707 

## Implementation details

exchange the order of audioEl and audioContext duration

## Screenshots

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,6 +78,7 @@
     "eslint": "^8",
     "eslint-config-prettier": "^8",
     "eslint-plugin-prettier": "^4",
+    "glob": "^11.0.0",
     "prettier": "^2",
     "rollup": "^4.17.2",
     "rollup-plugin-dts": "^6.1.0",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,9 +1,9 @@
-import glob from 'glob'
+import { glob } from 'glob'
 import typescript from '@rollup/plugin-typescript'
 import terser from '@rollup/plugin-terser'
 import dts from 'rollup-plugin-dts'
 
-const plugins = [typescript(), terser({ format: { comments: false } })]
+const plugins = [typescript({ declaration: false, declarationDir: null }), terser({ format: { comments: false } })]
 
 export default [
   // ES module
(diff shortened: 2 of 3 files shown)


katspaugh:
The problem is, playback is still happening through an audio element. With this change, getCurrentTime will be inconsistent with getDuration.

Besides, we'd need to adjust this code too then: https://github.com/katspaugh/wavesurfer.js/blob/1a0cc9de96492bc208690b46b07bb4bfade6cb3c/src/wavesurfer.ts#L430-L440

shellRaining:
How about we subscribe to the `decoded` event of the wavesurfer instance, and call `getDuration` function again when this event is triggered?

src/wavesurfer.ts
```diff
@@ -433,9 +433,7 @@ class WaveSurfer extends Player<WaveSurferEvents> {
       if (staticDuration) {
         resolve(staticDuration)
       } else {
-        this.mediaSubscriptions.push(
-          this.onMediaEvent('loadedmetadata', () => resolve(this.getDuration()), { once: true }),
-        )
+        this.on('decode', (duration) => resolve(duration), { once: true })
```
katspaugh, line 436: Isn’t this self-referential? This event is emitted at the end of this very method (loadAudio).

shellRaining: Sorry for pushing it up without thorough testing, I'll try again these few days.

katspaugh: No worries! We’ll probably have to split it into two checks: for an externally passed duration here, and for a decoded duration further down after the decoding.

PR #3858, opened 2024-09-05
Fix: adding nonce as a option in wavesurfer parameters to solve CSP problems

### LINKED ISSUE #3602: CSP set to nonce, <style nonce="QWHzeTLwpGPwTNySpBKyHH8k"> not set
## Bug description
If CSP policy is strict and requires a nonce, the generated <style> block wont work

## Environment
 - Browser: Firefox

## Minimal code snippet
Content-Security-Policy | script-src 'nonce-bTFivopG26dOZPgqLdM1Yg27'; style-src 'nonce-QWHzeTLwpGPwTNySpBKyHH8k'

## Expected result
<style nonce="QWHzeTLwpGPwTNySpBKyHH8k">
        :host {
          user-select: none;
          min-width: 1px;
        }
...
</style>

Or a seperate .css file would also work.

## Obtained result
<style>
        :host {
          user-select: none;
          min-width: 1px;
        }

## Screenshots
afbeelding

andrezarzur:
## Short description
Resolves #3602

## Implementation details
I've added the nonce property as a new parameter in the WaveSurferOptions type. In the renderer.ts it checks if it was passed, and if it was, it adds it to the style being injected. This assumes you have the nonce value in your frontend to be able to be passed as a option.

## How to test it
Try adding a valid nonce value to your WaveSurfer element like this: {nonce: 'example-value'}

## Screenshots
image
image

## Checklist

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -169,10 +169,10 @@ class Renderer extends EventEmitter<RendererEvents> {
     const div = document.createElement('div')
     const shadow = div.attachShadow({ mode: 'open' })
 
-    const nonce = this.options.nonce;
+    const cspNonce = this.options.cspNonce && typeof this.options.cspNonce === 'string' ? this.options.cspNonce.replace(/"/g, '') : '';
 
     shadow.innerHTML = `
-      <style ${nonce ? `nonce="${nonce}"` : ''}>
+      <style${cspNonce ? ` nonce="${cspNonce}"` : ''}>
         :host {
           user-select: none;
           min-width: 1px;
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -169,8 +169,10 @@ class Renderer extends EventEmitter<RendererEvents> {
     const div = document.createElement('div')
     const shadow = div.attachShadow({ mode: 'open' })
 
+    const nonce = this.options.nonce;
+
     shadow.innerHTML = `
-      <style>
+      <style ${nonce ? `nonce="${nonce}"` : ''}>
         :host {
           user-select: none;
           min-width: 1px;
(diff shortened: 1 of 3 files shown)


src/wavesurfer.ts
```diff
@@ -76,6 +76,8 @@ export type WaveSurferOptions = {
   fetchParams?: RequestInit
   /** Playback "backend" to use, defaults to MediaElement */
   backend?: 'WebAudio' | 'MediaElement'
+  /** Nonce for CSP if necessary */
+  nonce?: string
```
katspaugh, line 80: Should we call it cspNonce maybe?

src/renderer.ts
```diff
@@ -169,8 +169,10 @@ class Renderer extends EventEmitter<RendererEvents> {
     const div = document.createElement('div')
     const shadow = div.attachShadow({ mode: 'open' })
 
+    const nonce = this.options.nonce;
+
     shadow.innerHTML = `
-      <style>
+      <style ${nonce ? `nonce="${nonce}"` : ''}>
```
katspaugh, line 175: This should be sanitized for XSS. Something like nonce.replace(/"/g, '')

src/wavesurfer.ts
andrezarzur: Possibly, I was just following the pattern I saw other libs using, which is this one, but cspNonce would be more explict.

src/renderer.ts
katspaugh: Better above where you do const nonce

andrezarzur: @katspaugh I've also added some type validation just to avoid any errors

src/wavesurfer.ts
andrezarzur: Done

katspaugh (approved):
LGTM, thank you! ❤️

PR #3861, opened 2024-09-09
Fix: [Spectrogram] Resample spectrogram at Mel frequencies without ch…

lucastheis:
This changes `melFilterBank` such that power spectra are correctly resampled.

## Short description

The way that `melFilterBank` is computed now, the Mel spectrogram is _not_ calculated by linearly interpolating the linear spectrogram at nearby frequencies. Instead, it computes a weighted average where the sum of weights depends on the frequency. One consequence of this is that the Mel spectrogram of white noise does not appear to be white.

## Implementation details

The Mel spectrogram is now calculated by resampling and linearly interpolating the linear spectrogram.

## Screenshots

Spectrogram of Gaussian white noise before: 
mel_spectrogram_before

Spectrogram of Gaussian white noise after:
mel_spectrogram_after

`melFilterBank` before:
filterbank_before
sums_before

`melFilterBank` after:
filterbank_after
sums_after

## Checklist


katspaugh (approved):
Thank you!

PR #3934, opened 2024-11-19
Feat(Zoom): Add exponential zooming and a few bugfixes

gustavsollenius:
### **Changes Summary**:
1. Added exponential zooming option (`exponentialZooming`):
   - Default: `false` (uses linear zooming if disabled).
   - When enabled, zoom transitions use an exponential formula for consistent zoom effects. The reference to this formula is found here: [Smooth Zoom Formula](https://math.stackexchange.com/questions/4441136/smooth-zoom-formula).
   -  `iterations` option control the number of "steps" or the granularity of the zoom:
     - Default: `20`.

2. Fixed an issue where zooming wasn't accurate at certain browser widths:
   - Caused by scroll container widths on the right and left (i.e., widths that didn’t include the waveform was not being taken into account.)

3. Fixed a minor issue where zooming caused the user to "jump" out of the waveform when zooming very fast:
   - Likely occurred due to JavaScript load during rapid scroll events.
   - Fixed by setting `passive: false` on the global `wheel` event listener to prevent the browser’s default scrolling behavior.

4. Resolved floating-point inaccuracies in `pointerTime` by memorizing the previous mouse X position:
   - This prevents waveform drift during fixed zooming by recalculating only when the pointer moves.

### **Other Changes**:
- If `maxZoom` is not set, it defaults to the `clientWidth` of the container.

## Summary by CodeRabbit

- **New Features**
	- Introduced exponential zooming options for enhanced zoom control.
	- Added the ability to specify the number of iterations for reaching maximum zoom level.

- **Improvements**
	- Enhanced user experience by preventing native scrolling during zoom operations.
	- Adjusted zoom calculations to minimize drifting during fixed zooming.

### DIFF
--- a/src/plugins/zoom.ts
+++ b/src/plugins/zoom.ts
@@ -6,6 +6,7 @@
  * @author HoodyHuo (https://github.com/HoodyHuo)
  * @author Chris Morbitzer (https://github.com/cmorbitzer)
  * @author Sam Hulick (https://github.com/ffxsam)
+ * @autor Gustav Sollenius (https://github.com/gustavsollenius)
  *
  * @example
  * // ... initialising wavesurfer with the plugin
@@ -36,10 +37,28 @@ export type ZoomPluginOptions = {
    * @default 5
    */
   deltaThreshold?: number
+  /**
+   * Whether to zoom into the waveform using a consistent exponential factor instead of a linear scale.
+   * Exponential zooming ensures the zoom steps feel uniform regardless of scale.
+   * When disabled, the zooming is linear and influenced by the `scale` parameter.
+   *
+   * @default false
+   */
+  exponentialZooming?: boolean
+  /**
+   * Number of steps required to zoom from the initial zoom level to `maxZoom`.
+   *
+   * @default 20
+   */
+  iterations?: number
+
+
 }
 const defaultOptions = {
   scale: 0.5,
   deltaThreshold: 5,
+  exponentialZooming: false,
+  iterations: 20,
 }
 
 export type ZoomPluginEvents = BasePluginEvents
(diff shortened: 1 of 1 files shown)


src/plugins/zoom.ts
```diff
@@ -66,6 +93,15 @@ class ZoomPlugin extends BasePlugin<ZoomPluginEvents, ZoomPluginOptions> {
     }
     this.container = this.wrapper.parentElement as HTMLElement
     this.wrapper.addEventListener('wheel', this.onWheel)
+
+    // Prevent native scrolling obstructing the zoom on this waveform
+    document.addEventListener('wheel', this.preventScroll, { passive: false });
+
+    if(typeof this.options.maxZoom === 'undefined'){
+      this.options.maxZoom = this.container.clientWidth
+      console.log("maxZoom", this.options.maxZoom)
```
katspaugh, line 102: Please remove the log and also please run `yarn prettier`.
The code looks good otherwise, cheers!

src/plugins/zoom.ts
```diff
@@ -66,6 +93,15 @@ class ZoomPlugin extends BasePlugin<ZoomPluginEvents, ZoomPluginOptions> {
     }
     this.container = this.wrapper.parentElement as HTMLElement
     this.wrapper.addEventListener('wheel', this.onWheel)
+
+    // Prevent native scrolling obstructing the zoom on this waveform
+    document.addEventListener('wheel', this.preventScroll, { passive: false });
```
katspaugh, line 98: Just noticed this is assigned to the entire document element. Will it essentially prevent the page from scrolling whenever wavesurfer with Zoom is on the page? Can it be scoped to only the wavesurfer container?

gustavsollenius: Yes your right it should be attached to a different container. Otherwise, as you mentioned, it prevents scrolling for the entire page. I'll take another look at this.

src/plugins/zoom.ts
gustavsollenius: Ok!

src/plugins/zoom.ts
gustavsollenius, line 1: The issue with losing "focus" on the waveform when zooming very fast was resolved by attaching the listener to the scroll container instead of the wrapper container. This was challenging to troubleshoot because the issue only occurred in Mozilla Firefox and not in Google Chrome.

katspaugh (approved):
👍

PR #3941, opened 2024-11-26
Feat(Timeline): Add timeOffset in options so that the scale can be aligned with data

dinther:
I am using the timeline to display BPM (Beat's per minute) for a given sound file. But sound files don't always start exactly with beat 0 at 0 ms. options.timeOffset allows you to shift the labels up or down the timeline and line them up with the waveform.

I normally don't do git. In fact I just lost two days worth of work while fuckling around with git shit. but here you are.

I like your library though, it's easy to understand.

## Summary by CodeRabbit

- **New Features**
	- Introduced an optional `timeOffset` property for the timeline plugin, allowing users to customize the alignment of numeric labels.
- **Enhancements**
	- Improved timeline functionality by enabling adjustable offsets for better display alignment with other visual elements.

### DIFF
--- a/src/plugins/timeline.ts
+++ b/src/plugins/timeline.ts
@@ -24,6 +24,8 @@ export type TimelinePluginOptions = {
   primaryLabelSpacing?: number
   /** Interval between secondary numeric labels  in timeIntervals (i.e notch count) */
   secondaryLabelSpacing?: number
+  /** offset in seconds for the numeric labels */
+  timeOffset?: number  
   /** Custom inline style to apply to the container */
   style?: Partial<CSSStyleDeclaration> | string
   /** Turn the time into a suitable label for the time. */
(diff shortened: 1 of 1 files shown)


src/plugins/timeline.ts
```diff
@@ -143,6 +145,11 @@ class TimelinePlugin extends BasePlugin<TimelinePluginEvents, TimelinePluginOpti
     return 2
   }
 
+    // Return how many seconds the labels are offset.
+    private defaultTimeOffset(pxPerSec: number): number {
+      return 0;
+    }
```
katspaugh, line 151: This doesn't have to be a function, does it? It can be just a static value in [defaultOptions](https://github.com/katspaugh/wavesurfer.js/blob/2867c6d633bb275ed4d027a517ced0d192828652/src/plugins/timeline.ts#L37-L50).

katspaugh (approved):
Thank you!

PR #3945, opened 2024-11-29
Fix: region-clicked event and seek issue inside region (#3911, #3804)

### LINKED ISSUE #3804: [Regions] Clicking inside region does not seek
Since upgrading from 7.7.11 to 7.8.2, seeking no longer works when clicking inside a region.  Clicking outside a region works fine.  I'm seeing this behavior in both Chrome and Edge.  On Firefox, clicking inside a region seeks the audio to that spot, as expected.

Prior to the upgrade, the audio would seek to the correct spot when clicking inside a region on all browsers.

### LINKED ISSUE #3911: region-clicked Event Issue on Regions in WaveSurfer Zoomed State for Chrome and Safari
## Bug description

- Load https://wavesurfer.xyz/examples/?regions.js
- Use the zoom slider to zoom in until scrolling is enabled.
- Click on a region to start playback: OK.
- Click on the region again:
  -  Firefox (on macOS, Ubuntu, and Windows 11): Playback starts from the beginning of the region, and the region color changes: OK.
  -  Chrome (on macOS, Ubuntu, and Windows 11) and Safari (on macOS): The region-clicked event does not fire: NG.

## Environment
 - Browser: please see above

## Minimal code snippet

please refer to the https://wavesurfer.xyz/examples/?regions.js

```js
...
regions.on('region-clicked', (region, e) => {
    e.stopPropagation() // prevent triggering a click on the waveform
    activeRegion = region
    region.play()
    region.setOptions({ color: randomColor() })
  })
...
```
## Expected result

Please refer to the above.

## Obtained result

Please refer to the above.

## Screenshots

ubuntu: chrome: 

The region-clicked event may does not fire.

Olejqua:
## Short description
Added parentElement check in virtualAppend method to prevent redundant DOM operations that were breaking click event handling during zoomed playback.

Resolves #3911
Resolves #3804

This PR resolves two related issues with region-clicked events and seeking inside regions in WaveSurfer.js:

1. **Issue #3911:**  
   The `region-clicked` event was not firing on Chrome and Safari when the zoom slider was used. This behavior was inconsistent across browsers, as Firefox worked as expected.

2. **Issue #3804:**  
   Since version 7.8.2, clicking inside a region no longer seeks the audio position correctly on Chrome and Edge. The expected behavior (as seen in Firefox) is that clicking inside a region seeks the audio to the clicked position.
  

## Implementation details
Modified regions.ts virtualAppend method to check element's DOM state before append/remove operations:
```typescript
if (isVisible && !element.parentElement) {
    container.appendChild(element)
} else if (!isVisible && element.parentElement) {
    element.remove()
}
```

### DIFF
--- a/src/plugins/regions.ts
+++ b/src/plugins/regions.ts
@@ -527,9 +527,9 @@ class RegionsPlugin extends BasePlugin<RegionsPluginEvents, RegionsPluginOptions
       // Check if the region is between the scrollLeft and scrollLeft + clientWidth
       const isVisible = start + width > scrollLeft && start < scrollLeft + clientWidth
 
-      if (isVisible) {
+      if (isVisible && !element.parentElement) {
         container.appendChild(element)
-      } else {
+      } else if (!isVisible && element.parentElement) {
         element.remove()
       }
     }


katspaugh (approved):
Thanks!

PR #3959, opened 2024-12-13
Fix: (Renderer) splitChannels memory leak

rjett1:
## Short description
When using wavesurfer's `splitChannels` option there is a memory leak that occurs when zooming the waveforms. The code does not properly clean up scroll event handlers which can eventually lead to the page crashing.

## Implementation details
The original code stores an unsubscribe function for a scroll event handler that gets created when an audio channel is rendered. When multiple channels are rendered the previous unsubscribe functions get overwritten and their references are never properly cleaned up.

To fix the issue I've updated the unsubscribeOnScroll variable to be an array, so that it can store the unsubscribe function for each audio channel that gets rendered.

## How to test it
You can observe this behavior on the Zoom Plugin example page if you set the `splitChannels` option to `true`. As you zoom in and out the page will use more and more memory.

## Screenshots
**Before:**
before

**After:**
after

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Enhanced management of scroll event subscriptions, allowing multiple unsubscribe callbacks.
  
- **Bug Fixes**
	- Improved reliability of the unsubscribe mechanism for scroll events.

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -31,7 +31,7 @@ class Renderer extends EventEmitter<RendererEvents> {
   private lastContainerWidth = 0
   private isDragging = false
   private subscriptions: (() => void)[] = []
-  private unsubscribeOnScroll?: () => void
+  private unsubscribeOnScroll: (() => void)[] = []
 
   constructor(options: WaveSurferOptions, audioElement?: HTMLElement) {
     super()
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
Thank you, great catch!

PR #3965, opened 2024-12-20
Adding options to allow/disallow resizing the start or end of a region

ibash:
## Short description
This optionally allows disabling resizing one end of a region. My use case is to allow extending a region in only one direction. This doesn't change any of the styling because it's easy enough to override the styling for a handle with css.

## Implementation details
Just a few minor options on the region plugin. Similar to the resize option, we just check the setting in the onResize handler.

## How to test it
Set `resizeStart` or `resizeEnd` to false and check that you can't change the value of the start/end respectively.

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Introduced new properties for controlling the resizing of regions: `resizeStart` and `resizeEnd`.
	- Enhanced flexibility in region manipulation by allowing dynamic modification of resizing options.
  
- **Bug Fixes**
	- Improved checks during resizing operations to respect new flags, preventing unwanted resizing.

### DIFF
--- a/src/plugins/regions.ts
+++ b/src/plugins/regions.ts
@@ -60,6 +60,10 @@ export type RegionParams = {
   drag?: boolean
   /** Allow/dissallow resizing the region */
   resize?: boolean
+  /** Allow/dissallow resizing the start of the region */
+  resizeStart?: boolean
+  /** Allow/dissallow resizing the end of the region */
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
Nice, thank you!

PR #3982, opened 2025-01-01
Allow user to update peaks and/or duration with setOptions

botoxparty:
### Current Behavior
When syncing WaveSurfer with external audio/time sources (like Tone.js transport), calling `setTime()` triggers the 'seeking' event because it updates both the media element's time and the visual progress. This will allow seeking and interact events to continue to work when syncing with an external source.

### Problem
In scenarios where WaveSurfer is being used as a visual representation of audio that's controlled externally, we don't always want to trigger the media element's seeking behavior. Currently, there's no clean way to update just the visual progress without triggering seeking events.

### Proposed Solution
Add a new public method to update only the visual progress

Let me know your thoughts, maybe you have a better solution with a your understanding of the lib.

## Summary by CodeRabbit

- **New Features**
	- Enhanced waveform progress tracking with a new method to update visual progress more precisely during audio playback.

- **Improvements**
	- Improved handling of audio data options, allowing for better management of duration and peaks during playback. 
	- Improved synchronization between audio playback and waveform visualization. 
	- Added conditional checks for managing duration and peaks in audio options.

### DIFF
--- a/src/wavesurfer.ts
+++ b/src/wavesurfer.ts
@@ -353,6 +353,16 @@ class WaveSurfer extends Player<WaveSurferEvents> {
   /** Set new wavesurfer options and re-render it */
   public setOptions(options: Partial<WaveSurferOptions>) {
     this.options = Object.assign({}, this.options, options)
+    if (options.duration && !options.peaks) {
+      this.decodedData = Decoder.createBuffer(this.exportPeaks(), options.duration)
+    }
+    if (options.peaks && options.duration) {
+      // Create new decoded data buffer from peaks and duration
+      this.decodedData = Decoder.createBuffer(
+        options.peaks,
+        options.duration
+      );
+    }
     this.renderer.setOptions(this.options)
 
     if (options.audioRate) {


src/wavesurfer.ts
```diff
@@ -540,6 +540,11 @@ class WaveSurfer extends Player<WaveSurferEvents> {
     this.options.interact = isInteractive
   }
 
+  /** Update the visual progress */
+  public updateVisualProgress(time = this.getCurrentTime(), isPlaying = this.isPlaying()) {
```
katspaugh, line 544: I'm reluctant to add a low-level method like this to the public API, especially since it's targeting quite an edge case.

You can easily render wavesurfer w/o playback by providing peaks and duration but no audio. See this example: 
https://wavesurfer.xyz/examples/?0f80252d71c2f4a1ead0a572e1cefad3

It will still respond to `setTime` or `seekTo` but no actual playback will be happening, so you can visually sync it to an external player.

botoxparty:
So i've updated this PR, let me know if you think this is something that should be part of the libraries functionality.

Rather than destroying and initialising the wavesurfer all over again to update the duration (my application has a time stretch feature where the playback speed can be changed, therefore the duration of the same waveform can change. e.g. Warp in Ableton)

katspaugh (approved):
That's a good solution! Thanks!

katspaugh:
Released in 7.8.15.

PR #3995, opened 2025-01-13
feat: override blob MIME type

wfk007:
## Short description

Add a new option(mimeType) to override the blob MIME when the server doesn't send the correct MIME in content-type.

discussion: https://github.com/katspaugh/wavesurfer.js/discussions/3621
related pr: https://github.com/katspaugh/wavesurfer.js/pull/3471

## Implementation details

After fetchBlob in loadAudio, if there there is mimeType option, override the Blob MIME.

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Enhanced audio loading configuration by adding optional MIME type specification for audio files.

### DIFF
--- a/src/wavesurfer.ts
+++ b/src/wavesurfer.ts
@@ -79,7 +79,7 @@ export type WaveSurferOptions = {
   /** Nonce for CSP if necessary */
   cspNonce?: string
   /** Override the Blob MIME type */
-  mimeType?: string
+  blobMimeType?: string
 }
 
 const defaultOptions = {
@@ -437,9 +437,9 @@ class WaveSurfer extends Player<WaveSurferEvents> {
       }
       const onProgress = (percentage: number) => this.emit('loading', percentage)
       blob = await Fetcher.fetchBlob(url, onProgress, fetchParams)
-      const overridedMimeType = this.options.mimeType
-      if (overridedMimeType) {
-        blob = new Blob([blob], { type: overridedMimeType })
+      const overridenMimeType = this.options.blobMimeType
+      if (overridenMimeType) {
+        blob = new Blob([blob], { type: overridenMimeType })
       }
     }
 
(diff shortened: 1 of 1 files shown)


src/wavesurfer.ts
```diff
@@ -78,6 +78,8 @@ export type WaveSurferOptions = {
   backend?: 'WebAudio' | 'MediaElement'
   /** Nonce for CSP if necessary */
   cspNonce?: string
+  /** Override the Blob MIME type */
+  mimeType?: string
```
katspaugh, line 82: Since it's only for Blobs, should we call it `blobMimeType`?

src/wavesurfer.ts
```diff
@@ -435,6 +437,10 @@ class WaveSurfer extends Player<WaveSurferEvents> {
       }
       const onProgress = (percentage: number) => this.emit('loading', percentage)
       blob = await Fetcher.fetchBlob(url, onProgress, fetchParams)
+      const overridedMimeType = this.options.mimeType
+      if (overridedMimeType) {
+        blob = new Blob([blob], { type: overridedMimeType })
+      }
```
katspaugh, line 443:
```suggestion
      if (overridenMimeType) {
        blob = new Blob([blob], { type: overridenMimeType })
      }
```

wfk007:
> Tbh I'm not sure wavesurfer should be handling this. As mentioned in the linked discussion, your server should send the right content-type header. Alternatively, you could fetch and set the mime in your own code, then call `loadBlob`.

Yes, fetch the blob and call `loadBlob` in my code can solve this issue. But We need to implement the progress and abortController logic again. Not all backend services will follow the specification and return the correct MIME in content-type, and the adding of this configuration can also prevent other users from encountering the above breaking changes during the upgrade with the minimal changes.

katspaugh (approved):

katspaugh:
Thank you! Will release it now.

PR #4012, opened 2025-01-27
Updated player.ts to not pause audio for external media on destroy()

Stianlars1:
## Short description
In today's solution the playback audio will pause on calling destroy() for instances using external media. 
We probably dont need this for the external media's.

This is a PR which stems from this [discussion](https://github.com/katspaugh/wavesurfer.js/discussions/3190)

## Implementation details
Moved the called pause-function after the early return for external media :
```jsx
// previously this was called before the below early return
// this.media.pause()
if (this.isExternalMedia) return
this.media.pause() // now called after
```

## How to test it
Create a audio element thats shared globally between n + 1 WaveSurfer instances, then try playing the audio for one of the instances and destroy one of the wavesurfers. The audio will be paused.  

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved media destruction logic to handle external media sources more accurately.
	- Refined media pause behavior during player destruction.

### DIFF
--- a/src/player.ts
+++ b/src/player.ts
@@ -86,9 +86,7 @@ class Player<T extends GeneralEventTypes> extends EventEmitter<T> {
   }
 
   protected destroy() {
-    
     if (this.isExternalMedia) return
-    
     this.media.pause()
     this.media.remove()
     this.revokeSrc()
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
Awesome, thank you!

katspaugh:
It's a potentially breaking change for some use cases so I've released it as a new major version (7.9.0).

PR #4017, opened 2025-02-07
Allow aborting loading audio

ibash:
## Short description
This change makes loadAudio optionally take an AbortSignal which can be used to abort loading.
We use this to prevent a race condition where the wrong audio might be loading into wavesurfer.

## Implementation details
Passing through an AbortSignal. After any async call it needs to be checked with `throwIfAborted`. Not my favorite, but is effective.

## How to test it
Easiest is to add some delays in `loadAudio` and then try aborting.

Add manual delays throughout loadAudio with:
```
await new Promise((r) => setTimeout(r, 5000))
```

Then call loadAudio passing in a signal and aborting it.

## Checklist

## Summary by CodeRabbit

- **New Features**
	- Enhanced the audio playback experience by enabling the ability to cancel ongoing audio loading operations.
	- Improved error handling ensures that if an audio load is interrupted, the system responds more promptly and smoothly.

### DIFF
--- a/src/wavesurfer.ts
+++ b/src/wavesurfer.ts
@@ -418,7 +418,7 @@ class WaveSurfer extends Player<WaveSurferEvents> {
     return this.plugins
   }
 
-  private async loadAudio(url: string, blob?: Blob, channelData?: WaveSurferOptions['peaks'], duration?: number) {
+  private async loadAudio(url: string, blob?: Blob, channelData?: WaveSurferOptions['peaks'], duration?: number, signal?: AbortSignal) {
     this.emit('load', url)
 
     if (!this.options.media && this.isPlaying()) this.pause()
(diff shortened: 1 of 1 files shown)


katspaugh:
You can already pass an abort signal in the fetchParams option. See https://wavesurfer.xyz/docs/types/wavesurfer.WaveSurferOptions

Or is it not sufficient for your use case?

ibash:
We're using a singleton wavesurfer instance and loading different audio into it. When the second call to `loadAudio` happens we need to make sure no side effects from the first call get propagated out. Calling `throwIfAborted` after each async hop is the most fool-proof way of doing that. Though it's definitely possible that we could do a work around by unsubscribing / resubscribing any event listeners.

katspaugh:
Ah, so you aren't even interested in aborting the actual fetch or decoding process, only in interrupting the whole thing in between those async processes as if they threw?

src/wavesurfer.ts
```diff
@@ -434,6 +434,8 @@ class WaveSurfer extends Player<WaveSurferEvents> {
       }
       const onProgress = (percentage: number) => this.emit('loading', percentage)
       blob = await Fetcher.fetchBlob(url, onProgress, fetchParams)
```
katspaugh, line 436: Since there's a bit of an overlap with the existing fetchParams abortSignal, I'm thinking maybe we can use the new one for this as well?

Something like:

```suggestion
      blob = await Fetcher.fetchBlob(url, onProgress, {
        ...fetchParams,
        signal: signal ?? fetchParams?.signal
      })
```

Then if you abort it, it will also abort the fetching.

katspaugh:
Closing but please feel free to reopen it if you decide to follow up on my comment.

PR #4061, opened 2025-03-15
Updating an event listener in the example docs

shahil-yadav:
Resolves: Updates an event listener in the example documentation

I think `redraw` event listener should be the one that needs to triggered when visible waveform is drawn onto screen, isn't it ?

## Checklist
image

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated the audio waveform update mechanism to ensure more accurate and timely visual feedback for a smoother experience.


katspaugh (approved):
Thanks! 👍

PR #4066, opened 2025-03-20
Fix: play(start, stop) with WebAudio backend not worked

OwLu0905:
## Short description
Using `backend: 'WebAudio'` with the play(start, stop) feature doesn't work properly.
This PR fixes the issue while ensuring that the default `backend: 'MediaElement'` option continues to work correctly.

Related : #4014 

## Implementation details
- Move `media.stopAt()` call after `super.play()` to ensure `bufferNode` is initialized
- Ensure event listener for `ended` event can be properly registered

The problem occurs in the `stopAt` method where `bufferNode` is not yet initialized when called:
```javascript
stopAt(timeSeconds: number) {
  const delay = timeSeconds - this.currentTime
  this.bufferNode?.stop(this.audioContext.currentTime + delay)

  // The event listener won't be registered if bufferNode is not initialized yet
  this.bufferNode?.addEventListener(
    'ended',
    () => {
      this.pause()
    },
    { once: true },
  )
}
```

## How to test it
- Play audio with both start and end times specified using the WebAudio backend

## Checklist
* [ ] This PR is covered by e2e tests
* [ ] It introduces no breaking API changes

## Summary by CodeRabbit

- **Refactor**
  - Enhanced asynchronous handling in playback operations to ensure outcomes are reliably captured before proceeding. This update contributes to a smoother and more consistent media experience for users, particularly during playback initiation. It also sets the stage for further improvements in performance and stability for future updates.

PR #4067, opened 2025-03-21
Fix: prevent race condition in stopAt method

OwLu0905:
## Short description
The current implementation of `stopAt` in WebAudioPlayer has a race condition when the method is called multiple times in quick succession. Previously, this would cause the player to enter an incorrect state because event listeners from previous calls would still execute `this.pause()` even after a new playback had started.

Related: #4066 

## Implementation details
- Modified the `stopAt` method to capture the current buffer node in a local variable
- Added a conditional check within the `'ended'` event listener to ensure `this.pause()` is only called if the buffer node that triggered the event is still the active one

## How to test it
- Play audio with both start and end times specified using `backend: 'WebAudio'` option
- Click region multiple times in quick succession

## Screenshots
This video shows how to trigger the bug.
https://github.com/user-attachments/assets/64ab94c7-56af-462d-b556-f116ec25ec82

## Checklist
image

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced the reliability of audio playback controls, ensuring that stopping audio functions consistently even during rapid state changes.

### DIFF
--- a/src/webaudio.ts
+++ b/src/webaudio.ts
@@ -134,13 +134,16 @@ class WebAudioPlayer extends EventEmitter<WebAudioPlayerEvents> {
 
   stopAt(timeSeconds: number) {
     const delay = timeSeconds - this.currentTime
-    this.bufferNode?.stop(this.audioContext.currentTime + delay)
+    const currentBufferNode = this.bufferNode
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
Good catch, thanks so much!

PR #4082, opened 2025-04-26
getContent method and region-content-changed event created

adahox:
## Short description
Resolves #

## Implementation details
 -  implemented `getContent(asHTML:bool)`. it bring possibility to get content as HTML or just String value.
 - `region-content-changed`  RegionPlugin event:  It fires always `setContent(value)` is called.

## How to test it
 - Create a Region
 - Implement `region-content-changed` event on `RegionPlugin`.
 - using `region.getContent()` should  bring content value as string.
 - passing asHTML param  as true to `region.getContent()` method, it should return HTMLElement string.
 - using `region.setContent()` should fires `region-content-changed` event.
 - 
## Checklist

### DIFF
--- a/src/plugins/regions.ts
+++ b/src/plugins/regions.ts
@@ -28,6 +28,8 @@ export type RegionsPluginEvents = BasePluginEvents & {
   'region-in': [region: Region]
   /** When playback leaves a region */
   'region-out': [region: Region]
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
Thanks!

PR #4085, opened 2025-05-05
Feat(Regions): Add a `region-initialized` event that triggers early

DefV:
## Short description
Right now `region-created`` only triggers when the
user lifts their moust, which means it's not possible
to get any callbacks or events of their selection until
that moment. This new event triggers immediatly, allowing
developers to hook onto events, or do things like change
the color, add a number inside etc..

## Implementation details
I created a new event so existing code doesn't break.

## Checklist

### DIFF
--- a/src/plugins/regions.ts
+++ b/src/plugins/regions.ts
@@ -12,7 +12,7 @@ import createElement from '../dom.js'
 export type RegionsPluginOptions = undefined
 
 export type RegionsPluginEvents = BasePluginEvents & {
-  /** When a new region is initialized */
+  /** When a new region is initialized but not rendered yet */
   'region-initialized': [region: Region]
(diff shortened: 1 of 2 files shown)


src/plugins/regions.ts
```diff
@@ -12,6 +12,8 @@ import createElement from '../dom.js'
 export type RegionsPluginOptions = undefined
 
 export type RegionsPluginEvents = BasePluginEvents & {
+  /** When a new region is initialized */
```
katspaugh, line 15:
```suggestion
  /** When a new region is initialized but not rendered yet */
```

katspaugh:
Thanks! Looks good, I left one comment.

katspaugh (approved):
Thank you!

PR #4089, opened 2025-05-09
Feat(hover): implement option to keep the label to the left of the cursor

### LINKED ISSUE #2145: Cursor plugin: option to flip time element by default
Since #1776, the time element is flipped to the left of the cursor when close to the right edge of the waveform element. I'd like to have the option to reverse this behavior (always keep the time element to the left of the cursor unless close to the left edge of the waveform element). I've been using this patch for now which works fine for me, but a togglable option would be useful:

```patch
diff --git a/node_modules/wavesurfer.js/dist/plugin/wavesurfer.cursor.js b/node_modules/wavesurfer.js/dist/plugin/wavesurfer.cursor.js
index 2c9af8f..609d5a0 100644
--- a/node_modules/wavesurfer.js/dist/plugin/wavesurfer.cursor.js
+++ b/node_modules/wavesurfer.js/dist/plugin/wavesurfer.cursor.js
@@ -229,7 +229,7 @@ var CursorPlugin = /*#__PURE__*/function () {
       var y = 0;
       var x = e.clientX - bbox.left;
 
-      var flip = bbox.right < e.clientX + _this.outerWidth(_this.displayTime);
+      var flip = bbox.left < e.clientX - _this.outerWidth(_this.displayTime);
 
       if (_this.params.showTime && _this.params.followCursorY) {
         // follow y-position of the mouse
```

dpeukert:
## Short description
This PR adds an option to the hover plugin to keep the label to the left of the cursor. Resolves #2145 (which seems to be already closed). I've also added descriptions to the other hover plugin options and fixed a typo.

## Implementation details
A new `labelPreferLeft` boolean option (defaults to false) has been added. When enabled, the label prefers to be on the left side, unless near the left edge.

## How to test it
Via the added cypress test.

## Screenshots
false:
image

true:
image

## Checklist


katspaugh:
Released it in 7.9.7 🙏

PR #4113, opened 2025-06-21
fix(Record): clear media element source when no URL is provided

### LINKED ISSUE #4039: Recorder plugin in Firefox generates continuoues errors
## Bug description

When using the record plugin in firefox a continuous stream of erros occur in the developer tools console. This slows down drawing, making it lumpy. I expect the impact is minimal if dev tools are closed

`Invalid URI. Load of media resource  failed.`

This error is not seen in the examples webpages as the iframe seems to lose the error. I converted the record.js  example into a html page and served that from the waversurfer source and it is reproducable.

I can't find much info but it seems at some point Firefox devs decided to generate errors for setting invalid player media src values. the wavesurfer record plug triggers this by setting the `player.js` audio elements src property to an empty string which is invalid.

The code is in `plugins/record.js` [renderMicStream -> drawWaveform]((https://github.com/katspaugh/wavesurfer.js/blob/1df2bd924fdec4f86263de4ae0fa2c6bade9a289/src/plugins/record.ts#L174))

``` javascript
         this.wavesurfer
          .load(
            '',
            [this.dataWindow],
            this.options.scrollingWaveform ? this.options.scrollingWaveformWindow : totalDuration,
          )
```

To work around I don't set `media.src` in player.js if `newsrc` is falsey. ie

```javascript
    if (newSrc) {
      try {
        this.media.src = newSrc
      } catch (e) {
        this.media.src = url
      }
    }
```

But the error will still occur if media.src was already set and so is reset to ''.

This daft behaviour by Firefox as getSrc() returns '' but setSrc of '' logs an error.
 
## Environment
 - Browser: Firefox - no prob in Chrome or Edge (on Windows)

## Minimal code snippet

The Record.js example converted to a HTML page and using `npx server .`

## Expected result

No stream of errors

## Obtained result

Stream of errors

matemolnar8:
## Short description
Resolves #4039

Fixes the continuous warning logging on Firefox 

## Implementation details
Instead of setting `src` to `''` when loading the data from record plugin to wavesurfer, the `src` attribute is removed. 
I also updated other places where `src` was set to `''` as the media element interprets it as relative path.

I left the `''` URL being passed to `load`, but maybe passing `null` instead would be safer. Let me know if I should change that.

## How to test it
Start recording with record plugin on Firefox on the record example page

## Screenshots
Screenshot 2025-06-21 at 20 44 23
Before the fix:
Screenshot 2025-06-21 at 20 45 07

## Checklist

### DIFF
--- a/src/player.ts
+++ b/src/player.ts
@@ -75,7 +75,7 @@ class Player<T extends GeneralEventTypes> extends EventEmitter<T> {
 
     // Reset the media element, otherwise it keeps the previous source
     if (src) {
-      this.media.src = ''
+      this.media.removeAttribute('src')
     }
 
     try {
@@ -90,7 +90,7 @@ class Player<T extends GeneralEventTypes> extends EventEmitter<T> {
     this.media.pause()
     this.media.remove()
     this.revokeSrc()
-    this.media.src = ''
+    this.media.removeAttribute('src')
     // Load resets the media element to its initial state
     this.media.load()
   }
--- a/src/wavesurfer.ts
+++ b/src/wavesurfer.ts
@@ -451,8 +451,13 @@ class WaveSurfer extends Player<WaveSurferEvents> {
       }
     }
 
-    // Set the mediaelement source
-    this.setSrc(url, blob)
+    if (url == '') {
+      // If no URL is provided, clear the mediaelement source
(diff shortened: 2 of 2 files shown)


katspaugh (approved):
Thank you!

PR #4125, opened 2025-06-25
Optimize the spectrogram and introduce WindowedSpectrogramPlugin

LezdCS:
## Implementation details
- Ensure the removal of a plugin by creating the "unregisterPlugin" function. The current system using the .destroy() is not working correctly if you want to regsiter the plugin again after destroying it.

- Refactor the Spectrogram plugin to work on long audio files and support the zoom without FPS drops. (Still struggle with big FTT samples and still crash on too big ones).
- Create a plugin "WindowedSpectrogramPlugin" that render the spectrogram based on the window size. (Works well with big FTT samples even tho the initial loading time takes a bit of time the higher you go).

=> Basically for the 2 plugins it fixes the issue that after 32000 pixels (depends on the browser) the canvas was not rendering anymore. It also improves considerably the fps when dealing with a spectrogram while zoomed a lot.

## Screenshots
My videos example are too heavy to be added here but you can try on your local by combining Zoom and Spectrogram.

## Checklist

### DIFF
--- a/src/base-plugin.ts
+++ b/src/base-plugin.ts
@@ -12,6 +12,7 @@ export class BasePlugin<EventTypes extends BasePluginEvents, Options> extends Ev
   protected wavesurfer?: WaveSurfer
   protected subscriptions: (() => void)[] = []
   protected options: Options
+  private isDestroyed = false
 
   /** Create a plugin instance */
   constructor(options: Options) {
@@ -26,6 +27,12 @@ export class BasePlugin<EventTypes extends BasePluginEvents, Options> extends Ev
 
   /** Do not call directly, only called by WavesSurfer internally */
   public _init(wavesurfer: WaveSurfer) {
+    // Reset state if plugin was previously destroyed
+    if (this.isDestroyed) {
+      this.subscriptions = []
+      this.isDestroyed = false
+    }
+    
     this.wavesurfer = wavesurfer
     this.onInit()
   }
@@ -34,6 +41,9 @@ export class BasePlugin<EventTypes extends BasePluginEvents, Options> extends Ev
   public destroy() {
     this.emit('destroy')
     this.subscriptions.forEach((unsubscribe) => unsubscribe())
+    this.subscriptions = []
+    this.isDestroyed = true
+    this.wavesurfer = undefined
   }
 }
 
--- a/src/plugins/spectrogram-windowed.ts
+++ b/src/plugins/spectrogram-windowed.ts
@@ -1,0 +1,886 @@
+/**
+ * Windowed Spectrogram plugin - Optimized for very long audio files
+ * 
+ * Only renders frequency data in a sliding window around the current viewport,
+ * keeping memory usage constant regardless of audio length.
+ */
+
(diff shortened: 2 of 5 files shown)


katspaugh:
This looks cool, thank you! Could you also add an example js for the new plugin?

LezdCS:
> This looks cool, thank you! Could you also add an example js for the new plugin?

I created the example! 

The initial load is quite heavy so I am currently thinking of a way of fixing that. For example on a music of 9 minutes with fftSamples at 128 it takes 6 seconds but it gets to 15 seconds with fft samples at 512...  I realized the loading according to the scroll is not working properly, I am working on a fix right now.

I also found the delimitation between the rendered segments a bit too abrupt and laking of precision so I will work on that. 
Screenshot 2025-06-26 at 0 26 02

LezdCS:
@katspaugh last commit is fixing the loading of the spectrogram by rendering the segments only if they are visibile and keeps in memory the ones already loaded. I also added the possiblity to use a web worker so we don't block the main thread when we calculate the frequencies for the segments. We also have a progressive loading option that will by itself load the segments one by one without the user moving, it is quite fast and seemless if paired with the web worker.

I need to cleanup the code and run more tests on my side before confirming that the plugin is stable but I believe you can already look at the current example and compare the WindowedSpectrogram with the normal spectrogram plugin (the one in the 7.9.8).

katspaugh:
@LezdCS awesome, love it!

LezdCS:
Hi @katspaugh , I have something pretty stable on my branch https://github.com/LezdCS/wavesurfer.js/tree/rs-fft that uses WASM for heavy calculations (FFT) but I wanted to know if you prefere keeping your repository clean from Rust code and WebAssembly ? In thise case I will only send the JavaScript optimization I did to the fix-spectrogram branch! For now I didn't succeed in combining WASM and web workers but when I use WASM for calculation the speed increase by 3 times which is pretty considerable.

katspaugh:
@LezdCS 3x is impressive! I'd prefer if we keep only the JS version here but if you create a separate repo, we can link to it from the readme and the plugin docs.

LezdCS:
> @LezdCS 3x is impressive! I'd prefer if we keep only the JS version here but if you create a separate repo, we can link to it from the readme and the plugin docs.

I just pushed the refactored web worker to use the "rollup-plugin-web-worker-loader" plugin so now we have a real .ts file for the worker. 
I believe both Spectrogram and SpectrogramWindowed are now stable. 
I would say that for the Spectrogram plugin the loading perfomance have not changed but the usage of the web worker prevent the UI to freeze and we can zoom in and out smoothly.
For SpectrogramWindowed I am pretty happy with the result, but there is room for future improvement.

For the WASM repository I will do that in the upcoming weeks as I am really busy on other projects.

katspaugh (approved):

katspaugh:
@LezdCS many thanks! And sorry for the delayed response. I've released it in 7.10.0. 🎉

pengzhendong:
Awesome jobs.

PR #4143, opened 2025-07-05
[Fix] [Timeline] Make sure that first tick on left edge is drawing

jakerdy:
## Short description

Currently, when a timeline is displayed, the first major tick is not rendered. This occurs due to an edge case where the time value is zero, and the plugin's rendering condition uses a strict inequality (`>`) instead of a non-strict one (`>=`).

## Implementation details

Changed the comparison operator from `>` to `>=` to ensure the first major tick at time zero is rendered.

## How to test it

This issue is demonstrable in the existing examples. Please refer to them for testing.

## Screenshots

Observe the lower-left corner of the timeline.

Before:
image

After:
image

## Checklist
* [+] This PR is covered by e2e tests
* [+] It introduces no breaking API changes


katspaugh (approved):
Great catch, thank you!

PR #4160, opened 2025-07-18
fix(Renderer): fix cursor jump at end of playback

Tsuk1ko:
## Short description

Fix cursor jump at end of playback.

## Implementation details

This issue was caused by sudden translateX changes.

I fixed it by making the translateX smoothly increase based on the progress value.

## How to test it

Setting `cursorWidth` to a larger value makes it more visible.

## Screenshots

Before:

chrome_u3lTrAfiKJ

After:

chrome_62cGQEMQir

## Checklist

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -766,7 +766,9 @@ class Renderer extends EventEmitter<RendererEvents> {
     this.canvasWrapper.style.clipPath = `polygon(${percents}% 0%, 100% 0%, 100% 100%, ${percents}% 100%)`
     this.progressWrapper.style.width = `${percents}%`
     this.cursor.style.left = `${percents}%`
-    this.cursor.style.transform = `translateX(-${Math.round(percents) === 100 ? this.options.cursorWidth : 0}px)`
+    this.cursor.style.transform = this.options.cursorWidth
+      ? `translateX(-${progress * this.options.cursorWidth}px)`
+      : ''
 
     if (this.isScrollable && this.options.autoScroll) {
       this.scrollIntoView(progress, isPlaying)


src/renderer.ts
```diff
@@ -766,7 +766,9 @@ class Renderer extends EventEmitter<RendererEvents> {
     this.canvasWrapper.style.clipPath = `polygon(${percents}% 0%, 100% 0%, 100% 100%, ${percents}% 100%)`
     this.progressWrapper.style.width = `${percents}%`
     this.cursor.style.left = `${percents}%`
-    this.cursor.style.transform = `translateX(-${Math.round(percents) === 100 ? this.options.cursorWidth : 0}px)`
+    this.cursor.style.transform = this.options.cursorWidth
+      ? `translateX(-${progress * this.options.cursorWidth}px)`
```
katspaugh, line 770: Wouldn't this mess with the accuracy of the cursor? E.g. if the cursor is 10px wide
* in the beginning of the waveform, the offset will be 0, so the playhead will be aligned with the left side of the cursor
* in the middle of the waveform, the offset will be 5px, so the playhead will be in the middle of the cursor
* at the end, it will be aligned with the end of the cursor

The current behavior intends to just keep the cursor in viewport at the end of the waveform.

Tsuk1ko: I feel this problem is difficult to have the best of both worlds.

- The current behavior's jump affects the visual experience, and when users manually drag the cursor near the end, at a certain moment the cursor suddenly moves to the left, which is also quite confusing
- The changes I made can make this process smooth, but as you said, this will lose a bit of accuracy
  - I personally think this issue shouldn't be too significant, because typically cursors aren't set to such extreme width, and users should have difficulty perceiving the offset when dragging or clicking
  - If the cursor width is large, the current behavior would be even more peculiar - imagine the cursor suddenly shifting 10px to the left in the final stage
- If absolute accuracy is needed, then the translateX should always use `-${this.options.cursorWidth / 2}px`, but I notice the container currently `has overflow: hidden`, so if we want to implement it this way, the scope of changes would be much larger

katspaugh (approved):

src/renderer.ts
Tsuk1ko: Thanks!

PR #4171, opened 2025-07-23
feat: Autoscroll while dragging/updating/creating regions

clinton-encord:
## Short description
Resolves #4159

## Implementation details
- When region reaches the edge of the container, we autoScroll to bring it into view. This creates the effect where by leaving it at the edge will scroll the container.
- The draggable function also has to know about wavesurfer then to get the current scroll value.

## How to test it
- Set up wavesurfer, and zoom in such that its 
- Create/update/drag regions to the edge of the container

## Screenshots
Video available in the discussion.

## Checklist

### DIFF
--- a/AI_OVERVIEW.md
+++ b/AI_OVERVIEW.md
@@ -20,7 +20,7 @@ This document gives a condensed view of the project structure and build process 
 
 ## Contribution Notes
 
-- Follow the coding conventions in [`AGENTS.md`](../AGENTS.md).
+- Follow the coding conventions in [`AGENTS.md`](AGENTS.md).
 - Do not commit generated files from `dist/` or `node_modules/`.
 
 This overview should help an AI agent quickly locate relevant source files and scripts without traversing the entire repository.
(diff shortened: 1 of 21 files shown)


src/draggable.ts
```diff
@@ -1,8 +1,11 @@
+import Wavesurfer from './wavesurfer'
+
 export function makeDraggable(
   element: HTMLElement | null,
   onDrag: (dx: number, dy: number, x: number, y: number) => void,
   onStart?: (x: number, y: number) => void,
   onEnd?: (x: number, y: number) => void,
+  wavesurfer?: Wavesurfer,
```
katspaugh, line 8: The draggable module is currently wavesurfer-agnostic and I'd like to keep it this way. We could subscribe to the scroll event directly in the Regions plugin instead.

clinton-encord: Yeap that makes a lot of sense! Will change that!

clinton-encord:
Closing this PR as should actually be using the feat branch, instead of my fork's main branch.

PR #4178, opened 2025-07-30
fix: memory leak in Record and Regions

Tsuk1ko:
## Short description

I found that every time recording starts, `renderMicStream` is called, but it only disconnects audio nodes and closes the audio context in `onDestroy`, which causes these objects to not be released until the record plugin or wavesurfer instance is destroyed.

Additionally, one-time listeners are created but their references are not removed after triggered (accumulating in `this.subscriptions`). This also causes memory leak, and the objects they reference (like `AudioContext`) cannot be released.

## Implementation details

This is because calling `stopMic` in `onEnd` sets `this.stream = null`, so the next recording will call `startMic` -> `renderMicStream` again.

The above behavior is correct, so I moved the logic from `onDestroy` to `onEnd`.

Regarding the leak caused by `this.subscriptions`, I added an `addRemoveableSubscriptions` method to support removing function references after one-time events are triggered.

I found that both the Record and Regions plugins have behavior of pushing one-time events to `this.subscriptions`, so I made modifications to both.

Since I'm not very familiar with the Regions plugin, I can only do my best to ensure the modification logic is accurate. If you think this change is too significant, I can revert the Regions plugin part first.

### Before

In official demo: https://wavesurfer.xyz/examples/?record.js

I add log points:

image
image

After multiple recordings, you can see in the logs that there are no calls to `onDestroy`.

image

And after manual GC, there are still many `AudioContext` instances remaining here.

image

### After

Old `AudioContext` instances have been properly reclaimed.

image

## Checklist


katspaugh:
Thanks for the fix, @Tsuk1ko!

What do you think about this solution – #4179? I'd like to keep the changes as small as possible.

Tsuk1ko:
> Thanks for the fix, @Tsuk1ko!
> 
> What do you think about this solution – #4179? I'd like to keep the changes as small as possible.

Thanks, It looks good!

Also, I mentioned that the Regions plugin might have similar memory leak issues. If you have time, you could take a look at that too.

katspaugh:
Thanks! I’ll merge it then and will look into Regions separately. 👍

TD-jizhijun:
Hi @katspaugh, for this bug, when do you plan to release it? Recently I want to upgrade this repo. Thanks~

PR #4181, opened 2025-08-13
Feat/drag region while scrolling

clinton-encord:
## Short description
Resolves #4159

## Implementation details
- When region reaches the edge of the container, we autoScroll to bring it into view. This creates the effect where by leaving it at the edge will scroll the container.
- The draggable function also has to know about wavesurfer then to get the current scroll value.

## How to test it
- Set up wavesurfer, and zoom in such that its 
- Create/update/drag regions to the edge of the container

## Screenshots
Video available in the discussion.

## Checklist


clinton-encord:
@katspaugh from the discussion from the last PR (which I closed because that PR was made against my fork's main branch instead of a feature branch), I've made the feature work without introducing wavesurfer into draggable, which makes a lot of sense!

katspaugh:
Thank you @clinton-encord for this contribution! The regions plugin was substantially rewritten for v8 (reactive drag streams, scope-based lifecycle), so this branch can no longer be merged as-is.

The feature has been reworked for v8 in #4358, with your commits cherry-picked and adapted so the work remains attributed to you (the two feature commits there carry your authorship, with `cherry picked from` trailers referencing your original commits). Closing this PR in favor of #4358.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #4212, opened 2025-10-01
feat(Region): expose resized side on update end

ricky-c-2001:
## Short description
Exposes the side that is resized in the `update-end` event. 

My use case is that I want to play the surrounding audio when a side is resized.

## Implementation details
I've added `side` to the `update-end` event. 

So you can get the side like this:
```
regions.on('region-updated', (region, side) => {
  console.log('Updated region', region)
  console.log(side)
})
```

## How to test it

You can try it with yarn start and listen on `update-end`

## Screenshots

https://github.com/user-attachments/assets/cd0ad48c-4a35-4d2e-83d0-6462c5cdcea7

## Checklist


katspaugh (approved):
Thanks!

PR #4222, opened 2025-10-09
fix(Record): prevent waveform jitter in scrolling mode

### LINKED ISSUE #3844: Animation issues during scrolling
I'm seeing this even with the default renderer, exacerbated by using bar styling. @katspaugh 

https://github.com/user-attachments/assets/35bb68ef-fb1d-4d97-ba9a-44e78d68b848

_Originally posted by @sfc-gh-nbellante in https://github.com/katspaugh/wavesurfer.js/discussions/3318#discussioncomment-10468007_

katspaugh:
Resolves #3844.

## Short description
Fixes the jittering/dancing waveform issue in the Record plugin's scrolling mode.

## Implementation details

The jitter was caused by storing and rendering raw time-domain audio samples (rapid oscillations) instead of peak values (amplitude envelope). 

**Root cause:**
- Scrolling mode was storing ~240,000 raw audio samples per 5-second window
- These samples represent instantaneous amplitude oscillating at audio frequency
- Pixel-to-sample mapping captured different parts of oscillations on each render
- Result: Visual jitter as waveform peaks varied wildly

**Solution:**
1. Added `maxPeak` option to `WaveSurferOptions` for fixed normalization scale
2. Modified Record plugin to store peak values (one per frame) instead of raw samples
3. Reduced data window from `sampleRate × duration` to `FPS × duration` (500 vs 240,000 values)
4. Applied smaller FFT size (32) for responsive peak detection
5. Set fixed `maxPeak = 1` to prevent vertical scale changes during recording

## How to test it

1. Run the record example: `npm start` and open the record example
2. Enable "Scrolling waveform" mode
3. Start recording with audio input (speak or play music)
4. Observe the waveform scrolling smoothly without jittering or vertical scale changes

**Expected behavior:** Smooth amplitude envelope scrolling from right to left with stable vertical scale.

## Checklist
  - New `maxPeak` option is optional and backward compatible
  - Only affects Record plugin scrolling mode behavior

### DIFF
--- a/src/plugins/record.ts
+++ b/src/plugins/record.ts
@@ -103,7 +103,8 @@ class RecordPlugin extends BasePlugin<RecordPluginEvents, RecordPluginOptions> {
     const analyser = audioContext.createAnalyser()
     source.connect(analyser)
 
-    if (this.options.continuousWaveform) {
+    // Use smaller FFT size for more responsive peak detection
+    if (this.options.continuousWaveform || this.options.scrollingWaveform) {
       analyser.fftSize = 32
     }
     const bufferLength = analyser.frequencyBinCount
@@ -119,6 +120,9 @@ class RecordPlugin extends BasePlugin<RecordPluginEvents, RecordPluginOptions> {
       this.wavesurfer.options.interact = false
       if (this.options.scrollingWaveform) {
         this.wavesurfer.options.cursorWidth = 0
+        // Use fixed max peak in scrolling mode to prevent "dancing" waveform
+        this.wavesurfer.options.normalize = true
+        this.wavesurfer.options.maxPeak = 1
       }
     }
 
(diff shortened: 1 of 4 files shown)


stalar:
I have been using the Record plugin together with `normalize: true` to display the waveform during recording. An advantage of this has been that I see the waveform even if the sound level is very low. It seems that after this PR, I can no longer view the normalized waveform while recording - is this true or am I missing some setting that will allow this?

katspaugh:
@stalar correct, it sets a fixed max peak, so normalization doesn't work like before.

I'm not sure what would be a good way to normalize peaks in a scrolling waveform context. Peaks are added as audio is being recorded, so it would constantly re-normalize resulting in a "dancing" waveform.

stalar:
Thanks for confirming that. Would you be open to an option to allow normalization (i.e. not setting maxPeak) during recordings?

In our case the waveform has been a clear visual indicator for our users that a recording is happening and the audio is being captured. That the waveform is constantly re-normalizing during the recording has not been a problem.

PR #4223, opened 2025-10-16
Fix: center cursor at high levels of zoom

ricky-c-2001:
## Short description

At high zoom levels the centering of the progress bar doesn't keep up with the scrolling. 
Instead the progress bar gets to the end of the container then jumps back to the middle

## Implementation details
This was happening because the amount the cursor can jump was limited when centering.
I've removed this limit.

## How to test it

Zoom in to a waveform fully and press play

## Screenshots

Before:

https://github.com/user-attachments/assets/8583ee6a-6274-4b30-b319-aa49d92bca36

After:

https://github.com/user-attachments/assets/5f170538-0867-4579-8b0a-9af57fb61118

## Checklist

### DIFF
--- a/src/renderer.ts
+++ b/src/renderer.ts
@@ -689,7 +689,7 @@ class Renderer extends EventEmitter<RendererEvents> {
       // Keep the cursor centered when playing
       const center = progressWidth - scrollLeft - middle
       if (isPlaying && this.options.autoCenter && center > 0) {
-        this.scrollContainer.scrollLeft += Math.min(center, 10)
+        this.scrollContainer.scrollLeft += center
       }
     }
 


bernardwiesner:
Hi @katspaugh @ricky-c-2001 

This change removed the smooth scroll behavior. 

Before:
[Screencast from 2026-03-19 16-20-21.webm](https://github.com/user-attachments/assets/345b1dff-5798-4b36-bbb0-1aab39471a6f)

Now:
[Screencast from 2026-03-19 16-19-56.webm](https://github.com/user-attachments/assets/d9ee14fc-1032-421d-991b-d86f96453f37)

Could we have an option to enable or disable the smooth scroll instead?

katspaugh:
Ah, true, that clamping was for the smooth animation. I guess the magic 10 pixels didn't work for all zoom levels. We should probably replace it with a dynamic value based on zoom.

katspaugh:
@bernardwiesner does the PR above fix it for you?
Can be previewed here: https://copilot-replace-smooth-anima.wavesurfer-js.pages.dev/

bernardwiesner:
hi @katspaugh thanks for jumping on this, I tested it out [here](https://copilot-replace-smooth-anima.wavesurfer-js.pages.dev/zoom), but it does not look right:

[Screencast from 2026-03-19 18-10-15.webm](https://github.com/user-attachments/assets/80e9febb-19c6-4f47-ab62-23f134b2ff1e)

katspaugh:
I see, thanks. Looks like the math and numbers must be adjusted carefully so that it works on all zoom levels, and it'd have to be done by hand.

bernardwiesner:
@katspaugh 
Alternatively it would be nice if there could be an option to enable/disable the smooth scroll. The bug reported seems like an edge case that never bothered me as I never zoomed in that much while playing back. I think by default it could be enabled and if someone wants to disable the smooth scroll could pass an option to do so. 

Or figure out the math and fix it as AI sucks at that.

katspaugh:
@bernardwiesner should be good now. I restored the previous behavior when pix-per-sec is less than 600, and no smoothing when it's above that.

bernardwiesner:
thanks @katspaugh ! tested [here](https://copilot-replace-smooth-anima.wavesurfer-js.pages.dev/) and looks great.

PR #4247, opened 2025-11-24
fix: allow message in .destroy

CarrettaRiccardo:
Currently, aborting throws an error in console because no message was passed to abort, which is inside .destroy method (https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort#parameters)

## Implementation details
Added optional message prop to .destroy() method

In `src/wavesurfer.ts`
 ```tsx
 /** Unmount wavesurfer */
  public destroy(message?: string) {
    this.emit('destroy')
    this.abortController?.abort(message)
    this.plugins.forEach((plugin) => plugin.destroy())
    this.subscriptions.forEach((unsubscribe) => unsubscribe())
    this.unsubscribePlayerEvents()
    this.timer.destroy()
    this.renderer.destroy()
    super.destroy()
  }
```

### DIFF
--- a/src/wavesurfer.ts
+++ b/src/wavesurfer.ts
@@ -674,7 +674,10 @@ class WaveSurfer extends Player<WaveSurferEvents> {
     return this.renderer.exportImage(format, quality, type)
   }
 
-  /** Unmount wavesurfer */
+  /**
+   * Unmount wavesurfer
+   * @param message Optional message to pass to the abort signal when destroying. This can be used to provide a reason for the abort.
+   */
   public destroy(message?: string) {
     this.emit('destroy')
     this.abortController?.abort(message)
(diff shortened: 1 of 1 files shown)


katspaugh:
Good catch! Do we need to pass a message or can we just hard-code something like "Wavesurfer instance was destroyed"?

CarrettaRiccardo:
> Good catch! Do we need to pass a message or can we just hard-code something like "Wavesurfer instance was destroyed"?

Well, i think hardcoding might be okay too! Just want that console error to not print ;)

katspaugh:
@CarrettaRiccardo I'll have to close this PR because in v8, the abort signal is in a different place. Please feel welcome to make a new PR. Thanks!

PR #4268, opened 2025-12-20
fix: fix position jump on playback rate change

orionmiz:
## Short description

This PR fixes a bug where currentTime returns an incorrect value when playbackRate is modified, causing the progress tracking to jump abnormally.

## Implementation details

- Buffer-based playbackPosition: Renamed playedDuration to playbackPosition and refactored the logic to store the absolute offset in the audio buffer (in seconds) rather than cumulative wall-clock time.
- Improved currentTime Logic: Updated the currentTime getter to calculate the position as playbackPosition + (elapsed_time * playbackRate). This ensures that the tracked progress is independent of past rate changes.
- Accurate Rate Transitions: The playbackRate setter now snapshots the current playbackPosition by momentarily pausing and resuming. This maintains a seamless progress flow even during rapid rate changes, such as DJ deck adjustments.
- Direct Seek Assignment: The currentTime setter now assigns the target value directly to playbackPosition, eliminating incorrect rate-based scaling during seek operations.

## How to test it

1. Initialize WaveSurfer with the WebAudio backend.
2. Change the playbackRate while the audio is playing (example: call wavesurfer.setPlaybackRate(1.2)).
3. Verify: The progress tracking does not jump forward or backward and stays perfectly in sync with the audio.
4. Perform a seek operation and verify it lands on the correct second regardless of the current playbackRate.

## Screenshots

Here are videos changing playback rate dynamically:

Before:

https://github.com/user-attachments/assets/b45df7fb-7eaf-43a2-a21c-95aacfda4383

After:

https://github.com/user-attachments/assets/00e8e1a4-490c-467c-bb6c-37ea660ecfa0

## Checklist

### DIFF
--- a/src/webaudio.ts
+++ b/src/webaudio.ts
@@ -24,7 +24,7 @@ class WebAudioPlayer extends EventEmitter<WebAudioPlayerEvents> {
   private gainNode: GainNode
   private bufferNode: AudioBufferSourceNode | null = null
   private playStartTime = 0
-  private playedDuration = 0
+  private playbackPosition = 0
   private _muted = false
   private _playbackRate = 1
   private _duration: number | undefined = undefined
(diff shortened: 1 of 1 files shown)


orionmiz:
@katspaugh Could you review this PR?

PR #4270, opened 2026-01-03
fix region overlap detection code

osheroff:
## Short description

If calls to addRegion() came in all at once, we'd end up in a situation where all overlapping regions moved themselves to avoid each other (because they were all working with the same list of regions). This is incorrect, the first region shouldn't move at all.  We leverage the ordering of the `regions` array to ensure that only the later overlapping regions move.

And I added 2 pixels of room because it just looks nicer.

## Screenshots

before:
Screenshot 2026-01-03 at 10 48 35 AM

after:
Screenshot 2026-01-03 at 10 48 44 AM


katspaugh (approved):
Nice, thank you!

PR #4271, opened 2026-01-05
allow github builds

osheroff:
i'm impatient and want to point my project at the wavesurfer.js main branch, but it doesn't worth without being able to call the `prepare` task


katspaugh (approved):
Sure, makes sense!
Btw, I've added you as a collaborator to this repo, so you can make make PRs directly to it instead of from your fork.

PR #4307, opened 2026-05-07
fix zoom.js example to use hideScrollbar

87113:
## Short description

When I used the zoom.js example, the setting 'Scroll bar' didn’t work, I wondered why, and it's just that the setting was not correct any more.

I upgraded the file according to the official doc https://wavesurfer.xyz/docs/types/wavesurfer.WaveSurferOptions#__type.hideScrollbar

### DIFF
--- a/examples/zoom.js
+++ b/examples/zoom.js
@@ -34,7 +34,7 @@ wavesurfer.once('decode', () => {
 
 /*
 <html>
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
Ah, good catch, thank you!

PR #4312, opened 2026-05-25
Pitch Adjuster Plugin

fvlasie:
Adds a plugin for pitch adjustment using SoundTouchJS


katspaugh:
Wow, that's very cool!

Any chance you could add some e2e tests for it? Thank you!

katspaugh:
@fvlasie I'll have to close this PR, because CI is failing and there are unresolved comments from Copilot.
Since this plugin requires a 3rd-party dependency, perhaps it would be best if you host it in your own repository. I'd be more than happy to link it from wavesurfer's readme, just let me know. Thank you!

PR #4324, opened 2026-06-15
fix(Spectrogram): close ImageBitmap when its canvas is removed before it resolves

### LINKED ISSUE #4323: Spectrogram: ImageBitmap not closed when its canvas is removed before the bitmap resolves
### Description

In the regular `SpectrogramPlugin`, `drawSpectrogramSegment()` rasterises each chunk with `createImageBitmap()` and, in the promise callback, only calls `bitmap.close()` inside the `if (ctx.canvas.parentNode)` check:

https://github.com/katspaugh/wavesurfer.js/blob/48ddb37f3a132cfa7a5d417f2a8c1ad23a064252/src/plugins/spectrogram.ts#L780-L801

When the canvas is removed before the bitmap resolves, the resolved `ImageBitmap` is neither drawn nor closed. This can happen on:

- a zoom re-render — `drawSpectrogram()` calls `clearCanvases()` before redrawing;
- unregistering the plugin while a render is in flight;
- `destroy()`, which only does `this.pendingBitmaps.clear()` and never closes resolved bitmaps.

It isn't a crash (the `drawImage` stays guarded) and the wrapper is eventually garbage-collected, but `ImageBitmap.close()` is the deterministic way to release the bitmap's (often native) graphical resources. Relying on GC means large chunk bitmaps can be retained during rapid zoom/toggle/teardown, since the small JS wrapper creates little GC pressure relative to the backing store. It also defeats the existing `// Clean up bitmap to free memory` intent in exactly the case where the cleanup matters.

The `WindowedSpectrogramPlugin` is not affected — it awaits the bitmap and closes it unconditionally.

### Steps to reproduce

1. Register the (non-windowed) `SpectrogramPlugin` on a longer audio file.
2. Zoom in/out rapidly, or toggle the spectrogram off, while chunks are still rendering.
3. In a memory profiler, observe that bitmaps which resolve after their canvas was removed are not released until a later GC.

### Suggested fix

Always close the bitmap once it resolves, keeping the `drawImage` guarded — e.g. move `close()` into a `finally`. PR to follow.

koyukan:
## Short description

Resolves #4323

## Implementation details

In the regular `SpectrogramPlugin`, `drawSpectrogramSegment()` rasterises each chunk with `createImageBitmap()` and, in the promise callback, only called `bitmap.close()` inside the `if (ctx.canvas.parentNode)` guard. When the canvas was removed before the bitmap resolved — a zoom re-render (`drawSpectrogram()` runs `clearCanvases()` first), unregistering the plugin, or `destroy()` (which only does `pendingBitmaps.clear()` and never closes resolved bitmaps) — the resolved `ImageBitmap` was neither drawn nor closed, so its resources waited on GC instead of being released.

This moves the cleanup into a `finally`: the `drawImage` stays guarded by `ctx.canvas.parentNode`, but the bitmap is always closed once it resolves (and also if `drawImage` itself throws). No API or behavior change when the canvas is present.

The `WindowedSpectrogramPlugin` already closes unconditionally, so it was not affected.

## How to test it

Register the (non-windowed) `SpectrogramPlugin` on a longer audio file and zoom/toggle the spectrogram rapidly while chunks are rendering. With a memory profiler, bitmaps that resolve after their canvas was removed are now closed immediately rather than retained until a later GC. There is no visual change.

## Screenshots

n/a

## Checklist

### DIFF
--- a/src/plugins/spectrogram.ts
+++ b/src/plugins/spectrogram.ts
@@ -782,14 +782,16 @@ class SpectrogramPlugin extends BasePlugin<SpectrogramPluginEvents, SpectrogramP
         // Remove from pending set
         this.pendingBitmaps.delete(bitmapPromise)
 
-        // Check if canvas is still valid before drawing
-        if (ctx.canvas.parentNode) {
-          const drawHeight = (height * rMax1) / rMax
-          const drawY = yOffset + height * (1 - rMax1 / rMax)
+        try {
+          // Check if canvas is still valid before drawing
+          if (ctx.canvas.parentNode) {
+            const drawHeight = (height * rMax1) / rMax
+            const drawY = yOffset + height * (1 - rMax1 / rMax)
 
-          ctx.drawImage(bitmap, 0, drawY, canvasWidth, drawHeight)
-
-          // Clean up bitmap to free memory
+            ctx.drawImage(bitmap, 0, drawY, canvasWidth, drawHeight)
+          }
+        } finally {
+          // Clean up bitmap to free memory, even if the canvas was removed before it resolved
           if ('close' in bitmap) {
             bitmap.close()
           }


katspaugh (approved):
Great fix, thank you!

koyukan:
Gentle nudge on this one — it's approved with green checks, just needs a merge whenever you get a chance. Thanks!

katspaugh:
Merged, thanks!

PR #4326, opened 2026-06-17
fix(Spectrogram): make the web worker FFT timeout configurable

### LINKED ISSUE #4325: Spectrogram freezes the page on long audio when the web-worker FFT exceeds the hardcoded 30s timeout
## Bug description

When `useWebWorker: true`, the spectrogram FFT runs in a web worker with a **hardcoded 30s timeout** in `calculateFrequenciesWithWorker()`:

```js
setTimeout(() => {
  if (this.workerPromises.has(id)) {
    this.workerPromises.delete(id)
    reject(new Error('Worker timeout'))
  }
}, 30000) // 30 second timeout
```

For long recordings the worker FFT can legitimately take longer than 30s. When it does, the request is rejected and `getFrequencies()` falls back to computing the FFT **synchronously on the main thread**:

```js
if (this.useWebWorker && this.worker) {
  try {
    return await this.calculateFrequenciesWithWorker(buffer)
  } catch (error) {
    console.warn('Worker calculation failed, falling back to main thread:', error)
    // Fall through to main thread calculation
  }
}
```

That main-thread FFT blocks the UI thread (no input, no rendering) until it finishes — i.e. moving the work to a worker is exactly defeated for the files that need it most. The timeout is not configurable, so there is no way to avoid the freeze short of disabling the worker entirely.

The worker FFT path and the 30s timeout were introduced in #4125.

## Environment
 - Browser: Chrome (not browser-specific)

## Minimal code snippet

```js
import WaveSurfer from 'wavesurfer.js'
import Spectrogram from 'wavesurfer.js/dist/plugins/spectrogram.esm.js'

const ws = WaveSurfer.create({ container: '#app', url: '/long-audio.mp3' }) // a long recording (e.g. 30+ min)
ws.registerPlugin(Spectrogram.create({ labels: true, useWebWorker: true }))
// When the worker FFT exceeds 30s it is aborted, then recomputed on the main
// thread, and the page is unresponsive until that computation finishes.
```

The trigger is simply "worker FFT slower than the timeout"; multi-hour mono files reliably exceed 30s of worker compute on mid-range hardware.

## Expected result

A slow worker FFT keeps running in the worker — or, at least, the wait is configurable — so the main thread stays responsive.

## Obtained result

After 30s the worker result is discarded and the whole-file FFT is recomputed synchronously on the main thread, freezing the page until it completes.

## Screenshots

N/A (timing/responsiveness issue).

koyukan:
## Short description

Resolves #4325. The spectrogram web-worker FFT had a hardcoded 30s timeout; on long files the worker is aborted and the plugin falls back to a synchronous main-thread FFT that freezes the page. This adds a `workerTimeout` option so the wait can be scaled to the file length or disabled, with no change in default behaviour.

## Implementation details

- Add `workerTimeout?: number` to `SpectrogramPluginOptions` — default `30000` (unchanged behaviour), `0` disables the timeout. Only used when `useWebWorker` is `true`.
- `calculateFrequenciesWithWorker()` uses the option instead of the literal `30000`, and skips the timer entirely when it is `0`.
- Store the timer handle on the pending-promise record and clear it when the request settles (`onmessage`) and in `destroy()`, so a timer can't outlive its request — relevant once the timeout can be disabled.
- No default/behaviour change for existing users; no API removed.

## How to test it

Register the spectrogram with `useWebWorker: true` on a long recording whose worker FFT exceeds the timeout:

- With the default (`30000`) the worker is aborted after 30s and the FFT is recomputed on the main thread — the page freezes until it finishes (current behaviour).
- With `workerTimeout: 0` (or a higher value) the worker FFT runs to completion and the page stays responsive.

## Screenshots

N/A (timing/responsiveness change).

## Checklist
* [ ] This PR is covered by e2e tests — the timeout/worker timing isn't deterministically testable in the cypress suite; verified manually with long files. Happy to add a test if you'd like a particular shape.

### DIFF
--- a/src/plugins/spectrogram.ts
+++ b/src/plugins/spectrogram.ts
@@ -355,8 +355,10 @@ class SpectrogramPlugin extends BasePlugin<SpectrogramPluginEvents, SpectrogramP
       this.worker.terminate()
       this.worker = null
     }
+    const error = new Error('Spectrogram plugin destroyed')
     this.workerPromises.forEach((promise) => {
       if (promise.timer) clearTimeout(promise.timer)
+      promise.reject(error)
     })
     this.workerPromises.clear()
 
--- a/src/plugins/spectrogram.ts
+++ b/src/plugins/spectrogram.ts
@@ -112,6 +112,12 @@ export type SpectrogramPluginOptions = {
   maxCanvasWidth?: number
   /** Use web worker for FFT calculations (default: false) */
   useWebWorker?: boolean
+  /**
+   * Max time in milliseconds to wait for the web worker FFT result before the calculation is
+   * rejected (and falls back to the main thread). Set to 0 to disable the timeout. Only used when
+   * useWebWorker is true. (default: 30000)
+   */
+  workerTimeout?: number
 }
 
 export type SpectrogramPluginEvents = BasePluginEvents & {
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
Sorry for the delay, lgtm!

PR #4327, opened 2026-07-04
fix(Spectrogram): use the same window alpha defaults in the worker and main thread

koyukan:
## Short description

The spectrogram worker builds its FFT with `alpha || 0.16`, so a `gauss` spectrogram rendered with `useWebWorker: true` uses a 0.16 window width while the main thread uses the FFT's own gauss default of 0.25 (`fft.ts`) — the same options produce two visibly different images depending on where the FFT runs. 0.16 is blackman's default, misapplied to every window function. This passes `alpha` through untouched so the per-window defaults in `fft.ts` apply identically on both paths.

## Implementation details

- One line in `spectrogram-worker.ts`: drop the `|| 0.16` coercion at the FFT construction. Both plugins that use this worker (`spectrogram.ts` and `spectrogram-windowed.ts`) already send `alpha` through unmodified, and the main-thread paths already construct `FFT` with the raw option, so this is worker/main parity only.
- Output changes for `gauss` worker users who never set `alpha` — it starts matching the main-thread/documented behavior. For `gauss`, explicit falsy values (`alpha: 0`, `NaN`) also now resolve to the same FFT default the main thread applies to them. Alpha validation itself is untouched and worth a separate discussion.
- New jest suite (`src/__tests__/spectrogram-worker.test.ts`) drives the worker's `onmessage` handler in-process and asserts byte-exact parity with the main-thread compute path. The default-gauss case fails without the fix; a second case pins explicit-alpha pass-through.

## How to test it

Render the same file twice with `windowFunc: 'gauss'` and no `alpha`, once with `useWebWorker: true` and once without — before the fix the two spectrograms differ, after the fix they are pixel-identical (verified via canvas `toDataURL` comparison in Chrome). And `npm run test:unit`.

## Screenshots

`examples/audio/demo.wav`, `windowFunc: 'gauss'`, `scale: 'linear'`, no `alpha` set:

| | |
|---|---|
| main thread (unchanged) | main thread |
| worker, before | worker before |
| worker, after (matches main thread) | worker after |

## Checklist
* [ ] This PR is covered by e2e tests — the spectrogram e2e suite is currently disabled (`xdescribe`), so the numeric behavior is covered by the new jest suite instead.

### DIFF
--- a/src/__tests__/spectrogram-worker.test.ts
+++ b/src/__tests__/spectrogram-worker.test.ts
@@ -1,0 +1,118 @@
+import FFT from '../fft.js'
+import '../plugins/spectrogram-worker.js'
+
+type WorkerOptions = {
+  startTime: number
+  endTime: number
+  sampleRate: number
+  fftSamples: number
+  windowFunc: string
+  alpha?: number
+  noverlap: number
+  scale: 'linear' | 'logarithmic' | 'mel' | 'bark' | 'erb'
+  gainDB: number
+  rangeDB: number
+  splitChannels: boolean
+}
+
+const SAMPLE_RATE = 8000
+const GAIN_DB = 20
+const RANGE_DB = 80
+
+/** A 1 kHz sine, long enough for several FFT frames */
+function makeSine(length: number): Float32Array {
+  const signal = new Float32Array(length)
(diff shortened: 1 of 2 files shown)


katspaugh (approved):
👍

PR #4336, opened 2026-07-10
fix(Decoder): avoid closing a closed AudioContext

mjlbach:
## Short description

Prevent decoder cleanup from attempting to close an `AudioContext` that is already closed.

Safari can reject the promise returned by `AudioContext.close()` with `InvalidStateError: Cannot close a closed AudioContext.` Because the existing call is fire-and-forget, that rejection surfaces as an unhandled browser exception even though decoding has already completed.

## Implementation details

- Check the context state before closing it.
- Await the close promise so cleanup completes within the decoder promise instead of producing an unhandled rejection.
- Add unit coverage for both active and already-closed contexts.

### DIFF
--- a/src/__tests__/decoder.test.ts
+++ b/src/__tests__/decoder.test.ts
@@ -1,0 +1,41 @@
+import Decoder from '../decoder.js'
+
+describe('Decoder', () => {
+  const originalAudioContext = global.AudioContext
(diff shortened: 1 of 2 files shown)


katspaugh (approved):
Thank you!

mjlbach:
Thanks! Sorry I dropped off but glad you got it merged!

PR #4339, opened 2026-07-16
fix(regions): prune per-region subscriptions from plugin-level array on remove

jens-lilt:
## Problem

`RegionsPlugin` pushes an unsubscribe closure into the plugin-level `subscriptions` array for every event it wires up per region — roughly 11 per region:

- `saveRegion`: 7 (`update`, `update-end`, `play`, `click`, `dblclick`, `content-changed`, `remove`)
- `virtualAppend`: 4 (`scroll`, `zoom`, `resize`, region `render`)
- `addRegion` pre-decode: 1 (`once('ready')`)

That array is only ever emptied in `destroy()`. When a region is removed, its `once('remove')` handler *calls* the unsubscribe closures but never *removes them from the array*. Since each closure captures its region (and its DOM element), **every region ever created stays retained for the plugin's lifetime**.

In long-lived players that repeatedly add and remove regions (e.g. an editor that rebuilds regions as the user works), this is an unbounded memory leak. The ever-growing `subscriptions` array also turns each rebuild into quadratic work.

Note: #4258 fixed the *per-region* `SingleRegion.subscriptions` (DOM listeners) but not this *plugin-level* array, so the leak still reproduces on current `main`.

## Fix

Prune each region's closures from `this.subscriptions` at the moment the region is removed, and drop the pre-decode `once('ready')` unsubscriber once it has fired:

- `saveRegion` / `virtualAppend`: on `region.once('remove')`, after invoking the closures, `filter` them out of `this.subscriptions`.
- `addRegion`: after `once('ready')` fires, remove its unsubscriber from `this.subscriptions`.

Behavior is otherwise unchanged — the same closures still run on removal; they just no longer linger in the array afterwards.

## Verification

- `tsc --noEmit` clean
- `eslint src/plugins/regions.ts` clean

Repeatedly adding/removing regions now keeps `plugin.subscriptions` bounded (returns to baseline after each add/remove cycle) instead of growing without limit.

### DIFF
--- a/src/plugins/regions.ts
+++ b/src/plugins/regions.ts
@@ -731,7 +731,8 @@ class RegionsPlugin extends BasePlugin<RegionsPluginEvents, RegionsPluginOptions
       const unsubscribeRender = region.on('render', renderIfVisible)
 
       // Only push the unsubscribe functions, not the once() return values
-      this.subscriptions.push(unsubscribeScroll, unsubscribeZoom, unsubscribeResize, unsubscribeRender)
+      const virtualAppendSubscriptions = [unsubscribeScroll, unsubscribeZoom, unsubscribeResize, unsubscribeRender]
+      this.subscriptions.push(...virtualAppendSubscriptions)
 
       // Clean up subscriptions when region is removed
       region.once('remove', () => {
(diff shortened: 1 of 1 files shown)


katspaugh (approved):
Great catch!

katspaugh:
@codex review

katspaugh:
@codex it's already connected, review the PR

katspaugh:
@codex try again

Issue #4043, opened 2025-02-26
setTime and seekTo not working correctly on ios Chrome & Safari in on ready


danyal:
## Bug description
I have tested this in my own code as well as directly on https://wavesurfer.xyz/examples/?soundcloud.js . If you add setTime or seekTo to any of the 'ready' events (ready, redrawcomplete etc), hitting play on iOS (chrome & safari) starts the track at 0 and breaks the seek bar. The waveform looks like it's properly set to the right time, but it always starts at 0. I tested this with preloading peaks and without. On chrome OSX this works fine. It's only broken on iOS.

## Environment
 - Browser: Chrome & Safari
 - OS: iOS (latest)

## Minimal code snippet
 wavesurfer.on('ready', () => {
     wavesurfer.setTime(10);
 })

## Expected result
- Seeks to the right time
- Starts playing from that position when you hit play

## Obtained result
- Seeks to the right time
- Starts playing at 0
- Breaks seek bar (interaction)

## Screenshots
- Hard to provide a screenshot since you have to hear the audio.

katspaugh:
Thanks for the report!

Does it work with `backend: 'WebAudio'`?

danyal:
If I add that parameter, the following things happen on iOS (with seeking):
1. No audio -- couldn't get any audio output even though wavesurfer shows the track as playing
2. Whole file is loaded even with peaks provided -- not the desired behavior I'm looking for. Need tracks to load as they stream because these are large files
3. I can't tell if the seek behavior is fixed because I can't heard any audio

hadnet:
> Thanks for the report!
> 
> Does it work with `backend: 'WebAudio'`?

Confirming that switching to WebAudio fixes the issue, but I wanted to use MediaElement because it lets you listen in the background, since with WebAudio the audio gets paused when the app goes into the background. Why doesn’t it work properly with MediaElement? In Android environment is working well, though, problem is only in iOS. Did you find a solution @danyal ?

danyal:
Unfortunately not.

murillo94:
@danyal Did you find a solution? if not, are you using another lib to solve your issue?

danyal:
I didn't find a solution unfortunately. Just using it without this
functionality.

aribradshaw:
[...]

PR #4353, opened 2026-08-20
fix(Player): defer early media seeks until canplay

aribradshaw:
## Short description

Addresses the pre-playback MediaElement seek path reported in #4043.

WebKit has open defects where assigning `currentTime` during `loadedmetadata` can start playback at zero and leave seeking unusable ([WebKit #261512](https://bugs.webkit.org/show_bug.cgi?id=261512), [WebKit #201216](https://bugs.webkit.org/show_bug.cgi?id=201216)). This change avoids that unsafe timing window: an early WaveSurfer seek remains visible through `getCurrentTime()`, then reaches the native media element once it emits `canplay`.

The previous head passed all required checks. This test-only follow-up is ready for review, and its new fork workflows are awaiting repository approval. Physical iOS confirmation of the original scenario remains welcome as an additional validation.

## Implementation details

- Keeps only the latest seek requested before `HAVE_FUTURE_DATA`.
- Applies the pending time on `canplay`, immediately before playback when already playable, or after a resolved custom-media `play()` fallback.
- Avoids assigning `currentTime` from `loadedmetadata`.
- Cancels a stale pending value when an external seek chooses a different time.
- Clears pending state when the source, media element, or player lifecycle changes.
- Preserves immediate seeks for already-playable and WebAudio-backed media.

## Regression coverage

Unit tests cover:

- no native assignment during `loadedmetadata`
- application at `canplay`
- latest-request-wins behavior
- external seek cancellation
- play-before-`canplay` and already-playable paths
- source and media-element replacement
- existing stop-position clamping behavior

A focused Cypress test uses precomputed peaks so `ready` fires before the media can play, registers `canplay` before calling `setTime(10)`, verifies the deferred seek reaches the media element, then uses a retrying assertion to confirm playback advances from the requested position. The WaveSurfer instance is destroyed after each test.

## Verification

- `yarn test:unit --runInBand`: 39 suites, 580 tests passed
- leak suite: 7 tests passed
- `yarn tsc`: passed
- `yarn lint`: passed
- Prettier check: passed
- focused Cypress regression: passed repeatedly in Edge 151 and Electron 118
- additional local Player probe: passed 5/5 in Playwright WebKit

`yarn build` emits the main ESM, CJS, and minified bundles, then hits the repository's existing Windows Rollup plugin path constraint: the TypeScript `outDir` is outside the plugin output directory.

## Screenshots

Not applicable. This changes media timing behavior and adds automated browser coverage.

## Checklist

- [x] This PR is covered by e2e tests
- [x] It introduces no breaking API changes

### DIFF
--- a/cypress/e2e/player.cy.js
+++ b/cypress/e2e/player.cy.js
@@ -1,52 +1,58 @@
 describe('WaveSurfer player', () => {
+  let wavesurfer
+  let media
+
+  afterEach(() => {
+    wavesurfer?.destroy()
+    wavesurfer = undefined
+    media = undefined
+  })
+
   it('preserves a seek requested from ready before media can play', () => {
     cy.visit('cypress/e2e/player.html')
     cy.window().its('WaveSurfer').should('exist')
 
-    cy.window().then((win) => {
-      return new Cypress.Promise((resolve, reject) => {
-        const wavesurfer = win.WaveSurfer.create({
-          container: '#waveform',
-          url: '../../examples/audio/demo.wav',
-          peaks: [[0, 0.5, -0.5, 0]],
-          duration: 21.77,
-        })
+    cy.window()
+      .then((win) => {
+        return new Cypress.Promise((resolve, reject) => {
+          wavesurfer = win.WaveSurfer.create({
+            container: '#waveform',
+            url: '../../examples/audio/demo.wav',
+            peaks: [[0, 0.5, -0.5, 0]],
+            duration: 21.77,
+          })
 
-        wavesurfer.once('ready', () => {
-          const media = wavesurfer.getMediaElement()
-
-          try {
-            expect(media.readyState).to.be.lessThan(HTMLMediaElement.HAVE_FUTURE_DATA)
-            wavesurfer.setTime(10)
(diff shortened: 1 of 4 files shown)


katspaugh:
Looks good, @aribradshaw! Please let me know when it's ready for review.

aribradshaw:
Thanks! This is now ready for review. Lint, unit, and E2E checks are all green.

aribradshaw:
Follow-up pushed in `9d08112` after the timing review. The Cypress regression now registers `canplay` before `setTime()`, replaces the fixed 200 ms delay with a retrying assertion, and destroys the WaveSurfer instance after each test. I also added explicit lifecycle assertions for source replacement, media replacement, and destroy.

Local verification is green: 39 suites / 580 tests, TypeScript, ESLint, Prettier, repeated Electron and Edge runs, plus 5/5 local Playwright WebKit probes. Physical iOS confirmation remains welcome.

The new upstream workflow runs were created as `action_required` with no jobs, so they need the repository's fork-workflow approval before Chrome E2E, Jest, and eslint can run on this commit.

katspaugh (approved):
Thanks @aribradshaw!

